PR #51, opened 2013-12-02
Allow developers to easily upgrade to the latest git code

ljm42:


fisharebest:
I'm going to close this, as I don't think this has a place in the core code.

If needed, I think that a custom module could simply set the `WT_LATEST_VERSION` to point to a custom URL, and then the auto-updater should take over.

PR #88, opened 2014-02-02
Convert media files to the right code page on windows hosted sites, othe...

magicsunday:
...rwise files containing special chars (german umlauts for instance) won't be displayed.


KosherJava:
It looks like Patchwork-UTF8 1.2 was released
https://github.com/nicolas-grekas/Patchwork-UTF8/blob/master/CHANGELOG.md

fisharebest:
> It looks like Patchwork-UTF8 1.2 was released

...and webtrees updated to in on 2015-01-18 ;-)

KosherJava:
Does that mean that the issue is automatically resolved? or is it now possible to fix it?

fisharebest:
It means that as soon as I get time, I will read the documentation, and see if we can use it.

PR #334, opened 2014-12-23
Fix colors theme CSS overrides

vytux-com:
Custom module menu's may insert links to own CSS files before main CSS
is loaded and thus wrong entries can be chosen. This updates makes
colors theme more consistent with other themes


vytux-com:
Just confirming that after the latest theme changes this will no longer be an issue in the next release

Issue #348, opened 2015-01-04
Validation error in descendancy chart


ddrury:
I recently came to the conclusion that adding a class to an element on the descendancy chart would simplify the css (unordered list at line 209) and whilst messing about decided to fix the validation errors which were all pretty simple bar the following

WT_Controller_Descendancy->printFamilyDescendancy() wraps a call to WT_fact->summary() (line202) in an anchor. WT_fact->summary() in turn calls WT_Gedcom_Tag::getLabelValue() (line 399). 

This as it stands return it's html wrapped in a div so we have a div wrapped by an anchor which fails validation.

I tried changing the call to WT_Gedcom_Tag::getLabelValue() to make it return a span - 
this works in this instance but ruins the display elsewhere - I guess we could change the css for the broken calls to include "display:block"

changing line 202 of WT_Controller_Descendancy->printFamilyDescendancy() to

    echo ' <a href="', $family->getHtmlUrl(), '" class="details1">', str_replace('div', 'span',       $fact->summary()), '</a>'; 

also works but I would suggest isn't a real fix.

So ... over to you

ddrury:
Hm,

There are a lot of posts on the net saying that wrapping divs with anchors is valid HTML5 so maybe the validator is broken?

PR #351, opened 2015-01-06
Decouple popup menu css from charts css

ddrury:
Current code/css for ancestry & decendancy charts makes it very difficult to style the popup menus (Yeah I know you love them!) without having a knock on effect on the chart. 

By adding a class to the chart code (class="generation") it makes it possible to completely decouple the menu. At the same time
1.  I realized that the code for both charts was virtually identical which allows for common css (class="chart_common")
2.  I could remove validation errors (mostly table related stuff) - I've ignored nested divs within anchors (see issue #348)

### DIFF
--- a/ancestry.php
+++ b/ancestry.php
@@ -142,7 +142,7 @@ switch ($controller->chart_style) {
 case 0:
 	// List
 	$pidarr=array();
-	echo '<ul id="ancestry_chart">';
+	echo '<ul id="ancestry_chart" class="chart_common">';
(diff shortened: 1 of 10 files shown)


makitso:
Dave,
This is a good idea.  However,  I am restructuring all the themes and its taking longer than I expected.  So, can you hold off on any theme changes until I am completed?

Rob

ddrury:
Sure

Regards
David

Issue #350, opened 2015-01-05
 Missing spouse name in pedigree and family reports


loudubewe:
This is my very first move to help in webtrees development. Advices are welcome.

The name of the spouse of the root person is missing in the pedigree report.
The names of children spouses are missing in the family report.
This is due to a never met condition: $pid==@HUSB or $pid==@WIFE.
Never met because this condition is translated into "I1234"=="@I1234@" by the 'ifStartHandler' function in file Base.php (the second value is returned by the 'get_gedcom_value' function).
The 2 @s should be stripped off when the the returned gedcom value is an individual reference (e.g. "@I1234@") but not from another string (e.g. a comment like "see IND @1234@").

A simple and robust way to solve that problem would be to insert the following line after the calls to 'get_gedcom_value':
            $value = preg_replace("/^@([I]\d+)@$/", "$1", $value);
at line 1834, just before the line:
            $value = "\"" . addslashes($value) . "\"";

Shall I do that ?
Louis

fisharebest:
`/^@([I]\d+)@$/`

The IDs do not always take this format.  There is a constant WT_REGEX_XREF which matches them consistently.

I will look at the code later.

loudubewe:
You are right, using the constant WT_REGEX_XREF is quite more general.
Thus, the following line of code will strip off any occurrence of an ID from its enclosing @s (in the 'ifStartHandler' function of 'Base.php'):
	$value = preg_replace("/^@(".WT_REGEX_XREF.")@$/", "$1", $value);
This gives the correct result (spouse names are displayed in reports) and has no other unexpected consequence, as far as I could verify for all the other calls to the 'ifStartHandler' function in 'report.xml' files.
So, I will try to send a 'pull request' with the modified file 'Base.php'.
Thanks.

PR #353, opened 2015-01-09
Spouse bug corrected in Base.php

loudubewe:
(See issue https://github.com/fisharebest/webtrees/issues/350)
The spouse name of the root person in pedigree report & children spouse names in family report were missing.
This version of 'Base.php' file fix the bug.


fisharebest:
Thanks!

PR #356, opened 2015-01-12
lightweight markdown/html editor and module to use it

ddrury:
markitup is a lightweight markdown/html editor.

The blog module replaces the gedcom_news and user_blog modules (and at a push the HTML module). It supports language selection and embedded stats. It is written as a stand-alone module (although see below) but should it be adopted, the css and relevant database functions probably should be moved to the core, and a section within db_schema_3_4.php that removes the subject column uncommented. Adding and editing articles both utilise the markitup module within a modal dialog.

To accomodate the language selection the news table schema needs to be updated -  to restore the status quo, remove the languages column and revert the schema version


ddrury:
Needs rework due to theme changes

PR #357, opened 2015-01-13
Remove dialog related html from dom on closure

ddrury:
Each time a dialog is opened jQuery creates a new set of markup, however on closure it isn't remove from the DOM, so stale, inaccessible markup can accumulate.


fisharebest:
Thanks!

PR #358, opened 2015-01-14
Decouple popup menu css from charts css

ddrury:
By adding a couple of classes to the ancestry & descendancy chart code, it allows for the complete decoupling of the popup menu css making it simpler and more straightforward to style.

The css for generating the connecting lines on the ancestry & descendancy charts are the same, so have combined the css.

Removed inline css allowing the code to pass HTML5 validation (bar divs nested within anchors which although it fails validation is apparently valid in HTML5).

Although I think it looks quite nice, it's not part of the original styling so have commented out box-shadow on the popup menu.

### DIFF
--- a/ancestry.php
+++ b/ancestry.php
@@ -142,7 +142,7 @@ switch ($controller->chart_style) {
 case 0:
 	// List
 	$pidarr=array();
-	echo '<ul id="ancestry_chart">';
+	echo '<ul id="ancestry_chart" class="chart_common">';
 	$controller->printChildAscendancy($controller->root, 1, $OLD_PGENS-1);
 	echo '</ul>';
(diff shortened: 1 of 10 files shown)


fisharebest:
Github doesn't always seem to notice when PRs have been merged via the command line....

Thanks.  I've uncommented the shadows.

Art4:
@fisharebest Github closes issues if you put a keyword and the issue ID in your commit message, eg `Fixes #358`. See more keywords on this helpsite: https://help.github.com/articles/closing-issues-via-commit-messages/

fisharebest:
Ah I see.  "Merge" creates a second commit (containing the commit message), but "Rebase" doesn't...

Thanks.

Issue #405, opened 2015-02-01
English language typos?


vytux-com:
While translating LT.po i noticed a few weird english text such as:

begining with "nn"

__716 - admin_trees_config.php__
<?php echo /* I18N: Help text for the “Language” configuration setting */ WT_I18N::translate('nnIf a visitor to the website has not specified a preferred language in their browser configuration, or they have specified an unsupported language, then this language will be used.  Typically, this setting applies to search engines.'); ?>

__76 - functions.php__
return WT_I18N::translate('nnMissing PHP temporary folder');

begining with "e"

__413 - admin_site_upgrade.php__
echo '<li>', WT_I18N::translate('ePlace the website offline, by creating the file %s…', $lock_file_html);

fisharebest:
My computer was running very slowly.  I typed into one window, but the text went to the other :-(

vytux-com:
Unfortunately this has created errors in the .po files due to duplicate entries

PR #412, opened 2015-02-01
Fix PO Duplicate entries caused by #405

vytux-com:


fisharebest:
Thanks for doing this!

PR #420, opened 2015-02-04
Added puphpet box configuration for apache2 developer VM

WorkingDevel:
For development with webtrees it is nice to have a configured webserver running out of the box on your current devevelopment workspace. Virtual machines are a good way (and may be the best) for this especially under windows. https://puphpet.com/ is providing a configuration builder for such webserver-VMs.
I have created a configuration for a apache2 webserver that serves the webtrees code from workspace. PhpStorm is also able to control the created vagrant box.

At the moment the configuration is only for users of a Linux system, for Windows it is still needed to create a configuration, but it could be derived from the linux configuration as https://puphpet.com/ can import and modify already created puphpet confiugration files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,11 +2,16 @@
 
 .gitattributes export-ignore
 .gitignore     export-ignore
+.gitmodules    export-ignore
 .travis.yml    export-ignore
 Makefile       export-ignore
 build/         export-ignore
+tests/         export-ignore
+tools/         export-ignore
 js/*.patch     export-ignore
 phpunit.xml    export-ignore
+VAGRANT.sh     export-ignore
+README-vagrant.md                                              export-ignore
 library/bombayworks/zendframework1/library/Zend/Acl/           export-ignore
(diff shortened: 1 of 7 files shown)


fisharebest:
This is a lot of code (25MB, 7500 files).  Does it all need to be in the webtrees repository?

WorkingDevel:
Maybe it could be in a separate repo and checked out via the scripts I wrote. I'll try.

WorkingDevel:
@fisharebest I have created a git submodule version. So no one needs to download the 25Mb if he doesn't want to use vagrant for development.

fisharebest:
I have never used vagrant (although one of my colleagues spent a lot of time failing to get it working!).

How would I fire it up?

WorkingDevel:
There's a readme, that describes it: https://github.com/WorkingDevel/webtrees-vagrant/blob/master/README.md
This readme should be here if you decide to go with the vagrant option.

fisharebest:
I've added a Vagrantfile.  It uses PHP 5.3 - which we need for testing and `composer update`.

PR #424, opened 2015-02-04
Applied fixes to satisfy 'composer validate'

WorkingDevel:
Run of 'composer validate' required to add version strings in 'require'. Added them to match current used versions.

Besides of getting validate passing it's better to upgrade lib major versions 'by hand' via modifying composer.json intentionally than upgrade by accident through composer update.

Added also vagrant files to .gitignore in root directory. If vagrant/puphpet is introduced, at least 2 configurations (linux and windows) should be provided, but it is not possible to have both in the same directory. So the configs should be in sub directories and the one needed should be copied by the developer into the project root, but not committed -> so .gitignore them in root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 *.swp
 .idea/
 vendor/.git
+/Vagrantfile
+/puphpet/
+.vagrant/
--- a/composer.json
+++ b/composer.json
@@ -23,15 +23,15 @@
 		}
 	},
 	"require": {
-		"bombayworks/zendframework1": "*",
-		"ezyang/htmlpurifier": "*",
-		"fisharebest/ext-calendar": "*",
-		"ircmaxell/password-compat": "*",
-		"michelf/php-markdown": "*",
-		"patchwork/utf8": "*",
+		"bombayworks/zendframework1": "1.*",
+		"ezyang/htmlpurifier": "4.6.*",
+		"fisharebest/ext-calendar": "1.3.*",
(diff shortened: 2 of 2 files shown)


fisharebest:
> it's better to upgrade lib major versions 'by hand' via modifying composer.json intentionally than upgrade by accident through composer update.

I run `composer update` most days.  With this change, I would never learn about updates.

Isn't this the purpose of `composer.lock`?

fisharebest:
PS - you will notice that the vendor libraries are included in the webtrees repository :-(

I know this is not good, but we have many contributors (testers, translators), who know how to download a .ZIP from github, but would not be able to install/run composer.

Maybe with the latest refactoring, it is time to remove them....

WorkingDevel:
The `composer.lock` is updated by `composer update` to the latest version of a library that fits the version condition in `composer.json`. If there is no version constraint in the .json, it will upgrade also to the next major version of a library and that maybe breaks the functionality as the library could change their API.
The `composer.lock` unfortunately does not really lock the versions it just locks the versions for `composer install`.
Maybe at least the major version should be constraint. For update checking you could just remove the version constraint in the .json.

Maybe we can provide a script that downloads composer and runs it as gradle is doing it with its gradlew.sh and gradlew.bat scripts.

fisharebest:
I've added version numbers to composer.json.

After the next release, I will look again at removing the vendor code, and teaching users how to run composer. (Or simply providing a `vendor.zip` file for them...)

PR #425, opened 2015-02-04
Moved global var $WT_SESSION as public static to Globals class

WorkingDevel:
To get rid off global variables it is maybe as a first step to move the global variables to a class as public static class variable.
Advantage:
* no 'global' declaration in functions needed (less chance to forget)
* easier to locate via search
* PhpStorm can give code completion and propose globals on class
Disadvantage:
* more to write `Globals::$WT_SESSION` instead of `$WT_SESSION`.

I did it just for the `$WT_SESSION` the other could follow if proposal is accepted.

@fisharebest What do you think?

### DIFF
--- a/action.php
+++ b/action.php
@@ -18,12 +18,6 @@ namespace Webtrees;
 
 use Zend_Session;
 
-/**
- * Defined in session.php
- *
- * @global Zend_Session $WT_SESSION
- */
-
 define('WT_SCRIPT_NAME', 'action.php');
 require './includes/session.php';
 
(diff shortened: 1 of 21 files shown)


fisharebest:
> What do you think?

Many of the global variables are used in horrible ways. (e.g. read and write the same variable in many different contexts).

I am untangling them slowly.  The new `@global` annotations are temporary, to help me do this.  (PhpStorm can use them for code-completion).

Many are attributes of a tree `$tree->getPreference(...)`.  Since genealogy objects always belong to a tree, we can access them via  `$individual->tree()->getPreference()`.  This will help us fix a decade-old bug that affects privacy calculations when you display individuals from more than one tree on the same page.  Using a static/global class for this wouldn't help this problem.

Many others are attributes of a theme.  These were originally set as globals in `theme.php`, and are now (temporarily) available in a theme::getParameter() function.

This leaves a small number of truly global objects - session, request, response.  The long-term plan is to move to a "proper" PHP framework.  Probably something symfony-based (such as laravel?).  Definitely not Zend.    So, the solution for these is likely to be to use the symfony foundation class objects for these.

fisharebest:
The session handler/global has been replaced by a class (which has the same interface as the laravel session handler).

PR #445, opened 2015-02-20
Remove global variable usage and other chart work

ddrury:
Removed global variables from charts and chart related functions (either pass variable as a parameter or obtain from controller)

Most charts - remove unused or broken parameters from forms, remove unused variables from controllers

Pedigree chart remove chart and canvas css offset and remove most (but not all) embedded 'magic numbers'


fisharebest:
Thanks for this!

makitso:
Dave,  just a FYI, your latest change has impacted the printing of the pedigree chart.

ddrury:
How are you printing it? Any particular browser?

makitso:
I am using FF on Ubuntu and doing a preview.

You need to remove the top from the @media print

     #pedigree_chart {
         left: -40px;
         top:  -260px;
     }

ddrury:
Good catch - missed that one

PR #465, opened 2015-03-07
Further pedigree chart work

ddrury:
Moved all positioning logic into the controller. Now the only elements that are absolutely positioned are the node containers. Simplified the line drawing logic in javascript by using sosa triples (ie child + parents)  and using the node location as the base for calculating the point positions.

This has no impact on Meliza's requirements

### DIFF
--- a/app/Controller/PedigreeController.php
+++ b/app/Controller/PedigreeController.php
@@ -41,14 +41,20 @@ class PedigreeController extends ChartController {
 	/** @var integer Number of generation to display */
 	public $generations;
 
-	/** @var array are there ancestors for people byond the extent of the chart */
-	public $ancestors = array();
+	/** @var array data pertaining to each chart node */
+	public $nodes = array();
 
 	/** @var integer Number of nodes in the chart */
 	public $treesize;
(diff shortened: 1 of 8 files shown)


fisharebest:
1) Instead of creating new global functions `get_menu()` and `gotoPreviousGen()`, would these be better as functions within the controller?

2) Do we need to create the global JS variable `WT_PEDIGREE_CHART`.  Can we use just `(function() {...})();`

ddrury:
Re you queries below

1. I wanted too but thought as they generated output they shouldn't go 
in the controller (I'll get the hang of this MVC stuff one day).

2. I think this was one of my changes but not recently - The original 
intention was to have it as an instantiated object with parameters 
passed into it, but it never worked out like that.

I'll make the changes and push then up.


Regards
David

fisharebest:
@ddrury - The "pedigre map" chart (in the google maps module) uses the PedigreeController->ancestors array - which no longer exists.

Can you take a look?

ddrury:
Oops, Ok

Regards
David

PR #468, opened 2015-03-09
Bug fix

ddrury:


ddrury:
Actually hang fire on this - I think using the pedigree controller is not the right thing to do here!

PR #469, opened 2015-03-09
Bug fix

ddrury:
Changed use of PedigreeController to ChartController as the sosaAncestors method is all that is needed


fisharebest:
Thanks!

PR #477, opened 2015-03-13
Fix English typos

vytux-com:


fisharebest:
For changes like this, it is also necessary to modify all the `msgid` entries in the `*.po` files.

Otherwise, the existing translations will be lost.

I'll do this...

vytux-com:
I normally would have done that... but I wasn't sure worth the new translation system :-)

PR #482, opened 2015-03-14
Dutch language file updated.

JustCarmen:


fisharebest:
Hi Carmen.  Have you seen the new translation server http://translate.webtrees.net

As well as editing on line, you can also upload/download .PO files.

I have added these updates using the new service.

JustCarmen:
I missed it, but I've registered now. But I don't see a way to upload translations. Only a download button is available.

fisharebest:
untitled

JustCarmen:
Found it. Thanks.

Art4:
@fisharebest Will it be possible to add translation through pull requests too or only via weblate?

fisharebest:
> Will it be possible to add translation through pull requests too or only via weblate?

Yes - but I would prefer people to use weblate.

If weblate sees that there have been changes made externally, it will re-import all the .PO files.
This is very slow.  It takes 45-60 minutes, and uses 100% CPU on the server.

The only time we **must** change the .PO files is when we make corrections to the English text.  For example, I am changing all the "%d" to "%s", so that we can correctly translate languages that do not use digits 0123456789.

If I did not do this, we would lose the existing translations.

PR #483, opened 2015-03-14
Fix duplicate translation

vytux-com:
It also seems that weblate does not support (or is not configured for) single line translations as we used to prefer....

### DIFF
--- a/language/af.po
+++ b/language/af.po
@@ -199,13 +199,6 @@ msgid_plural "%s logged-in users"
 msgstr[0] ""
 msgstr[1] ""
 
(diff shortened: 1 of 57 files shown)


fisharebest:
If I had read my emails in the reverse order, I would have seen this PR before I made the exact same updates...

fisharebest:
> It also seems that weblate does not support (or is not configured for) single line translations as we used to prefer....

I have installed a pre-commit hook which should do this.

It is a problem, because it prevents us from making trivial corrections (e.g. spelling, punctuation) in the English text.

It also causes simple changes to have effects on many lines, which increases the size of the repository.

PR #493, opened 2015-03-18
Javascript error in autocomplete function - TypeError: $ is not a functi...

JustCarmen:
...on

### DIFF
--- a/assets/js-1.7.0/webtrees.js
+++ b/assets/js-1.7.0/webtrees.js
@@ -1170,8 +1170,8 @@ function autocomplete(selector) {
 	}
 
 	jQuery(selector).each(function() {
-		var type = $(this).data("autocomplete-type"); // What type of field
-		var ged  = $(this).data("autocomplete-ged"); // Which family tree
+		var type = jQuery(this).data("autocomplete-type"); // What type of field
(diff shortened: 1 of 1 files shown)


fisharebest:
A long time ago (in PhpGedView), there were two libraries which both used `$()`.

To fix these conflicts, I changed all the jQuery calls to use the long version, `jQuery()`.

Now, we only have jQuery, and there should be no conflicts.  Hence both `$()` and `jQuery()` should work equally.

I am happy to accept this pull-request (for consistency), but I am curious to learn why it is not working.  Do you have another jQuery library in your module?

JustCarmen:
Only in the colors or clouds theme I get the javascript error "$ is not a function". Since I thought we always use jQuery in stead of $ and changing $ in jQuery did solve the problem, I didn't look any further.

But I did now and I found out that I myself had in one of my scripts an "$" in stead of "jQuery". Changing that also solves the problem. So in that case there is indeed no need to change the webtrees.js file and I can solve it within the module.

Apparently the $ in my script and the $ in webtrees.js were in conflict with each other. I recall I have run into this issue once before in a conflict with a module of someone else. So in the module environment it still is the safest to use jQuery in stead of $ and maybe that counts for the core too, but that is up to you to decide.

PR #494, opened 2015-03-19
Use fisharebest/localisation...

ddrury:
and move label outside input to allow use by assistive technologies


fisharebest:
FYI, the comment about assistive technologies is incorrect.  It was a bug in the validator, which I have since passed to the maintainer of the validation tool.

fisharebest:
PS in the admin/languages page, I use CSS columns (rather than HTML tables) for this.  It works well (on supported browsers!)

ddrury:
Do you want me to do something similar?

Regards
David

PR #496, opened 2015-03-19
function edit_language_checkboxes changed markup...

ddrury:
...from TABLE to DIV and SPAN and set CSS display properties to display as table. (didn't use CSS columns as IE only supports them from v10 onwards)


ddrury:
Withdrawn; superceded by 246069e19df9f6df62dcf3e62c23cefc3e3c6a3a

PR #499, opened 2015-03-20
Lightweight markdown and HTML editor and news module to use it

ddrury:
Herewith 2 modules: markItUp an implementation of Jay Salvats lightweight editor (see http://markitup.jaysalvat.com/home/ for details) and blog, a module to demonstrate its use.

blog can replace both the FamilyTreeNews & UserJournal modules (and at a pinch the html module). It provides language selection on the Family Tree page and allows for any of the webtrees stats tags to be embedded in an item.

it is currently written to be completely self contained (but see below) and therefore is not ready for production use although the necessary changes are trivial.

When enabled it adds a languages field to the news table, bumps the NB_SCHEMA_VERSION in the site_settings table and changes the preferences format, so if removed it will be necessary to revert those changes.


vytux-com:
Shouldn't these be better as custom modules especially with the planned changes in the 1.8 version?

ddrury:
Withdrawn due to module class changes

fisharebest:
This looked interesting, but I haven't yet found time to look at it.

ddrury:
Thanks,

I'll maybe resubmit them when the module classes have stabilised

Regards
David

fisharebest:
> when the module classes have stabilised

There will be another significant change soon.

Specifically, the namespaces will change to allow PSR-4 autoloading.  Currently, webtrees can only use a classmap autoloader.

PR #519, opened 2015-03-27
Rework lifespan chart

ddrury:
* Made look and feel more like other charts
* Works for RTL languages
* Simplified chart by removing the pan/scroll arrows and replaced with scrollbars
* Removed birth and death markers and replaced with a popup showing key events when the sex-image/name/lifespan on the bar is hovered over - this also replaces the existing (non functional) code which should have placed key event markers (and a popup) along the length of the bar.
* popups use code structure found elsewhere in webtrees
* Generate bar colors on the fly using the hsla model (doesn't work for ie8 which gets pink for a girl, blue for a boy)
* For short lifespans sensibly trucate the info on the bar


fisharebest:
It certainly looks much nicer!

Is the intention to remove horizontal scrolling, or is it still being worked on?

ddrury:
Hm,

I hadn't planned to remove the horizontal scroll. Take for example 
searching on (a popular) place, you're likely to get a lot of bars over 
a long time-frame. If it were scaled, the bars would be too small to be 
readable.

Any suggestions?

Regards
David



ddrury:
Withdrawn pending further work on calendars

PR #526, opened 2015-03-29
fixed gender pronouns from exclusively male

dmangot:
All the pronouns were male, but users of webtrees are not exclusively maIe.   I tried to give my daughter and account to edit our family tree and it said 'his', even though she is not male.  This pull request aims to fix those conditions.   Thanks!

### DIFF
--- a/admin_site_config.php
+++ b/admin_site_config.php
@@ -471,7 +471,7 @@ $controller->pageHeader();
 		<div class="col-sm-9">
 			<?php echo edit_field_yes_no('USE_REGISTRATION_MODULE', Site::getPreference('USE_REGISTRATION_MODULE')); ?>
 			<p class="small text-muted">
-				<?php echo /* I18N: Help text for the “Allow visitors to request account registration” site configuration setting */ I18N::translate('Gives visitors the option of registering themselves for an account on the website.<br><br>The visitor will receive an email message with a code to verify his application for an account.  After verification, an administrator will have to approve the registration before it becomes active.'); ?>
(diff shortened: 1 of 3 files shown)


fisharebest:
I agree with the sentiment of this change, but will implement it slightly differently.  At present, it will cause all 60 current translations to become untranslated, and require that all translators re-translate it.

Since the change does not affect the meaning, and since most languages have already translated this using their own gender-neutral pronoun, I'll also need to update all the existing translation files.

Incidently, I've had complaints the other way, from linguists who tell me that "their" is plural, and should not be used as a singular...  There may also be differences between british and american usage. See https://en.wikipedia.org/wiki/Singular_they

fisharebest:
When this

> Add this individual, his direct line ancestors, and their families.

is changed to this

> Add this individual, their direct line ancestors, and their families.

we are using "their" as both singular and plural in the same sentence.  It becomes unclear whether the second "their" is the same as the first "their"

These sentences probably need rewriting completely.

dmangot:
Thanks Greg.   If there are translations files that need to be edited or rewriting of some sentences that needs to be done, I'm happy to put in some time for that and offload you.   Just let me know specifically how I can help and I'll modify the PR.

Cheers.

vytux-com:
Also for future reference you should not modify the vendor files as they are 3rd party and changes would be overridden by any future updates.

fisharebest:
For the clippings cart, we actually know the individual, so we can accurately use "his" or "her".

For the help text, it needs rewriting completely.

PR #552, opened 2015-04-07
Added defaultCalendar function

ddrury:
As per email conversation

### DIFF
--- a/app/I18N.php
+++ b/app/I18N.php
@@ -33,6 +33,11 @@ class I18N {
 	/** @var Translator */
 	private static $translator;
 
+	/**
+	 * @var string calendar (e.g @#DGREGORIAN@)
+	 */
+	private static $default_calendar;
+
 	// Digits are always rendered LTR, even in RTL text.
 	const DIGITS = '0123456789٠١٢٣٤٥٦٧٨٩۰۱۲۳۴۵۶۷۸۹';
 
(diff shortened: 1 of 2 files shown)


fisharebest:
Thanks.  After our discussion about the public `$d`, `$m` and `$y` member variables, I've been looking at shifting more of the functionality into the `ext-calendar` library.

(The advantage being that these external libraries are much easier to test in isolation, and hence have full unit-tests, etc.)

So, I've changed the API to return a calendar object instead of a calendar escape string, which should work better with these changes.  Instead of creating a GregorianDate, you would create a CalendarDate, and pass a GregorianCalendar object in the constructor...

ddrury:
Well,

The chart now works with RTL languages and displays the scale with 
native characters, a problem was that the length of the year character 
string in Hebrew is longer than that of an Arabic language so I've had 
to change the scaling to suit. I don't suppose you happen to know which 
language would likely produce the longest strings would you?

Regards
David

fisharebest:
Hebrew is likely to be the longest, as it uses a "roman numerals" style, with variable lengths.

All other languages have a simple substitution of digits 0-9

ddrury:
Sorry about this - another question

I can't get my head around this locale/language/I18N inter-relationship.

If I want to display a number in the local language, how do I do it.

i.e. something like

$people = count($controller->people);
echo  I18N::number($people); // doesn't work!!


Regards
David

fisharebest:
```php
$people = count($controller->people);
echo  I18N::number($people); // doesn't work!!
```

It should work.

```php
echo  I18N::number(1234); // should give "1,234" (in English locales)
```

If you want hebrew numerals, then these are only used for dates (not for general numbers), and hence are only accessibly by taking a JewishDate object containing that year, and formatting it with (from memory) `->format('y')`

ddrury:
Oh well that makes it simple, it was a simple number (not a date) so I 
don't have to worry!

Regards
David

PR #566, opened 2015-04-25
Rework lifespan chart

ddrury:
Made look and feel more like other charts
Works for RTL languages
Added support for different calendars
Simplified chart by removing the pan/scroll arrows and replaced with scrollbars
Removed birth and death markers and replaced with a popup showing key events when the sex-image/name/lifespan on the bar is hovered over - this also replaces the existing (non functional) code which should have placed key event markers (and a popup) along the length of the bar.
popups use code structure found elsewhere in webtrees
Generate bar colors on the fly using the hsla model (doesn't work for ie8 which gets pink for a girl,
blue for a boy)
For short lifespans sensibly trucate the info on the bar
Hide form on print


ddrury:
Withdrawn - bug in calendar working

PR #600, opened 2015-06-07
Rework lifespan chart

ddrury:
Made look and feel more like other charts.
Works for RTL languages.
Added support for different calendars.
Simplified chart by removing the pan/scroll arrows and replace with scrollbars.
Removed birth and death markers and replaced with a popup  showing key events (vital events plus other events that meets the search criteria) when the sex-image/name/lifespan on the bar is hovered over - this also replaces the existing (non functional) code which should have placed key event markers (and a popup) along the length of the bar.
popups use code structure found elsewhere in webtrees.
New file ColorGenerator.php generates bar colors on the fly using the hsla model (doesn't work for ie8 which gets pink for a girl, blue for a boy)
For short lifespans sensibly trucate the info on the bar.
Hide form on print.
If searching a place, include marriage events.


fisharebest:
Merged

PR #602, opened 2015-06-08
This little piece of code seems to have gone missing somewhere in the…

ddrury:
… transition

### DIFF
--- a/lifespan.php
+++ b/lifespan.php
@@ -54,7 +54,7 @@ $controller
 					<?php echo I18N::translate('Include the individual’s immediate family?'); ?>
 				</label>
 				<span class="optionbox">
-					<input id="addFamily" type="checkbox" value="yes" name="addFamily">
+					<input id="addFamily" type="checkbox" value="yes" name="addFamily" <?php echo $controller->showDetails;?>>
 				</span>
 			</div>
 		</fieldset>


fisharebest:
Sure - but I'm not sure I understand the connection between the `PEDIGREE_FULL_DETAILS` setting and the "include family members" option.

ddrury:
Yeah,

I tend to agree but rather than add a config variable I thought "include 
family members" could be (loosely)  interpreted as "Full Detaiils"

Regards
David



fisharebest:
Maybe it would be better leave this box unticked by default, and allow the user to tick it when needed?

I don't use this chart much, so I don't have a feel for how it is used.

ddrury:
I don't have strong feelings one way or the other - in the original it 
was always set to I guess we could revert to that behaviour (or leave it 
unset - as you think best)

Regards
David



fisharebest:
If it is unset by mistake, then the user can simply select it again.
If it is set by mistake, then the user needs to start over to remove the unwanted family members.

I don't think it follows that if a user selects it for one individual, they will necessarily want it selected for the next individual.

So, unless anyone has other opinions, I'd think that the natural (and least confusing?) option is not to set the default, and leave it unticked always.

ddrury:
Fine by me

PR #608, opened 2015-06-12
Add http 403 if person exists but cannot be seen by this viewer.

### LINKED ISSUE #607: Private should give 403 instead of 404
If an individuals' record is private, you get a 404 as if it didn't exist. This is a problem because there may be links to it. So Google will see that you habe broken links (link to 404 = broken). 

403 Forbidden would be better, because the link is not considered broken anymore. It is a small fix in session.php, pull request comming in a few hours.

bmhm:
Adjust error messages accordingly. fixes #607.
Also does a little code cleanup. "else if" is easier to track if there are more then two elses.

### DIFF
--- a/individual.php
+++ b/individual.php
@@ -80,12 +80,18 @@ if ($controller->record && $controller->record->canShow()) {
 				'</p>';
 		}
 	}
-} elseif ($controller->record && $controller->record->canShowName()) {
+} else if ($controller->record && $controller->record->canShowName()) {
 	// Just show the name.
 	$controller->pageHeader();
 	echo '<h2>', $controller->record->getFullName(), '</h2>';
 	echo '<p class="ui-state-highlight">', I18N::translate('The details of this individual are private.'), '</p>';
 
+	return;
+} else if ($controller->record && !($controller->record->canShowName())) {
+	http_response_code(403);
+	$controller->pageHeader();
+	echo '<p class="ui-state-error">', I18N::translate('You do not have permission to view this individual.'), '</p>';
+	
 	return;
 } else {
 	http_response_code(404);


fisharebest:
Historically, we had separate 404 and 403 pages.  There was concern that this allowed people to determine the existance of private individuals.

As a result, the two error messages were combined into a single "This record does not exist, or you do not have permission to see it" error.

So, while separate 403 and 404 responses may be technically correct, I'm not sure if the user community would want it.

PS - note that the PSR-2 coding style prefers `elseif` to 'else if`.

bmhm:
>  There was concern that this allowed people to determine the existance of private individuals.
I think this is already solved by "Show private relationships". If set to no, there are no 404 pages we are linking to.  If set to yes, we already know there is an individual with this xref and don't need to take care about hiding already shown information, do we?

About PSR-2: I will fix this.

fisharebest:
> There was concern that this allowed people to determine the existance of private individuals.

Suppose a family has 4 children, with XREFs of I101, I102, I103 and I104.

Suppose you decide to make the second child private.

A visitor would see three records: I101, I103 ad I104.  If they could also see that I102 is a private record (403) as opposed to just a gap in the sequence (404), then there was concern that a visitor might be able to infer the existance of this second child.

Arguments such as these were used to justify the removal of the original 403 pages.

bmhm:
This would imply they were created in the correct order. This is not that often the case. So it turns out a numbering gap is only a guess, not evidence. I102 could be a totally different person.

If it were hard evidence, I'd agree. But I'm not 100% convinced right now, although I can understand why this helps (a little bit).

fisharebest:
Just been reading the HTTP specification....

## 403

> The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated. If the request method was not HEAD and the server wishes to make public why the request has not been fulfilled, it SHOULD describe the reason for the refusal in the entity. If the server does not wish to make this information available to the client, the status code 404 (Not Found) can be used instead. 

This suggests that since we do not want to give a reason for refusing the request, then we may use 404 instead.

bmhm:
> that since we do not want to give a reason
I do. Ok, I'll try to find a reasonable solution, because links to 404 are never good (in my opinion). And 404 when not logged in and 200 when logged in is bogus, too.

Another suggestions: Can we make it configurable (403 vs 404)? Or set no links if individual is private (but still do show them). Would that be an option?

fisharebest:
> because links to 404 are never good

We should never generate links to pages that give 404 errors.

If this is happenening anywhere, then this is a bug.

All that can happen is that a user will login, bookmark a link, then logout, then visit the link.

bmhm:
| All that can happen is that a user will login, bookmark a link, then logout, then visit the link.

Yes it does happen. Enable privacy for living descendants, but choose "Show relations" i.e. uncheck hide relations to private individuals.

Now webtrees will happily generate links to "private individuals". Those are 404'd.

bmhm:
Still happens in v1.7.2. I'd be happy to see this in webtrees (perhaps configurable), as this function cannot be modularized.
So for SEO you might want to chose a 403 error for linked, but completely unviewable individuals. But if your concern is privacy (don't guess individuals from numbers), it should be 404. Both options are reasonable,
Still, I don't think you can guess *relatives* from counting up numbers, because it might be completely random which person you add next.

fisharebest:
Suppose someone in my tree requests that I remove them.

Since I do not want to delete my research data, I add a privacy restriction so that only the administrator (me) can view it.

The person now sees a 404 - and is happy.  But with this change, they would see a 403, and know that I hadn't actually deleted the data.

We've fixed the bug where "private" boxes on charts included links.  If you find any other links being generated to private individuals, please report them.

PR #609, opened 2015-06-13
Family cleanup

bmhm:


fisharebest:
The logic for changing theme/language has been replaced with AJAX calls.

PR #623, opened 2015-06-20
Update Menu.php

magicsunday:
Fixed fatal error: Call to a member function getMenuAsList() on a non-object in \app\Menu.php on line 273

### DIFF
--- a/app/Menu.php
+++ b/app/Menu.php
@@ -266,10 +266,12 @@ class Menu {
 			$link = '';
 		}
 		$html = '<a' . $link . $attrs . '>' . $this->label . '</a>';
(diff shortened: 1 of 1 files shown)


fisharebest:
`__construct()` only takes an array.
`setSubmenus()` only takes an array.

`Menu::submenus` should always be an array, so I don't see how this check could help.

The error message suggests that one the array elements is not a Menu object.

Most likely, there is an array of menus containing `null`.  Elsewhere we filter these using `array_filter()`

magicsunday:
Ok. The error occurs due the last changes of the base classes if a module has not been updated. So maybe it could ignored.

PR #624, opened 2015-06-20
Update reportengine.php

magicsunday:
Handle given precision parameter in reports engine xml files.

### DIFF
--- a/reportengine.php
+++ b/reportengine.php
@@ -215,8 +215,8 @@ case 'setup':
 			foreach ($options as $option) {
 				$opt                   = explode('=>', $option);
 				list($value, $display) = $opt;
-                		if (preg_match('/^I18N::number\((.+?)(,([\d+]))?\)$/', $display, $match)) {
-                    			$display = I18N::number($match[1], isset($match[3]) ? $match[3] : 0);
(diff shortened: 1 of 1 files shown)


reportengine.php
```diff
@@ -215,8 +215,8 @@
 			foreach ($options as $option) {
 				$opt                   = explode('=>', $option);
 				list($value, $display) = $opt;
-				if (preg_match('/^I18N::number\((.+)\)$/', $display, $match)) {
-					$display = I18N::number($match[1]);
+                		if (preg_match('/^I18N::number\((.+?)(,([\d+]))?\)$/', $display, $match)) {
+                    			$display = I18N::number($match[1], isset($match[3]) ? $match[3] : 0);
```
fisharebest: Can you convert spaces to tabs for indentation.
http://wiki.webtrees.net/en/Coding_Standards

Maybe make the second pattern non-capturing.  e.g. `(,` becomes `(?:,` - or maybe even `(?:, *`

Amgine0:
Nice! but are not differences to the standard normally stated as EXCEPT?

reportengine.php
magicsunday: uh.. I did the change just here in github.

magicsunday:
Please take another look at this. I changed the spaces to tabs.

PR #660, opened 2015-07-25
Closing div without opening tag in Search page

jon48:
There is a closing div tag trailing at the end of the search page, which can cause issue with theme, as it may close a unwanted div from the theme header.


fisharebest:
I think we are also missing `</div>` in `SearchController::printResults()`...

jon48:
I have done some further changes to the div structure (actually reintroducing the trailing `</div>`... ).
There was indeed a missing `</div>`in `SearchController::printResults()` in the case there is a result to display (I was testing without any result).
I have as well factorised the `<div id="search-page">`, and include the results div within it (as I think it was intended to be)

PR #677, opened 2015-08-10
Fix quotes

vytux-com:


fisharebest:
It would be much better if you could submit these changes to the translation server.

(TIP: if it is easier, you can upload a .PO file to the translation server).

The translation server assumes that it will be responsible for all translations.  If translations are changed via github, then I must re-import all translations, which takes one hour, and prevents other translators for working for this time.

I have copied/pasted these into the translation server.

Issue #679, opened 2015-08-11
User (non-administrator) unable to delete own account


exxamalte:
As a non-administrator user you are supposed to be able to delete your own user account from the 'My Account' page. This does not work currently. 
After confirming the deletion of the own user account, the user is logged out, but the account remains untouched.
Turns out that the code that is trying to delete the user account first logs out the user and afterwards tries to use a reference of the logged in user (`Auth::user()`) to delete the account. This fails because the reference is then a visitor user. Instead, the code should keep a reference of the logged in user, and use that to delete the account.

PR #680, opened 2015-08-11
New administrator account recognised correctly as admins, and users can delete their own account

### LINKED ISSUE #678: Unable to add new administrator account
The administrator flag has been partially renamed with the following effects:
* The user detail form for an existing administrator (created with webtrees 1.6.x) does not show the `Administrator` checkbox checked.
* Newly created administrator user (created in webtrees 1.7.x) does not actually have admin rights in the system.

It appears that the `canadmin` setting has been renamed to `admin` at some point in time in some code instances.

In my webtrees 1.7.1 instance's database table `wt_user_setting` the admin has a setting `canadmin` set to `1`. And this is apparently what `app/Auth.php#isAdmin` checks for. However, in `admin_users.php` when rendering the checkbox for the user detail form the code looks for a setting `admin` (this setting name has been introduced in https://github.com/fisharebest/webtrees/commit/07b70ae14d34ec3d428f7482e6df3ea0b1e7ef47). 
And when you add a new user with the `Administrator` checkbox checked, this user ends up with a `admin` setting set to `1`, but no `canadmin` setting.

Here the original discussion in the forum: https://www.webtrees.net/index.php/en/forum/help-for-ver-1-7-x/30550-administrator-account-s

exxamalte:
Fixes #678 and #679.

### DIFF
--- a/admin_users.php
+++ b/admin_users.php
@@ -60,7 +60,7 @@ case 'save':
 		$contact_method = Filter::post('contact_method');
 		$comment        = Filter::post('comment');
 		$auto_accept    = Filter::postBool('auto_accept');
-		$admin          = Filter::postBool('admin');
+		$admin          = Filter::postBool('canadmin');
 		$visible_online = Filter::postBool('visible_online');
 		$verified       = Filter::postBool('verified');
 		$approved       = Filter::postBool('approved');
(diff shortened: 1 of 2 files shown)


fisharebest:
Thanks!

PR #692, opened 2015-08-20
Update da.po

jacoline:
Please accept this - annoying to do this everytime


fisharebest:
This file was uploaded directly to the translation server - which will merge/commit the changes.

PR #713, opened 2015-09-04
Pre error messages if modules get disabled on exceptions

bmhm:
Formated log.
To make dataTables render HTML, we need a render function. Also, the html tags are written into the database (escaped), which should hopefully not be a problem.
Also added a style so that the pre tag doesn't destroy the layout.

### DIFF
--- a/admin_site_logs.php
+++ b/admin_site_logs.php
@@ -224,7 +224,23 @@ $controller
 				clear: "fa fa-trash-o"
 			}
 		});
-	');
+	')
+	->addInlineJavascript('
+var renderAsHtml = function (data, type, full, meta) {
+	return decHTMLifEnc(data);
+};
+var isEncHTML = function(str) {
+	if(str.search(/&amp;/g) != -1 || str.search(/&lt;/g) != -1 || str.search(/&gt;/g) != -1)
+		return true;
+	else
+		return false;
+};
+
+var decHTMLifEnc = function(str){
+	if(isEncHTML(str))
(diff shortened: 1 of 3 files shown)


bmhm:
Preview:

webtree-log-exception-128

fisharebest:
IMHO, adding HTML markup to the error messages is very dangerous, as the logs contain user-supplied data.

Why do we need `<pre>`?   If we are trying to retain the link-breaks in a PHP stack dump, then simple css (e.g. `style="white-space:pre-wrap;"` in admin-site-logs.php line ~165) could be all that is needed...

bmhm:
Good idea. I'll do a new PR next day

On Fri, 4 Sep 2015 15:40 Greg Roach <notifications@github.com> wrote:

> IMHO, adding HTML markup to the error messages is very dangerous, as the
> logs contain user-supplied data.
>
> Why do we need <pre>? If we are trying to retain the link-breaks in a PHP
> stack dump, then simple css (e.g. style="white-space:pre-wrap;" in
> admin-site-logs.php line ~165) could be all that is needed...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/fisharebest/webtrees/pull/713#issuecomment-137740572>.
>
-- 
Mit freundlichen Grüßen,
      Benjamin Marwell

      Jöhrensstraße 3
      30559 Hannover

      Telefon: +49 (5 11) 47 41 69 50
      Mobil  : +49 ( 1 70) 7 13 93 17
      E-Mail: E-Post@bmarwell.de
      PGP-Key: FD58 0CC0

PR #729, opened 2015-09-19
Turn batch-update form into bootstrap form layout

JustCarmen:
This will make the batch-update form responsive by using the bootstrap form layout.


fisharebest:
Thanks!

PR #732, opened 2015-09-27
Google maps form in bootstrap layout

JustCarmen:
I've split the changes into multiple commits so it is easier for you to see the changes.

### DIFF
--- a/app/Module/GoogleMapsModule.php
+++ b/app/Module/GoogleMapsModule.php
@@ -4541,40 +4541,44 @@ class GoogleMapsModule extends AbstractModule implements ModuleConfigInterface, 
 			}
 		}
 		</script>
-		<?php
-		echo '<div id="gm_breadcrumb">';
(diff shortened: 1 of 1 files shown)


JustCarmen:
Some checks failed. What should I do to correct this?

fisharebest:
> Some checks failed. What should I do to correct this?

You can ignore this one.  It means that after your change, a smaller % of lines of code were covered by the unit-tests.

I guess your new code simply has more lines than the old code.

fisharebest:
PS:

When a `<label>` encloses an `<input>`, then you do not need the `for="XXX"` and `id="XXX'`.

Some validation tools get this wrong ;-)

JustCarmen:
>> When a <label> encloses an <input>, then you do not need the for="XXX" and id="XXX'
Didn't know that. Do you want me to change it back?

fisharebest:
> Do you want me to change it?

If/when you have some spare time...

PR #749, opened 2015-10-17
Don't seperate the tables. It breaks the layout when adding more fields.

JustCarmen:


makitso:
I can't see where this happens, can you post a screen shot?

JustCarmen:
Without fix:

fireshot screen capture 083

This happens when adding extra fields:

fireshot screen capture 085

With fix and extra fields:

fireshot screen capture 086

The only thing you should consider is adding the extra fields above the father/mother rows or add an extra separator.
But by having all fields in one table you won't get different table widths, which doesn't look nice.

makitso:
The only problem with having all the rows in one table is that you ?Add more fields", it places the new row at the bottom of the Mother.

makitso:
What about this for a layout?
search

JustCarmen:
Looks good. Can you check your latest layout with another language? Because in Dutch for example the extra select field gets wider because of long texts in the option fields.

JustCarmen:
I've seen your changes with commit 4a52c222daf697180be99fdfa2fe62dad67aa7bd, which are fine with me. So I'll withdraw this pull request.

PR #756, opened 2015-10-25
Change type of settings values and make welcome message bigger

wodka:


wodka:
Bigger & Working welcome message

anmeldung_und_registrierung

fisharebest:
FYI - in MySQL, a `TEXT` column can store 65535 bytes (not characters).  `LEFT(x, 65535)` takes the first 65535 characters.  So, if a text contains multi-byte characters, you could still get a fatal error if your text is too long and you are using STRICT mode.

I would prefer not to use a `TEXT` column for performance reasons.

How many characters do you really need?  1000? 2000?

wodka:
yes, makes sense - I think 2000 should work for most cases. (definitely for mine)

JustCarmen:
I use a slightly different text based on one of the defaults. If we look at the default texts we see that the longest text holds about 500 characters. But a custom text may not hold more then 255 characters which is too short to be really useful. Eventually I solved it by adding my custom welcome text to the custom languages files by changing the translations of one of the defaults. Maybe if this is such a problem, you should consider not to offer the option at all (I can't imagine it is used a lot because of its limitation). Otherwise I think a character amount of 2000 will definitely be enough.

fisharebest:
Merged

Issue #744, opened 2015-10-11
XMP face tag support


bobosch:
Face thumbnails improves the ancestors chart. But it's a exhausting work to create all these face thumbnails. Another difficulty is to use always the same aspect ratio to get a consistent look of the charts with images.

I'm using Google's Picasa to manage my pictures. Picasa can automatically mark faces on a photo. If you have enabled the "Store name tags in photo" option
![picasasavetags](https://cloud.githubusercontent.com/assets/625157/10418763/e3ce2b98-7064-11e5-8898-00a32046e7d9.jpg)
Picasa stores this information in the file as XMP tag. The information also includes position, width and height of the face
![facetags](https://cloud.githubusercontent.com/assets/625157/10418778/25e9e29c-7065-11e5-8213-304463f12b58.jpg)
By the way: The selections have always the same AR (and the nose in the middle ;-) )...

It would be a great help, if the media management would use this information to create additional thumbnails for each individual. If the media object is linked to a person, it should display the face thumbnail instead of the thumbnail of the full image.

I've tried to implement this myself, but you need additional information: The mapping of XMP tag to an gedcom individual. I thought this should be stored in the media element
![mediaelement](https://cloud.githubusercontent.com/assets/625157/10418773/1629b3a0-7065-11e5-826a-b508341fd6a8.png)
but I have no idea how to store this information in the gedcom standard...

You can access this data with PHP like
<code>
$content = file_get_contents('family.jpg');
$xmp_data_start = strpos($content, '<x:xmpmeta');
$xmp_data_end   = strpos($content, '</x:xmpmeta>');
$xmp_length     = $xmp_data_end - $xmp_data_start;
$xmp_data       = substr($content, $xmp_data_start, $xmp_length + 12);
</code>

bobosch:
I've wrote a function to create the image thumbnails from a .jpg file
```php
<?php
create_face_thumbnails('data/media/family.jpg');

function create_face_thumbnails($file) {
	$xmp_data = get_xmp_tag_from_file($file);
	
	if($xmp_data) {
		$faces = get_face_data_from_xmp_tag($xmp_data);

		if($faces) {
			$image=imagecreatefromjpeg($file);
			$path_parts=pathinfo($file);
			$base=$path_parts['dirname'].'/'.$path_parts['filename'].' ';
		
			foreach($faces as $name=>$face){
				$crop=get_face_image($image,$face);
				$thumb=imagescale($crop,100);
[...]

PR #757, opened 2015-10-27
Show face as thumbnail if image has a XMP face tag.

bobosch:
This is a proof of concept of feature request #744 . If an image contains an XMP face tag, and the individual name is in this tag, a face thumbnail is created for this individual.


bobosch:
Sorry for the mess, I'm not familiar with github. Please contact me if you like a current pull request, I will create a new one.

Issue #764, opened 2015-11-19
Batch update / Changes log bug?


ddrury:
Using 1.7.3 imported a gedcom, ran batch update "Add missing married names" got a strange changes log whereby (I think) the xref of the first record processed is repeated for all change records and the xrefs of the old & new gedcom records don't match. See attached screen dump
![dump](https://cloud.githubusercontent.com/assets/104515/11272308/cffd9176-8ec3-11e5-8b0f-9d3072f92278.png)

Further: just ran same process on v1.6.2 and got same output

OK, I can see the problem:

BatchUpdateModule.php about line 142 - the 'update_all' case uses $this->xref which never changes - I've changed all occurrences to plain $xref and this fixes things, therefore for the 'update all' line in function BaseUpdatePlugin/getActionButtons does a xref need to be included?

fisharebest:
Fixed via #765

PR #765, opened 2015-11-21
Fixes problems with updating records when using the update all button

ddrury:
As per issue #764. It was simpler to leave the parameters of function getActionButtons(...) alone for the 'Update all' case as it affects the test in function createSubmitButton for enabling or disabling the button.


fisharebest:
I haven't managed to test this - but the fix "looks right".

PR #766, opened 2015-11-24
Unified before/After changes list

ddrury:
Many moons ago I created a simplified version of this using a 3rd party script and posted an image on the forum. It didn't go down well. Stephen (Toyguy) felt it made it difficult to use cut and paste when a change needed modifying. I think this submission addresses that whilst still making the actual change very visible.

Use Fisharebest\Algorithm\MyersDiff to create a unified before/after display in the style of the batch update facility (more or less duplicating code found in BatchUpdateBasePlugin functions getActionPreview, decorateInsertedText, decorateDeletedText and createEditLinks). In addition added an undo icon for pending and accepted changes

### DIFF
--- a/admin_site_change.php
+++ b/admin_site_change.php
@@ -15,6 +15,7 @@
  */
 namespace Fisharebest\Webtrees;
 
+use Fisharebest\Algorithm\MyersDiff;
 use Fisharebest\Webtrees\Controller\PageController;
 use Fisharebest\Webtrees\Functions\FunctionsEdit;
 use PDO;
@@ -50,7 +51,6 @@ $search = Filter::get('search');
 $search = isset($search['value']) ? $search['value'] : null;
 
 $statuses = array(
-	''         => '',
 	'accepted' => /* I18N: the status of an edit accepted/rejected/pending */ I18N::translate('accepted'),
 	'rejected' => /* I18N: the status of an edit accepted/rejected/pending */ I18N::translate('rejected'),
 	'pending'  => /* I18N: the status of an edit accepted/rejected/pending */ I18N::translate('pending'),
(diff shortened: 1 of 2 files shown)


fisharebest:
Hi Dave,

I like this.  I think there might be a few edge-cases that need looking at first.

* I think that we should only be able to undo the latest change - i.e. where the "new" data is the same as the "current" data.  Otherwise the action (revert to the state before this edit) is not quite what the user asked for (revert just this edit).
* What about insert/delete?  Should we also delete/insert?

Instead of the "delete icon", perhaps a text button "undo" would be easier to understand?

Also, would it make sense to combine the "Status" and "Undo" columns.  After all, we are undoing the "accept" action.

fisharebest:
I guess we could also implement "redo" for rejected changes.  As before, this would only make sense where the "old" data is equal to the current data.

And while we are at it, for complete control, we could include accept/reject actions for pending changes ;-)

ddrury:
Blimey!

I originally had a button instead of the delete icon but changed it to save a bit of space but given your other suggestions it would make sense to have an "action" column where the button label is context sensitive. 

I don't understand how to combine the status and undo columns - are you saying don't show the current status?

Is there a function for comparing two gedcoms? and would doing that not affect performance when the change table is large.

I'll need to get my head around the current / next status sequences

Anyway I'll have a go.

fisharebest:
> I don't understand how to combine the status and undo columns - are you saying don't show the current status?

I meant just put both things in the same column.  i.e. put the undo button in the "status" column, underneath "accepted" status.

ddrury:
> I meant just put both things in the same column

Ah, Gotcha

ddrury:
Just started playing with this. Need some advice on visuals. For those records where a button isn't appropriate would you prefer a disabled button (don't know what text would be on it) or nothing at all

ddrury:
Hi Greg,

I think I'm just about there but I can't get my head around one thing - is it possible to make a change and accept it when there is an earlier change still pending?

ddrury:
OK,
I've resubmitted with all (I think) the changes you suggested. I'm not wholly convinced with the text on the buttons - please feel free to change then if you can think of anything better.

ddrury:
The more I thought about it the more convinced I was that changing the status of a record was wrong so I removed the code for that

ddrury:
Withdrawn pending resolution of a problem with addition of a new record or deletion of a record

PR #772, opened 2015-12-03
Unified before /after display for changes log

ddrury:
Use Fisharebest\Algorithm\MyersDiff to create a unified before/after display in the style of the batch update facility (more or less duplicating code found in BatchUpdateBasePlugin functions getActionPreview, decorateInsertedText, decorateDeletedText and createEditLinks). In addition added an undo icon for pending and accepted changes

Simplify selectEditControl code for statuses
Fix datetimepicker not initialising with earliest and latest dates
Simplified version without undo / redo buttons (too many cases to catch and handle)

### DIFF
--- a/admin_site_change.php
+++ b/admin_site_change.php
@@ -15,6 +15,7 @@
  */
 namespace Fisharebest\Webtrees;
 
+use Fisharebest\Algorithm\MyersDiff;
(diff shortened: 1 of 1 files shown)


fisharebest:
Sorry it has taken me so long to look at this.  Two jobs plus lots of projects....

PR #775, opened 2015-12-07
Update TopSurnamesModule.php

tunandras:
I hope it’s technically alright.

### DIFF
--- a/app/Module/TopSurnamesModule.php
+++ b/app/Module/TopSurnamesModule.php
@@ -62,6 +62,12 @@ class TopSurnamesModule extends AbstractModule implements ModuleBlockInterface {
 		$COMMON_NAMES_THRESHOLD = $WT_TREE->getPreference('COMMON_NAMES_THRESHOLD');
 
 		$num       = $this->getBlockSetting($block_id, 'num', '10');
+		// The “Minimum number of occurrences” input field from Control Panel’s Preferences page 
+		// has to be copied here otherwise user doesn't understand why the result is different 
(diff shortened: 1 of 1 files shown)


tunandras:
The modified module fixes two issues.

tunandras:
The modified module fixes two issues.

fisharebest:
> The modified module fixes two issues.

What are the two issues?

tunandras:
1. After setting “Number of items to show” input field in “Configure — Top
surnames” form, the user may not understand the result if the “Minimum
number of occurrences” input field on Control Panel’s Preferences page
specifies not the expected.

2. If user sets too high number in “Minimum number of occurrences” field a
short message “No data available in table” appears by
webtrees\packages\datatables-1.10.7\js\jquery.dataTables.min.js JavaScript
file without the possibility of translating (and only if presentation style
is table; all other styles result no message at all)

(I wrote them in comments into the code...)



---
A hálóhelyem: tuna.hu, az Index egyik rovata pedig a Dívány
<http://divany.hu/>.

fisharebest:
I think the problem is that:

a) we have configuration options (`COMMON_NAMES_REMOVE`, etc.) that should only apply to the common surnames list (in the Family tree statistics block).
b) we are using these in the Top surnames block

IMHO, a better solution is to ignore these options in this block. https://github.com/fisharebest/webtrees/commit/784e1b4bc2b28a54307c23676a35e0393f5f7e80

fisharebest:
8e7b29b fixes the untranslated text in empty lists

PR #813, opened 2016-01-22
Config gedcom tags

tronsmit:


fisharebest:
I thought this change was a good idea.  Why close it?  Do you plan to improve/resubmit the code?

tronsmit:
It was an unintended request (I am still trying to get to grips with Git -
and PHP, and ...)

I had not intended to submit for consideration yet, I only wanted to have a
copy of the branch on my remote. I will revise and resubmit soon.

fisharebest:
> I had not intended to submit for consideration yet

OK - I'll let you work on the code some more.  There is no rush.

I was going to ask if there was a reason for the substr(0,10) in the switch statement.

tronsmit:
One can display the same tags for a set of different boxes, eg for families

FAM_FACTS_ADD
FAM_FACTS_QUICK
FAM_FACTS_UNIQUE

Cutting out the first 10 characters, almost worked (the ADVANCED_N,
ADVANCED_P just fit in), but not for QUICK_REQUIRED...

Alternatively (possibly cleaner) would be to use the full id and make use of

case "FAM_FACTS_ADD":
case "FAM_FACTS_QUICK":
case "FAM_FACTS_UNIQUE":
case "QUICK_REQUIRED_FAMFACTS":
   ..code

PR #838, opened 2016-02-17
Added England & Wales1939 Register as Census

ddrury:
Have taken column headings from record rather than transcript and checked against the National Archives at http://www.nationalarchives.gov.uk/help-with-your-research/research-guides/1939-register/#4-what-information-can-i-find-in-the-1939-register

### DIFF
--- a/app/Census/CensusColumnBirthDayMonthSlashYear.php
+++ b/app/Census/CensusColumnBirthDayMonthSlashYear.php
@@ -15,35 +15,27 @@
  */
 namespace Fisharebest\Webtrees\Census;
 
-/**
- * Definitions for a census
- */
-class CensusOfWales extends Census implements CensusPlaceInterface {
-	/**
-	 * All available censuses for this census place.
-	 *
-	 * @return CensusInterface[]
-	 */
-	public function allCensusDates() {
-		return array(
-			new CensusOfWales1841(),
-			new CensusOfWales1851(),
-			new CensusOfWales1861(),
-			new CensusOfWales1871(),
-			new CensusOfWales1881(),
-			new CensusOfWales1891(),
-			new CensusOfWales1901(),
-			new CensusOfWales1911(),
-			new RegisterOfWales1939()
-		);
-	}
+use Fisharebest\Webtrees\Individual;
 
+/**
+ * The individual's date of birth.
+ */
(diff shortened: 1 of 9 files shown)


fisharebest:
Thanks for adding the test cases!

Isn't `CensusColumnFullNameReversed` the same as `CensusColumnSurnameGivenNames` - or have I missed something?

ddrury:
Nope, I missed that completely  - Am I close enough to warrant making 
the necessary changes?

Regards
David



fisharebest:
> Am I close enough to warrant making the necessary changes?

That's the only thing I spotted when reviewing the code, so if you can update this, I'll merge the PR.

ddrury:
Hold fire - There's  case changes needed on a couple of file names

fisharebest:
> Hold fire

While you are doing whatever this is, can I make one more (picky!) change.  Last one, I promise.

For consistency/clarity, can you rename `CensusColumnBirthDaySlashYear` to `CensusColumnBirthDayMonthSlashYear`, to match the existing `CensusColumnBirthDaySlashMonthYear`

ddrury:
Can you clarify something for me. I'd named a file RegisterofEngland1939.php but called the class RegisterOfEngland1939 (lower case o in Of). I thought that composer didn't like things like that but travis-ci didn't pick anything up. For consistency I corrected the file name but did I really need to?

fisharebest:
> Can you clarify something for me. I'd named a file RegisterofEngland1939.php but called the class RegisterOfEngland1939 (lower case o in Of)

These need to agree exactly.  It may work on case insensitive filesystems (e.g. FAT, NTFS), but won't work on case  sensitive ones (e.g. ext2/3/4)

> I thought that composer didn't like things like

It depends on which autoloader you are using.

For a classmap autoloader (where it scan every file, and creates an array of `class` => `file`) it doesn't matter.  For a dynamic autoloader (where each class is expected to be in a particular location), it does.

ddrury:
> For a dynamic autoloader (where each class is expected to be in a particular location), it does

Shouldn't Scrutinizer pick up this type of error?

fisharebest:
> Shouldn't Scrutinizer pick up this type of error?

Possibly.  But it does a static code analysis, and doesn't know whether you've installed the dependencies using `composer dump-autoload`  or `composer dump-autoload --optimize`.

PR #843, opened 2016-02-22
Missed a bit from last commit

ddrury:
Also standardised header text to come into line with other censuses


ddrury:
travis-ci/pr check seems to be failing with 

> [Composer\Downloader\TransportException]  Could not authenticate against github.com

fisharebest:
> travis-ci/pr check seems to be failing

I think it is hitting download limits.

PR #844, opened 2016-02-22
Dynamically build the census column headers for shared notes

ddrury:
In function formatCensusNote, build the $headers array dynamically from the census column definitions rather than using a static array which can easily become out of date


fisharebest:
`if ($column->abbreviation()) {`

Do we have any columns without an abbreviated title?

ddrury:
Both Czech censuses

Regards
David



fisharebest:
Thanks

PR #848, opened 2016-02-23
CensusAssistantModule (again)

ddrury:
In function formatCensusNote, try to select the correct census from which to derive the column headers
Sorry, I'm getting a bit anal!

### DIFF
--- a/app/Module/CensusAssistantModule.php
+++ b/app/Module/CensusAssistantModule.php
@@ -28,6 +28,7 @@ use Fisharebest\Webtrees\I18N;
 use Fisharebest\Webtrees\Individual;
 use Fisharebest\Webtrees\Menu;
 use Fisharebest\Webtrees\Note;
+use Fisharebest\Webtrees\Soundex;
 
 /**
  * Class CensusAssistantModule
@@ -279,18 +280,6 @@ class CensusAssistantModule extends AbstractModule {
 	public static function formatCensusNote(Note $note) {
 		global $WT_TREE;
 
-
-		$headers = array();
-		foreach (Census::allCensusPlaces() as $allCensusesOfPlace) {
-			foreach ($allCensusesOfPlace->allCensusDates() as $census) {
-				foreach ($census->columns() as $column) {
-					if ($column->abbreviation()) {
-						$headers[$column->abbreviation()] = $column->title();
-					}
-				}
-			}
-		}
-
 		if (preg_match('/(.*)((?:\n.*)*)\n\.start_formatted_area\.\n(.*)((?:\n.*)*)\n.end_formatted_area\.((?:\n.*)*)/', $note->getNote(), $match)) {
 			// This looks like a census-assistant shared note
 			$title     = Filter::escapeHtml($match[1]);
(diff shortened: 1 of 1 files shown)


ddrury:
Was attempting to select specific census by the date & place the note referred to but couldn't handle the situation when the text was internationalised

app/Module/CensusAssistantModule.php
```diff
@@ -299,12 +288,39 @@ public static function formatCensusNote(Note $note) {
 			$data      = Filter::escapeHtml($match[4]);
 			$postamble = Filter::escapeHtml($match[5]);
 
+			// Get the column headers for the census to which this note refers
+			// assumes a soundex match between the fact place name and the specific
+			// census censusDate() function
+			$xref      = $note->getXref();
+			$firstIndi = $note->linkedIndividuals('NOTE')[0];
```
fisharebest: Dereferencing arrays like this won't work in PHP 5.3.  You need a temp variable.

What if there are no linked individuals?

app/Module/CensusAssistantModule.php
```diff
@@ -299,12 +288,39 @@ public static function formatCensusNote(Note $note) {
...
+			foreach ($firstIndi->getFacts('CENS') as $fact) {
+				if (trim($fact->getAttribute('NOTE'), '@') === $xref) {
```
fisharebest: Simpler to write `if ($fact->getTarget() === $note)`

app/Module/CensusAssistantModule.php
```diff
@@ -299,12 +288,39 @@ public static function formatCensusNote(Note $note) {
...
+				}
+			}
+
 			$fmt_headers = array();
-			foreach ($headers as $key => $value) {
-				$fmt_headers[$key] = '<span title="' . Filter::escapeHtml($value) . '">' . $key . '</span>';
+			foreach (Census::allCensusPlaces() as $allCensusesOfPlace) {
+				if (Soundex::compare($country, Soundex::daitchMokotoff($allCensusesOfPlace->censusPlace()))) {
+					foreach ($allCensusesOfPlace->allCensusDates() as $census) {
+						if (strcasecmp($census->censusDate(), $date) === 0) {
```
fisharebest: Is there a reason we can't simply write `if ($census->censusDate() == $fact->getAttribute('DATE')`

ddrury:
> What if there are no linked individuals?
I originally planned to scan the note to pick up the date & place but thought that was a bit iffy. The worst that happens is that the column headers won't get a title attribute

I'll address the code suggestions you made

fisharebest:
> worst that happens is that the column headers won't get a title attribute

I was more concerned that you're taking the first element of an empty array, which will give a PHP error...

ddrury:
> I was more concerned that you're taking the first element of an empty array, which will give a PHP error

I was planning to test and bail out if necessary

ddrury:
> if ($fact->getTarget() === $note)

always returns null, $fact->getTarget() calls $this->getValue to extract the xref, however on a simple census fact the gedcom is like the following (no xref)
```
1 CENS
2 DATE 29 SEP 1939
2 PLAC England
2 NOTE @N587@
```

PR #856, opened 2016-02-27
Consolidated changes to census assistant notes

ddrury:
Added styling for Census Assistant notes (same as markdown).
census-edit.php fixed bug in function updateCensusText whereby hidden <th> elements were included in the column count when generating the note text.
fixed bug in CensusAssistantModule.php function formatCensusNote to remove leading \n char from $data (real solution would be to change regex on line 294 but I go cross eyed looking at that).

Accomodate censuses with many columns (eg US 1930). To do this it was necessary to move the "Events of close relatives etc." line outside the facts table as colspan and table-layout: fixed don't play well together

NotesTabModule.php also need some work to properly show censuses with many columns

Select the correct census from which to derive the column headers

Use Soundex to perform place matching

### DIFF
--- a/app/Module/CensusAssistantModule.php
+++ b/app/Module/CensusAssistantModule.php
@@ -28,6 +28,7 @@ use Fisharebest\Webtrees\I18N;
 use Fisharebest\Webtrees\Individual;
 use Fisharebest\Webtrees\Menu;
 use Fisharebest\Webtrees\Note;
+use Fisharebest\Webtrees\Soundex;
 
 /**
  * Class CensusAssistantModule
@@ -279,18 +280,6 @@ class CensusAssistantModule extends AbstractModule {
 	public static function formatCensusNote(Note $note) {
 		global $WT_TREE;
 
-
-		$headers = array();
-		foreach (Census::allCensusPlaces() as $allCensusesOfPlace) {
-			foreach ($allCensusesOfPlace->allCensusDates() as $census) {
-				foreach ($census->columns() as $column) {
-					if ($column->abbreviation()) {
-						$headers[$column->abbreviation()] = $column->title();
-					}
-				}
-			}
-		}
-
 		if (preg_match('/(.*)((?:\n.*)*)\n\.start_formatted_area\.\n(.*)((?:\n.*)*)\n.end_formatted_area\.((?:\n.*)*)/', $note->getNote(), $match)) {
 			// This looks like a census-assistant shared note
 			$title     = Filter::escapeHtml($match[1]);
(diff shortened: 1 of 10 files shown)


fisharebest:
Hi Dave.

A couple of things.  Firstly, what is the intention of this:

```
.census-assistant-note {
    overflow-x: auto;
}
```

According to https://developer.mozilla.org/en/docs/Web/CSS/overflow-x this is browser dependent.  (And surely "auto" is the default?).

Secondly, the selector `notes_content > div` is very inefficient.  (Every `div` element on the page will need to be checked against this rule, which will then need to search for its parent.)

I can see that this is added to allow you to use `layout:fixed`.  Would the use of `<col>` help here?  See http://stackoverflow.com/questions/17831358/colspan-messes-with-fixed-width-table

ddrury:
overflow-x: is necessary to ensure a horizontal scroll bar for wide censuses (eg US 1930) otherwise the columns just shoot off the right hand size of the page. I think the default is 'visible'. You could use 'scroll' but that puts the scrollbar on all censuses wanted or not.

as far as notes_content > div goes it only check direct children of notes_content and there are only 4 children (two of which are scripts) so I don't think it's that inefficient but I can tighten it up a bit. I tried the col stuff but it seemed easier to go the way I have.

fisharebest:
> as far as notes_content > div goes it only check direct children of notes_content and there are only 4 children (two of which are scripts) so I don't think it's that inefficient

Every element gets checked against every rule.

Matching starts at the right-hand end of the selector - not the left.

So, for every `div` on the page, the browser will need to havigate up the document tree to its parent element and test the parent against `#notes_content`.

I'm out for most of the day, but will look at this when I get back.

fisharebest:
See https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Writing_efficient_CSS#Tag-categorized_rules_should_never_contain_a_child_selector.21

fisharebest:
Thanks Dave - I have merged this.

FYI, I made a couple of changes to the CSS.

Firstly, the overflow-x is actually need on **all** markdown, not just the census stuff.  There was an issue recently where preformatted text (i.e. markdown with 4 leading spaces) broke the layout in a similar way.

Secondly, I also added some `<colgroup>` and `<col>` elements to the facts tables, which allows us to combine a fixed table layout with a `colspan` in the first row - without any additional CSS rules.

ddrury:
Secondly, I also added some <colgroup> and <col> elements to the facts 
tables, which allows us to combine a fixed table layout with a colspan 
in the first row - without any additional CSS rules.

I look forward to seeing what you did - I couldn't make it work!

Regards
David

PR #859, opened 2016-02-28
Include table prefix in foreign key names to avoid conflicts

mnoorenberghe:
If you do a second installation of webtrees in a database with an existing WT installation but with a different table prefix, the installation will fail with some MySQL servers (e.g. MariaDB 10.0.24) because some of the foreign key names don't include the table prefix.

The error message was something like: "Error 1005: Can't create table (errno: 121)" but I don't have it handy since I ran into this last weekend. I confirmed this patch solves the problem for a new installation in a DB with an existing install with a different table prefix.

`app/Schema/Migration2.php` was the main issue but I searched for other occurrences with:
`git grep "FOREIGN KEY" | grep -v -E "FOREIGN KEY[^(]+##[^(]+\("`

### DIFF
--- a/app/Module/FamilyTreeFavorites/Schema/Migration2.php
+++ b/app/Module/FamilyTreeFavorites/Schema/Migration2.php
@@ -37,8 +37,8 @@ class Migration2 implements MigrationInterface {
 		try {
 			Database::exec(
 				"ALTER TABLE `##favorite`" .
-				" ADD FOREIGN KEY favorite_fk1 (user_id  ) REFERENCES `##user`   (user_id) ON DELETE CASCADE," .
-				" ADD FOREIGN KEY favorite_fk2 (gedcom_id) REFERENCES `##gedcom` (gedcom_id) ON DELETE CASCADE"
+				" ADD FOREIGN KEY `##favorite_fk1` (user_id  ) REFERENCES `##user`   (user_id) ON DELETE CASCADE," .
+				" ADD FOREIGN KEY `##favorite_fk2` (gedcom_id) REFERENCES `##gedcom` (gedcom_id) ON DELETE CASCADE"
 			);
 		} catch (PDOException $ex) {
 			// Already updated?
--- a/app/Module/FamilyTreeNews/Schema/Migration2.php
+++ b/app/Module/FamilyTreeNews/Schema/Migration2.php
@@ -36,8 +36,8 @@ class Migration2 implements MigrationInterface {
 		try {
 			Database::exec(
 				"ALTER TABLE `##news`" .
-				" ADD FOREIGN KEY news_fk1 (user_id  ) REFERENCES `##user`   (user_id)   ON DELETE CASCADE," .
-				" ADD FOREIGN KEY news_fk2 (gedcom_id) REFERENCES `##gedcom` (gedcom_id) ON DELETE CASCADE"
+				" ADD FOREIGN KEY `##news_fk1` (user_id  ) REFERENCES `##user`   (user_id)   ON DELETE CASCADE," .
+				" ADD FOREIGN KEY `##news_fk2` (gedcom_id) REFERENCES `##gedcom` (gedcom_id) ON DELETE CASCADE"
 			);
 		} catch (PDOException $ex) {
 			// Already updated?
--- a/app/Schema/Migration2.php
+++ b/app/Schema/Migration2.php
@@ -37,7 +37,7 @@ class Migration2 implements MigrationInterface {
 			" imported        BOOLEAN                NOT NULL DEFAULT FALSE," .
 			" PRIMARY KEY     (gedcom_chunk_id)," .
 			"         KEY ix1 (gedcom_id, imported)," .
-			" FOREIGN KEY fk1 (gedcom_id) REFERENCES `##gedcom` (gedcom_id)" .
+			" FOREIGN KEY `##gedcom_chunk_fk1` (gedcom_id) REFERENCES `##gedcom` (gedcom_id)" .
 			") COLLATE utf8_unicode_ci ENGINE=InnoDB"
 		);
 


mnoorenberghe:
The failing checks seem unrelated to this commit and this change makes these migrations consistent with the others in the repo.

fisharebest:
Hi Matt,

Firstly, we need back-ticks around the object name `` `##foo` `` instead of ` ##foo `, as the prefix is user-defined and may contain special characters.

Secondly, I'm not sure it will have the desired effect (or I don't understand the exact problem).

See http://dev.mysql.com/doc/refman/5.7/en/create-table-foreign-keys.html

The "index name" specified in migration scripts is actually ignored (Ive been getting it wrong for years!!).

It is only used if the column is not indexed, and an index needs to be created.  Looking at the first example (favorite), both `user_id` and `gedcom_id` are already indexed in the previous schema script.  Therefore no indexes need to be created, and the index name is ignored.

MariaDB says the same thing (https://mariadb.com/kb/en/mariadb/foreign-keys/) although I don't have a mariadb database on hand for testing.

After running the installer using the PR, I get the following table structure:

```
> SHOW CREATE TABLE wt_favorite;
CREATE TABLE `wt_favorite` (
  `favorite_id` int(11) NOT NULL AUTO_INCREMENT,
  `user_id` int(11) DEFAULT NULL,
  `gedcom_id` int(11) NOT NULL,
  `xref` varchar(20) COLLATE utf8_unicode_ci DEFAULT NULL,
  `favorite_type` enum('INDI','FAM','SOUR','REPO','OBJE','NOTE','URL') COLLATE utf8_unicode_ci NOT NULL,
  `url` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
  `title` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
  `note` varchar(1000) COLLATE utf8_unicode_ci DEFAULT NULL,
  PRIMARY KEY (`favorite_id`),
  KEY `news_ix1` (`gedcom_id`,`user_id`),
  KEY `wt_favorite_fk1` (`user_id`),
  CONSTRAINT `wt_favorite_ibfk_1` FOREIGN KEY (`user_id`) REFERENCES `wt_user` (`user_id`) ON DELETE CASCADE,
  CONSTRAINT `wt_favorite_ibfk_2` FOREIGN KEY (`gedcom_id`) REFERENCES `wt_gedcom` (`gedcom_id`) ON DELETE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci
```

No reference to `wt_favorite_fk1` ...

What does `show create table ....` give on your database?

fisharebest:
> No reference to wt_favorite_fk1 ...

Oops - I see it now (it's late!).  But why would naming this index, and not news_ix1 make a difference?

mnoorenberghe:
> Firstly, we need back-ticks around the object name `##foo` instead of ##foo, as the prefix is user-defined and may contain special characters.

Fixed

> Secondly, I'm not sure it will have the desired effect (or I don't understand the exact problem).

It does fix the problem on my installations though I admit I was also a bit confused from the error message and reading the documentation but various websites (including stack overflow) pointed to a duplicate foreign key name as the problem. This made sense for webtrees since my installation was failing at Migration2.php only when installing the 2nd site in the same DB (different prefix). The other non-module migration files already added the prefix to the FK name which explains why they worked fine.

> The "index name" specified in migration scripts is actually ignored (Ive been getting it wrong for years!!).

Well, as you can see from your `CREATE TABLE` output, the supposed `index name` ends up being interpreted as the `CONSTRAINT` `symbol`, notice that it moves before `FOREIGN KEY`, but don't ask me why :-P

> What does show create table .... give on your database?

```
CREATE TABLE `wt_favorite` (
`favorite_id` int(11) NOT NULL AUTO_INCREMENT,
`user_id` int(11) DEFAULT NULL,
`gedcom_id` int(11) NOT NULL,
`xref` varchar(20) COLLATE utf8_unicode_ci DEFAULT NULL,
`favorite_type` enum('INDI','FAM','SOUR','REPO','OBJE','NOTE','URL') COLLATE utf8_unicode_ci NOT NULL,
`url` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
`title` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
`note` varchar(1000) COLLATE utf8_unicode_ci DEFAULT NULL,
PRIMARY KEY (`favorite_id`),
KEY `news_ix1` (`gedcom_id`,`user_id`),
KEY `wt_favorite_fk1` (`user_id`),
CONSTRAINT `wt_favorite_fk1` FOREIGN KEY (`user_id`) REFERENCES `wt_user` (`user_id`) ON DELETE CASCADE,
CONSTRAINT `wt_favorite_fk2` FOREIGN KEY (`gedcom_id`) REFERENCES `wt_gedcom` (`gedcom_id`) ON DELETE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci
```

> But why would naming this index, and not news_ix1 make a difference?
An index is scoped to the table whereas a constraint name is for the whole database. See  https://dba.stackexchange.com/a/107119

In case the above didn't convince you more, remember that a) this does fix the problem for me and b) this just aligns these migrations with the rest of the WT migration style. (b) alone could be enough reason to fix this.

Thanks

fisharebest:
> this does fix the problem for me

I don't disagree.  Indeed, I have raised the very same issue on the laravel framework.

mnoorenberghe:
@fisharebest what are the next steps for this?

fisharebest:
Your change relies on undocumented behaviour (a bug??) in mariadb.

1) I haven't had the time to set up a mariadb server to investigate this.

2) It would be better to use the published syntax for creating named foreign keys, as this will work on both MySQL and MariaDB.

3) Historically, webtrees had separate "install" and "update" scripts.  These have both been changed over the years.  The result is that you will have a slightly different schema if your install/upgrade sequence was

1.0.0 -> 1.4.0 -> 1.7.4
1.4.0 -> 1.7.4
1.7.4

What I need to do is set up a test harness that will automate the different permutations of install/upgrade, and use a tool such as `mysqldbcompare` on the results.  I did this once, but ran out of disk-space on my laptop and had to delete it :-(

Then, knowing what possible schemas might exist, add a new migration to make them consistent.

As for this specific problem, I think it is caused by the lack of an index on `user_id`.  So, would a better approach be to update the original `create table` to add it - and then to add a new migration to add it as well (if it does not exist).

This way, the foreign key constraint won't need to create the index (with the wrong identifier).

Issue #864, opened 2016-03-01
note.php - needs colgroup...


ddrury:
... and appropriate class.

Not submitting PR 'cos can't figure out how to submit without sending all the dev stuff

wkitty42:
On 03/01/2016 04:40 PM, David Drury wrote:
> ... and appropriate class.
>
> Not submitting PR 'cos can't figure out how to submit without sending all the
> dev stuff

don't use "git add *" or "git add ." at the git command line... specifically add 
only those files that you want in the commit...

eg: $ git add README hello.rb


-- 
  NOTE: No off-list assistance is given without prior approval.
        *Please keep mailing list traffic on the list* unless
        private contact is specifically requested and granted.

fisharebest:
Fixed via #867

PR #867, opened 2016-03-03
note.php requires colgroup (issue #864)

ddrury:
Also to handle partial dates, only check census year to find match in CensusAssistantModule function formatCensusNote


fisharebest:
There are two changes in one PR.

The first change (note.php) is fine.

The second change (afaict) introduces a bug, rather than fixes one?
Before the change, I see `title="..."` for the column headings.  After the change, I don't.

So, I have extracted just the note.php changes.  If you want to open a new issue for the other changes, please do.

PR #872, opened 2016-03-08
show "pending changes exists" to moderator

josefpavlik:
The "pending changes exists" in the secondary menu was not showed to moderator. However, as the moderator is able to confirm the changes, it's right that the menu is shown.  This patch resolve the problem.

Don't worry about other commits, it's the patch that has been already included in the mainstream.


fisharebest:
Thanks.

PR #874, opened 2016-03-13
Standardise ...Controller->getEditMenu logic

ddrury:
Copy functionality of other controllers to IndividualController->getEditMenu() 
Remove "Add to favourites" from the edit menu as this functionality also appears as a menu item in the secondary menu (which to my mind is it's logical home)


ddrury:
problem with RTL languages

PR #875, opened 2016-03-13
Standardise xxxController->getEditMenu logic

ddrury:
1. Copy functionality of other controllers to IndividualController->getEditMenu()
2. Remove "Add to favourites" from the edit menu as this functionality also appears as a menu item in the secondary menu (which to my mind is it's logical home)
3. Fixed various RTL icon problems on the edit menu

Theme owners may like to review my choice of icons for the "Edit names" and "Re-order families" menu options

### DIFF
--- a/app/Controller/FamilyController.php
+++ b/app/Controller/FamilyController.php
@@ -90,10 +90,8 @@ class FamilyController extends GedcomRecordController {
 					'onclick' => 'return reorder_children("' . $this->record->getXref() . '");',
 				)));
 			}
-		}
 
-		// delete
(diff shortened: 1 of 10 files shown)


fisharebest:
We recently removed all the functionality from top-level menus, as this breaks touch-screen devices.  The tap can either open the submenu or activate the top menu - but not both.  So, I removed this from all menus (rather than adding it to all).

Also, I'm not sure about the "edit name" link.  Where there are mutliple names, it won't necessarily edit the first one, and (like "add to favorites") it could be confusing to have two different menu items for the same thing.

PR #877, opened 2016-03-25
Improve logic for obtaining census column headings

ddrury:
When attempting to obtain census headings, only match on (place and) year (rather than full date) to accomodate partial dates (eg US 1930 census - Apr 1930)


fisharebest:
Create two censuses - USA 1830 and USA 1860.

There are no tooltips on the 1830 census - but there **are** tooltips on the 1860 census.

If you apply this patch, the situation is reversed.  Tooltips are shown on the 1830 census but **not** on the 1860 census.

ddrury:
Fixed problem mentioned above, however note that the US 1830 census headings will be incorrect as there are duplicate column abbreviations in CensusOfUnitedStates1830::columns()

ddrury:
Closed as no longer appropriate

PR #879, opened 2016-03-27
Improve reliability of Media::getServerFilename() function

ddrury:
I recently notice a few medialog entries on my production system and when I checked, saw huge numbers of similar errors on my windows dev system.

The error types I saw were 
- Invalid mime type
- failure to create an already existing thumb folder
- recreating an already existing thumb file
- streaming failure from the mediafilewall

Further investigation revealed they were all caused by the path to the relevant file occasionally being blank in media::getServerFilename(). This was resolved by using the do loop near the top of the getServerFilename function (I used realpath as I symlink to the media folder on my dev system and initially thought this may contribute to the problem - I don't think it does but it doesn't hurt)

Interestingly I put a count variable in the do loop and var_dumped it on exit and it always showed 1, but removing the loop led to the failures re-appearing so I wonder if the underlying problem is some sort of timing error.

Anyway I submit this for your consideration!

**Later**: I found that function realpath did occasionally return false (real false not empty string etc.) and it could take a number of iterations through the loop before the genuine path was returned. I've tried this with PHP versions 5.3 - 7 on windows 10. And PHP 7 on my production Linux system

### DIFF
--- a/app/Media.php
+++ b/app/Media.php
@@ -145,12 +145,23 @@ class Media extends GedcomRecord {
 		if ($this->isExternal() || !$this->file) {
 			// External image, or (in the case of corrupt GEDCOM data) no image at all
 			return $this->file;
-		} elseif ($which == 'main') {
+		}
+
+		$real_path = realpath(WT_DATA_DIR . $MEDIA_DIRECTORY);
+
+		if (!$real_path) {
+			Log::addMediaLog('Unable to retrieve the path to ' . $this->file . ' for ' . $this->getXref());
+			return $this->file; // can't do anything else
+		}
+
+		$real_path .= DIRECTORY_SEPARATOR;
+
+		if ($which == 'main') {
 			// Main image
-			return WT_DATA_DIR . $MEDIA_DIRECTORY . $this->file;
+			return $real_path . $this->file;
 		} else {
 			// Thumbnail
-			$file = WT_DATA_DIR . $MEDIA_DIRECTORY . 'thumbs/' . $this->file;
+			$file = $real_path . 'thumbs' . DIRECTORY_SEPARATOR . $this->file;
 			// Does the thumbnail exist?
 			if (file_exists($file)) {
 				return $file;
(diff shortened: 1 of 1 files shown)


ddrury:
After further investigation I've found that increasing the value of realpath_cache_size in php.ini makes the call reliable without the do loop, however I guess many people won't have sufficient access to change this config item.

PS I realise you're planning on using phpleague's flysytem eventually but I submit this as a short term sticking plaster

fisharebest:
I'm trying to reproduce the error.
I'm filling up the realpath cache, and looking at the results from `realpath()` and the cache's memory usage.

Can you run this script for me, with the PHP.INI settings indicated.

```
<?php
// php.ini
// realpath_cache_size = 16K
// realpath_cache_ttl  = 120

mkdir('tmp', 0777, true);
for ($i=0; $i<10000; ++$i) {
	touch('tmp/file' . $i);
	echo realpath('tmp/file' . $i), ' ', realpath_cache_size(), PHP_EOL;
}
```

What I see is the cache_size column increasing up to just under the limit (16384), and then stopping there.  i.e. the cache fills and old values are discarded.  (I've tried on PHP5.3 and PHP7.0)

If I understand yoru problem correctly, I'd expect that when your server reaches the cache limit, the next call to real_path() will be empty.

```
/Users/greg/tmp/file195 15850
/Users/greg/tmp/file196 15930
/Users/greg/tmp/file197 16010
/Users/greg/tmp/file198 16090
/Users/greg/tmp/file199 16170
/Users/greg/tmp/file200 16250
/Users/greg/tmp/file201 16330
/Users/greg/tmp/file202 16330
/Users/greg/tmp/file203 16330
/Users/greg/tmp/file204 16330
```

ddrury:
> What I see is the cache_size column increasing up to just under the limit (16384), and then stopping there. i.e. the cache fills and old values are discarded

So do I (I put clearstatcache(true); before the loop to make the results repeatable)

> D:\david\vhosts\webtrees\175\tmp\file0 621
> D:\david\vhosts\webtrees\175\tmp\file1 850
> ...
> D:\david\vhosts\webtrees\175\tmp\file66 15906
> D:\david\vhosts\webtrees\175\tmp\file67 16138
> D:\david\vhosts\webtrees\175\tmp\file68 16370
> D:\david\vhosts\webtrees\175\tmp\file69 16370
> D:\david\vhosts\webtrees\175\tmp\file70 16370
> D:\david\vhosts\webtrees\175\tmp\file71 16370
> etc


BTW the way I invoke the error is to look on the album tab for a each of a number of individuals with a significant number of images attached. I also tried putting clearstatcache in Media.php but it made no difference.

ddrury:
There is definitely a problem with realpath & symlinks on windows. (There is an open PHP bug report but it was raised in 2012 and it seems dormant) - so the solution is don't use symlinks!

I've left the realpath line in place so it resolves paths of the form `webtrees\data\..\..\media`
and also the additional logging I added when trying to find this problem.

ddrury:
Withdrawn - not appropriate for v1.8

PR #884, opened 2016-03-31
Charts As Modules

ric2016:
All existing charts as modules, as suggested in the forum.

Migration34.php should ensure that the 'new' modules are initially enabled, so that the change is transparent for users.


app/Schema/Migration34.php
```diff
@@ -0,0 +1,43 @@
...
+	public function upgrade() {
+		Database::exec(
+			"INSERT IGNORE INTO `##module` (module_name, status, tab_order, menu_order, sidebar_order) VALUES" .
+			" ('ancestors_chart', 'enabled', NULL, NULL, NULL),".
+			" ('compact_tree_chart', 'enabled', NULL, NULL, NULL),".
+			" ('descendancy_chart', 'enabled', NULL, NULL, NULL),".
+			" ('family_book_chart', 'enabled', NULL, NULL, NULL),".
+			" ('fan_chart', 'enabled', NULL, NULL, NULL),".
+			" ('hourglass_chart', 'enabled', NULL, NULL, NULL),".
+			" ('lifespans_chart', 'enabled', NULL, NULL, NULL),"
```
fisharebest: The migration script is missing `.` at the end of several lines.
It also fails to set the module privacy, so after installation, no modules are visible.

Instead, consider calling `Module::getInstalledModules()` which does all this automatically.

app/Module/CompactTreeChartModule.php
```diff
@@ -0,0 +1,70 @@
...
+	public function defaultAccessLevel() {
+		return Auth::PRIV_PRIVATE;
+	}
+	
+	/**
+	 * Return a menu item for this chart.
+	 *
+	 * @return Menu
+	 */
+	public function getChartMenu() {
```
fisharebest: This function is called with a parameter $individual.

Use this, and remove the globals.

(Same in all other charts)

app/Theme/AbstractTheme.php
```diff
@@ -945,24 +945,23 @@ protected function individualBoxMenu(Individual $individual) {
...
-			$this->menuChartRelationship($individual),
-			$this->menuChartTimeline($individual),
-		));
-
-		usort($menus, function (Menu $x, Menu $y) {
-			return I18N::strcasecmp($x->getLabel(), $y->getLabel());
-		});
+		$menus = array();
+		foreach (Module::getActiveCharts($this->tree) as $chart) {
+			$menu = $chart->getChartMenu($individual);
```
fisharebest: Some charts are specific to an individual.  e.g. pedigree.
Others are not.  e.g. statistics.

The non-individual charts do not belong here.

admin_modules.php
```diff
@@ -181,7 +181,7 @@ function (index, value) {
 			'<td class="text-center text-muted hidden-xs">', $module instanceof ModuleTabInterface ? '<i class="fa fa-folder" title="' . I18N::translate('Tab') . '"></i>' : '-', '</td>',
 			'<td class="text-center text-muted hidden-xs">', $module instanceof ModuleSidebarInterface ? '<i class="fa fa-th-large" title="' . I18N::translate('Sidebar') . '"></i>' : '-', '</td>',
 			'<td class="text-center text-muted hidden-xs">', $module instanceof ModuleBlockInterface ? (($module->isUserBlock() ? '<i class="fa fa-user" title="' . I18N::translate('My page') . '"></i>' : '') . ($module->isGedcomBlock() ? '<i class="fa fa-tree" title="' . I18N::translate('Home page') . '"></i>' : '')) : '-', '</td>',
-			'<td class="text-center text-muted hidden">', $module instanceof ModuleChartInterface ? '<i class="fa fa-check" title="' . I18N::translate('Chart') . '"></i>' : '-', '</td>',
+			'<td class="text-center text-muted hidden-xs">', $module instanceof ModuleChartInterface ? '<i class="fa fa-check" title="' . I18N::translate('Chart') . '"></i>' : '-', '</td>',
```
fisharebest: Can we find a better icon than `fa-check`?
How about `fa-share-alt`?

fisharebest:
In `AbstractTheme`, you removed all the `function menuChartXxxxx()`.

This will break custom themes - including the one I use on my own site ;-)

For backwards compatibility, keep these, but implement them using the new modules???

```
/**
 * @deprecated
 */
function menuChartXXXX(Individual $individual) {
  $chart = new ChartXXXX;
  return $chart->getChartMenu($individual);
}
```

app/Module/CompactTreeChartModule.php
ric2016: Ok, done.

app/Theme/AbstractTheme.php
ric2016: Ok, done (via separate interface method).

admin_modules.php
ric2016: Ok, done.

ric2016:
Thanks for the fast review and suggestions! I have also restored the menuChartXxxxx() in AbstractTheme, as suggested. Only the Migration is still todo.

app/Schema/Migration34.php
ric2016: this doesn't seem to work because Module::getInstalledModules('enabled') cannot be called during migration (I18N not initialized at that point in session.php, but used in module constructors for getTitle()).

fisharebest: I will need to check, but I do not think that it is necessary to sort the modules in `Module::getInstalledModules` 

So, we can remove the sort - and the call to `I18N::xxx`

ric2016: It's not just the sort that requires I18N, it's the '$this->getTitle();' in the AbstractModule constructor. I have just added the required updates directly to Migration34 after all.

fisharebest:
FYI, I have just committed a change to `Modules::getInstalledModules()` to remove the dependency on `I18N::translate()`.

Does this mean you can not use it in `Migration34`?

ric2016:
Hm, I don't see this change ... In the master branch? (but I'm also still struggling with Git, that could be the reason)

fisharebest:
Oops - I had committed the change to the wrong branch.

I have now added it to master.

Sorry!

ric2016:
Ok, it works now - I can use it in Migration34 to enable all new chart modules via `Module::getInstalledModules('enabled');` However, this has the effect of also enabling all other modules, which may not be intended (some of them may have been disabled explicitly). So, after all I'm not sure this is the best way to go?

fisharebest:
> However, this has the effect of also enabling all other modules, which may not be intended (some of them may have been disabled explicitly). 

If a module has been disabled, then `getInstalledModules()` will not change it.

It will only set the status for a module that has not been seen before.

If you think it is not working correctly, can you give a specific example?

ric2016:
You're right, I didn't test this correctly - Disabled modules stay disabled. Ok, I have simplified `Migration34` now.

fisharebest:
Thanks.  I will test this later today.

You have committed a OAUTH token into .travis.yml.  Did you mean to do this?

ric2016:
> You have committed a OAUTH token into .travis.yml. Did you mean to do this?

That was to get rid of the failing tests (Scrutinizer etc.) I didn't see a better way to enable these tests - But maybe I should just ignore them?

fisharebest:
I think this is a bug in TRAVIS.

I have added my own OAUTH token to the configuration.
If I make it public (i.e. displayed in the logs), it works.
If I make it private (i.e. not displayed in the logs), it used to work, but does not work now.

I will exclude this commit when I merge the changes.

fisharebest:
I have made various small changes:

* removed the OAUTH commit
* PHPDoc
* global variables
* use new chartbox menu in fanchart.php

These have been added to a new branch `charts`.  (You can create PRs against this branch)

There is still a little more work to do before we can merge this branch into master.
However, I have to stop now, and will return to this later.

* There are hard-coded references to various charts, especially the relationship chart.  If the chart/module is disabled, we should not show these links.
* Each chart should refuse to display (e.g. HTTP redirect back to index.php) if it is disabled for the current tree.

`if (Module::getModuleByName('my_module')) { /* enabled */ } else { /* disabled */ }`

ric2016:
Ok, thank you for your help and comments so far - I'll see if I can address the todos you listed.

ric2016:
Hm, the pull request didn't seem to go to the 'charts' branch ... Let me try to fix that.

> Each chart should refuse to display (e.g. HTTP redirect back to index.php) if it is disabled for the current tree.

I have addressed this exemplarily for the compact tree chart. I think it should not be displayed as well if a respective access level is configured, so I have added a helper function for this check in `Module`. If this is ok, I'll handle the other charts in the same way.

fisharebest:
FYI, I've already implemented that in 0e4f4a5

ric2016:
Ok, I think I have figured out how to handle further changes - They would go into a new Pull Request, and this one should be closed, right?
I'll stop making changes for now though, so we won't work in parallel.

fisharebest:
These changes have been merged into the master branch.

Thanks!

ric2016:
> There are hard-coded references to various charts, especially the relationship chart. If the chart/module is disabled, we should not show these links.

These links are currently still shown, but redirect to index.php. 

Maybe these links should be provided via a separate method in `ModuleChartInterface`, something like `getLinkForRelationship(Individual $individual1, Individual $individual1)` (returning null if not applicable for the respective chart)? This would make the references a bit less hard-coded, and there could even be custom chart modules providing these links. Then, if no suitable chart is active/accessible, the link itself could be hidden.

(if there are multiple charts returning non-null links, we have to decide which one to use, though)

PR #905, opened 2016-04-15
Add options to give users access to their direct ancestors and/or direct descendants.

kroseneg:
This change adds an options to give users access to their direct ancestors and/or direct descendants, addionally to the path calculations of relatives.
    
To find the ancestors/descendants the link database table is used.
Relations between individuals are calculated using the FAMC, FAMS, HUSB, WIFE and CHIL facts, and stored in an array. This array is used to find the ancestors/descendants.


fisharebest:
The "relationship privacy" function only applies to **living** individuals.  Dead individuals are always visible.

So, if you set the path lengths to 2, it would mean that a user could:

* see their living children and grandchildren - but not see their living great-grandchildren.
* see their living parents - but not see their living great-grandparents

I do not understand the need for this.

What is the problem that this is solving?

kroseneg:
Dead individuals are only visible if the privacy settings of the tree permit it (setting SHOW_DEAD_PEOPLE).

I want to restrict the visibility of details, so not all members see the details of all individuals, and I want members to be able to edit their part of the family, but not others.
With the path length calculation it is only possible to to set an undirected count of hops. Setting it to 2 would permit the parents of myself and my wife to edit details of our daughter, setting it to 4 would permit my wifes parents to access the siblings of my parents. I want neither of both. But I want them to be able to add/edit details of their own ancestors.

I already noticed that my first approach with only adding the ancestors/descendants feature will not fit my needs, so i'm currently extending it with multiple access rules, each one starting from its own individual and having the ability to define a undirected path length and/or ancestors/descendants, resulting in one access level per rule. Users will have an access level on the whole tree (e.g. 'Visitor') which can be raised to 'Member' or 'Editor' if one of the rules matches.
This is working quite nicely already, but I still have to implement a few things because until now webtrees only knows one access level per tree, not per individual.

Furthermore I need to distinguish between a real visitor (the whole internet) and a known user, because I don't want to show anything to the whole internet, but show the names (without details) of all individuals to known users, but that's independent from the access rules mentioned above.

fisharebest:
> would permit my wifes parents to access the siblings of my parents. I want neither of both.

If you do not want your family to see your wife's family, then create two trees.  One for you and one for your wife.

> Furthermore I need to distinguish between a real visitor (the whole internet) and a known user,

Use the option "Show family tree" -> show to members.

fisharebest:
> If you do not want your family to see your wife's family, then create two trees. One for you and one for your wife.

Or you could disconnect the two halves of your tree by deleting the family record with your wife.

You could then:
* add a duplicate copy of your record to your wife
* and add a duplicate copy of your wife to you.

There would now be two copies of both you and your wife.
Your wife's family would see you - but not your relations.
Your family would see your wife - but not her relations.

PR #916, opened 2016-04-28
HTML validation error

ddrury:
For show_common_surnames section (~ line 200), corrected unbalanced `<div></div>` and simplified code by removing two if {...} statements from foreach loop


fisharebest:
Looking at `FunctionsDb::getCommonSurnames()`, I think it already filters out `@N.N.` - so perhaps we don't need to do it again?

ddrury:
Removed un-necessary array_filter(..) as per last comment

Issue #913, opened 2016-04-23
Feature request: option to prevent relationship through sibling in relationship chart


gvd:
I often want to search what the relationship is between two individuals through just their ancestors. Currently I often get back a relationship chart that connects the individuals through the spouse of a sibling.

ric2016:
I have implemented a modified relationship chart for my tree a while ago, which provides this functionality.
[See here for an example](http://cissee.de/webtrees/relationship2.php?ged=mueller&pid1=I1089&show_full=1&show_full-GUI-ONLY=1&pid2=I1090&find=4). This is not the most recent version, but you get the idea.

Option 4 uses the original algorithm. Option 3 is based on a modification of the Dijkstra algorithm (using weights to model a preference for blood relations), and the first two options use common ancestors only. I plan to provide this as a custom module (once the 'Charts as Modules' functionality is released), but at least the modified Dijkstra could also be integrated directly.

gvd:
That looks great! Would you possibly share the code for that?

ric2016:
Sure, but it's probably easier to do this after the next webtrees release (it will be possible, then, to add additional charts, like this one, to the list of charts via the control panel - in the current version, you'd have to adjust the code).

fisharebest:
> Option 3 is based on a modification of the Dijkstra algorithm (using weights to model a preference for blood relations)

I'm curious to know what you did here.  I did try to use weightings, but didn't get satisfactory results.

ric2016:
The algorithm is based on weights plus a cost function (i.e. the weights are not just added up):

The general idea is to view a path from one individual (or family) to another as a series of either ascending (FAMC, HUSB, WIFE) or descending (FAMS, CHIL) steps, and to find a path with minimal turns, where a turn is defined as a descending step, followed by an ascending step (these turns are undesirable because the individuals (families) at the ends of such a turn are not related by blood).

Example: The relation
`father's mother's daughter's husband`
is
`FAMC-HUSB-FAMC-WIFE-FAMS-CHIL-FAMS-HUSB`
i.e.
`a(scending)-a-a-a-d(escending)-d-d-a`
and therefore a path with one turn (right at the end).

The algorithm uses a cost function based on the weights of the last two edges in the path to the current node: Descending edges are weighted d, ascending edges are weighted a. If the last two edges have weights d and a (i.e. a turn), the cost for the step is some large number, e.g. 10000, otherwise the cost is 1.

ric2016:
[...]

PR #925, opened 2016-05-08
Charts as Modules: Remove hard-coded references

ric2016:
As suggested in a comment in Pull Request #884 (which was already completed at that time), this removes hard-coded links to the relationships chart (currently, these links are displayed but redirect to the main page if the relationship chart is inaccessible).

These links are now provided via a separate method in `ModuleChartInterface`, returning null if not applicable for the respective chart. Thus, custom chart modules may also provide these links. If no suitable chart is active/accessible, the link itself is no longer displayed.


fisharebest:
I do not think that `Module::getLinkForRelationship` belongs in `class Module`.

Elsewhere, we use code like this:

```
if (Module::getModuleByName('XXX')) {
    ...
    // Code that uses module XXX
    ...
}
```

The relationship chart module can take 1 or 2 individuals.
Other charts have a variable number of individuals (lifespans, timelines).

Perhaps add a function `ModuleChartInterface::getChartLink(array $individual) { }` which generates the link.

Then we could have something like (in class `Functions`??)

```
function xxx(Individual $individual1, Individual $individual2) {
    if (Module::getModuleByName('RelationshipsChartModule')) {
        $link = RelationshipsChartModule::getChartLink(array($individual1, $individual2));
        return $relationship_link...
    } else {
        return $relationship_name...
    }
}
```

I think this is a cleaner structure,
What do you think?

ric2016:
The idea was to remove any 'hard-coded' references to a specific chart (not only hard-coded uris).
I.e. something like `if (Module::getModuleByName('RelationshipsChartModule'))` is still problematic.
(One reason for this is that users may want to use a different relationship chart, which would e.g. be a solution for issue #913 )

> I do not think that Module::getLinkForRelationship belongs in class Module.

ok - we could just move this method to Functions?

ric2016:
Do you have any further thoughts on this?

Basically, the question is whether to only use links to a specific chart (RelationshipsChartModule in this case), or to allow links to a chart of a specific type (any chart displaying relationships), which are then resolved to a suitable specific chart. Or do you think this isn't necessary at all?

fisharebest:
I think that the best way to implement this sort of problem is to use PHP interfaces to allow a module to state its capabilities.

For example, this sort of logic,

``` php
// relationships.php
$relationship_algorithms = [];
foreach ($modules as $module) {
  if ($module instanceof InterfaceCalculatesRelationships) {
    $relationship_algorithms[] = $module;
  }
}

if (count($relationship_algorithms) > 1) {
  // Show an option to allow the user to choose algorithm.
}
```

Or perhaps

``` php
foreach ($modules as $module) {
  if ($module->hasCapability('canCalculateRelationships') {
    $relationship_algorithms[] = $module;
  }
}
```

Or perhaps a "hook" system like drupal or wordpress.


Whatever, this approach means that we move away from modules that are just a simple chart/menu/report/etc.   Instead, a module is just a set of capabilities.

IMHO, the best fix for #913 would be to add this functionality to the core code.

ric2016:
> IMHO, the best fix for #913 would be to add this functionality to the core code.

Ok, I'll have to adjust the code a bit though - It turns out the approach as described there is not correct in all cases.

fisharebest:
I think the latest 2.0 code solves this.

ric2016:
> I think the latest 2.0 code solves this.

I don't see how - The links are still generated regardless of whether a relationship chart is actually enabled (in FunctionsPrintFacts line 542).
Note also that there seems to be a missing route definition here (the generated link referes to the undefined route 'relationships').

ric2016:
ok, that will work. 

If no chart isn't accessible to the current user (Access level set via module adminstration), a link is still shown though. In general the chart is still actually accessible, it just doesn't appear in the chart menu. Is it supposed to work this way? Probably the same for reports etc., when a url is entered directly.

fisharebest:
You are right.  I was looking for the module in the list of *all modules*, not the list of *available charts*.
This was skipping the check for access level.

Fixed.

> Is it supposed to work this way? Probably the same for reports etc., when a url is entered directly.

This is similar to typing URLs of private individuals.
A similar check is needed.

Fixed.

PR #926, opened 2016-05-08
Update GoogleMapsModule.php

mpskeeter:
build gmarks as a  0 based array.


fisharebest:
I think another change is needed at line 2022?

Maybe this syntax is easier to read?

```php
$gmark = array(
  ...
);
$gmarks[] = $gmark;
```

PR #927, opened 2016-05-08
Suppress a few elements when printing

ddrury:


fisharebest:
Ref the CSS changes.

In some cases, there is a 4th icon - such as shared notes.

Instead of adding the 3 elements copy/edit/delete, would it work to add the parent element? `<div class="editfacts">`?

ddrury:
I can't find the 4th icon you're talking about, but anyway, unfortunately not every occurrence of those icons are wrapped in `<div class="editfacts">`
For example the edit & delete icons a few levels down the hierarchy in `<div id="header_accordion1">`

fisharebest:
screen shot 2016-05-08 at 21 40 47

ddrury:
OK, changed as suggested

Issue #705, opened 2015-08-26
Latest Beta + 1.7.2 - Google Maps  Off by One


eric215:
Fails on 1.7.2 and on latest demo build

Select Google Maps tab.  Select any event.  The location highlighted is for the event one above the item selected

In the image below the "Birth" link was selected
 
![googleoffbyone](https://cloud.githubusercontent.com/assets/13973937/9485233/8f8ed74e-4b84-11e5-890f-bb151dec8741.png)

fisharebest:
The code is using the index-number of the fact, instead of the index-number of the location, in the call to `myclick()`.

PR #931, opened 2016-05-12
GoogleMapsModule.php

mpskeeter:
Refactored to create new $gmark and push to $gmarks[].

This will properly display a single fact location when used multiple times in the list.


fisharebest:
Fixes #705

PR #942, opened 2016-05-22
Bug fix and code tidy

ddrury:
Sorry - should have said in title - relates to googlemap module

- Cleared javascript warnings and errors by updating calls to Googlemaps api - now using the latest version, this has necessitated the reintroduction of the API key (it works without one but generates a warning) see https://developers.google.com/maps/documentation/javascript/get-api-key

- Corrected validation errors in placelist.php - table elements could be improperly closed depending on the number of places to be displayed.

For code related to the Individual tab 

- fixed bug whereby clicking on the events in the sidebar produced arbitrary results (the index into the marker array was being incorrectly calculated)

- shifted about 90% of the heavy lifting from javascript to the server (to my mind the response seems snappier but it may be wishful thinking!)

Started on a general code tidy but there is plenty more to do

### DIFF
--- a/app/Module/GoogleMapsModule.php
+++ b/app/Module/GoogleMapsModule.php
@@ -252,6 +252,7 @@ class GoogleMapsModule extends AbstractModule implements ModuleConfigInterface, 
 			->setPageTitle(I18N::translate('Google Maps™'));
 
 		if (Filter::post('action') === 'update') {
+			$this->setSetting('GM_API_KEY', Filter::post('GM_API_KEY'));
 			$this->setSetting('GM_MAP_TYPE', Filter::post('GM_MAP_TYPE'));
 			$this->setSetting('GM_USE_STREETVIEW', Filter::post('GM_USE_STREETVIEW'));
 			$this->setSetting('GM_MIN_ZOOM', Filter::post('GM_MIN_ZOOM'));
@@ -321,6 +322,17 @@ class GoogleMapsModule extends AbstractModule implements ModuleConfigInterface, 
 			<input type="hidden" name="action" value="update">
 
 			<!-- GM_MAP_TYPE -->
+			<div class="form-group">
+				<label class="control-label col-sm-3" for="GM_API_KEY">
+					<?php echo I18N::translate('API Key') ?>
+				</label>
+				<div class="col-sm-9">
+					<input id="GM_API_KEY" class="form-control" type="text" name="GM_API_KEY" value="<?php echo $this->getSetting('GM_API_KEY') ?>">
+					<p class="small text-muted"><?php echo I18N::translate('All JavaScript API applications require authentication using an API key (or a client ID for Google Maps API for Work customers). Including a key when loading the API allows you to monitor your application\'s API usage in the Google Developers Console, enables per-key instead of per-IP-address quota limits, and ensures that Google can contact you about your application if necessary.'); ?>
+						<a href="https://developers.google.com/maps/documentation/javascript/get-api-key"><?php echo I18N::translate('Get API Key'); ?></a>
+					</p>
+				</div>
+			</div>
 			<div class="form-group">
 				<label class="control-label col-sm-3" for="GM_MAP_TYPE">
 					<?php echo I18N::translate('Default map type') ?>
@@ -522,7 +534,8 @@ class GoogleMapsModule extends AbstractModule implements ModuleConfigInterface, 
 	 * @return string
 	 */
 	private function googleMapsScript() {
-		return 'https://maps.google.com/maps/api/js?v=3.2&amp;sensor=false&amp;language=' . WT_LOCALE;
+		$key = $this->getSetting('GM_API_KEY');
+		return 'https://maps.google.com/maps/api/js?v=3&amp;key=' . $key . '&amp;language=' . WT_LOCALE;
 	}
 
 	/**
@@ -2533,7 +2546,7 @@ class GoogleMapsModule extends AbstractModule implements ModuleConfigInterface, 
 		echo '<tr style="vertical-align:top;"><td>';
 		if ($STREETVIEW && $level != 0) {
 			// Leave space for the Street View buttons, so that the maps align vertically
-			echo '<div id="place_map" style="margin-top:31px; border:1px solid gray; width: ', $this->getSetting('GM_PH_XSIZE'), 'px; height: ', $this->getSetting('GM_PH_YSIZE'), 'px; ';
+			echo '<div id="place_map" style="margin-top:25px; border:1px solid gray; width: ', $this->getSetting('GM_PH_XSIZE'), 'px; height: ', $this->getSetting('GM_PH_YSIZE'), 'px; ';
 		} else {
 			echo '<div id="place_map" style="border:1px solid gray; width:', $this->getSetting('GM_PH_XSIZE'), 'px; height:', $this->getSetting('GM_PH_YSIZE'), 'px; ';
 		}
@@ -2555,12 +2568,12 @@ class GoogleMapsModule extends AbstractModule implements ModuleConfigInterface, 
 				$adminplaces_url .= '&amp;parent=' . $latlng['pl_id'];
 			}
 			$update_places_url = 'admin_trees_places.php?ged=' . $WT_TREE->getNameHtml() . '&amp;search=' . urlencode(implode(', ', array_reverse($parent)));
-			echo '<p>';
+			echo '<div id="place_map_links" style="width:' . $this->getSetting('GM_PH_XSIZE') . 'px;">';
 			echo '<a href="module.php?mod=googlemap&amp;mod_action=admin_config">', I18N::translate('Google Maps™ preferences'), '</a>';
 			echo ' | <a href="' . $adminplaces_url . '">' . I18N::translate('Geographic data') . '</a>';
 			echo ' | <a href="' . $placecheck_url . '">' . I18N::translate('Place check') . '</a>';
 			echo ' | <a href="' . $update_places_url . '">' . I18N::translate('Update place names') . '</a>';
-			echo '</p>';
+			echo '</div>';
 		}
 		echo '</td>';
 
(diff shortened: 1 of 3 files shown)


ddrury:
Noticed a number of Scrutinizer issues - I'll fix where possible and resubmit

fisharebest:
> Noticed a number of Scrutinizer issues - I'll fix where possible and resubmit

There's a lot of changes here.

I'll probably cherry-pick the "tidy up" changes first.  This will make it easier to review the functional changes.

fisharebest:
> I'll probably cherry-pick the "tidy up" changes first.

Done.  You may want to merge them into your branch.

ddrury:
Merged, fixed as many Scrutiniser issues as is practical and re-pushed

ddrury:
I'm not aware of any outstanding bugs (famous last words) so any further code tidy up will be submitted as follow on PRs

placelist.php
```diff
@@ -218,28 +199,27 @@
...
-		if ($myindilist) {
+		if (!empty($myindilist)) {
 			echo '<div id="places-indi">', FunctionsPrintLists::individualTable($myindilist), '</div>';
 		}
-		if ($myfamlist) {
+		if (!empty($myfamlist)) {
 			echo '<div id="places-fam">', FunctionsPrintLists::familyTable($myfamlist), '</div>';
 		}
-		if (!$myindilist && !$myfamlist) {
+		if (empty($myindilist) && empty($myfamlist)) {
```
fisharebest: There isn't a corresponding tab for this block.

However, I'm not sure this code branch is even needed.  Surely, there must be at least one individual or one family - otherwise where did the place come from?

placelist.php
```diff
@@ -94,7 +91,7 @@
 	$numfound = count($child_places);
 
 	//-- if the number of places found is 0 then automatically redirect to search page
-	if ($numfound == 0) {
+	if (!$numfound) {
```
fisharebest: $numfound is a number.  Change to `===` if anything?

placelist.php
ddrury: Agreed - I'll remove it

placelist.php
ddrury: Changed

fisharebest:
> I'm not aware of any outstanding bugs (famous last words) so any further code tidy up will be submitted as follow on PRs

I'm not sure I understand the second problem that is being fixed.  (I thought that PR #931 fixed this a couple of weeks ago - or is this a different issue?)

I was hoping to release the next version pretty soon.  So, I'm tempted to leave this until after the release.

ddrury:
I hadn't noticed this PR but I was up to date with the master branch and it still failed. But in any case this is the area where I shifted a lot of processing from javascript to php so a few loops got removed and the gmarks array was made almost redundant (reincarnated as $events).

Also the children's births were tacked onto the end of the events list in the sidebar. With the new structure it was easy to sort then into their correct place.

I appreciate there is a lot of changed code here - when (if) you accept it is up to you.

fisharebest:
> I appreciate there is a lot of changed code here

Too many changes for github.  It can't display the changes, which means I can't comment on them.

I've merged a few more bits, to reduce the noise.  Here are a comments.

* `glob()` sorts its output by default, so no need to sort.

* You have taken the space out of `function() {}` in many places.  Was this intended?

Replacing `echo '<x>', $foo, '</x>';` with `printf('<x>%s</x>');` goes against the existing code style.  It may be more efficient, but the long-term goal is to use templates/views, and the former construct is closer to the target of `<x><?= $foo ?></x>` or `<x>{{ $foo }}</x>`

ddrury:
> You have taken the space out of function() {} in many places. Was this intended?
I seem to remember removing  them to fix some of Scrutiniser's issues - I'll put them back.

> Replacing echo '<x>', $foo, '</x>'; with printf('<x>%s</x>');

I addopted that style as I always had problems with single/double quotes around the echoed bit - I'll revert them

fisharebest:
Not being able to add line-comments to the changes is somewhat frustrating...

* The docblock for `printGoogleMapMarkers()` looks wrong.
* Stray newlines and space-indents.
* If we're going to rewrite comma-separated lists, a better way to do it is to create an array, and then use `implode(I18N::$list_separator, $array)`, since not all languages use `', '` to separate list items.
* I'm not sure that `Session::get('gmlocation')` is the right approach.  If you have two tabs/windows open, one in each context, then the return from one will overwrite the other...  Perhaps just an extra form field?
* Output from `glob()` is already sorted.  No need to sort it again?

Can you rebase this branch against master?  Hopefully the number of changes are now under the github threshold...

PS - You've uncommented the block "BH not sure yet what this if statement is for".  Does this mean you've worked out what Brian's code was doing ;-)

ddrury:
>  Does this mean you've worked out what Brian's code was doing ;-)

Uncommented by mistake - I still haven't got a clue what it's supposed to do ;-(

> If we're going to rewrite comma-separated lists

If I understand you correctly, you want `echo $x, %y;` to be replaced by `echo $x . $y;`

The other stuff I'll work through and fix

fisharebest:
> If I understand you correctly, you want `echo $x, %y;` to be replaced by `echo $x . $y;`

Nope.  I have no strong preference for ` . ` versus `, ` in this context.

I meant generting human-viewable lists.  e.g. "England, Scotland, Wales."

Here, the ", " is a language depenedent string.  Non-latin alphabets don't always use commas.

You modified some code to follow this pattern:

```
$countries = '';
$countries .= 'England';
$countries .= ', ';
$countries .= 'Scotland';
$countries .= ', ';
$countries .= 'Wales';
$countries .= ', ';
echo trim($countries, ', ');
```

Instead, use the localised list separator.  `implode()` is useful here, instead of adding and trimming commas.

```
$countries = [];
$countries[] = 'England';
$countries[] = 'Scotland';
$countries[] = 'Wales';
echo implode('I18N::$list_separater, $countries);
```

ddrury:
Oh OK - that saves a bit of work

ddrury:
Am I correct in thinking that the separator between place parts should be I18N::$list_separator ? If so there's a bit more work than first thought because they're passed from function to function and I'll need to figure out what's going on - the simple ones I've fixed.

BTW would the attached file be of any use? Its a compressed diff of master against the recent changes I've made (not yet pushed so don't use it in anger)
[diff.zip](https://github.com/fisharebest/webtrees/files/288016/diff.zip)

ddrury:
Changes made as requested. The diff is now visible.

BTW I've found another bug - If there is only one place to show on the pedigree map then the zoom level is so great that you think the map drawing has failed - for another day

app/Module/GoogleMapsModule.php
```diff
@@ -1609,8 +1624,8 @@ private function adminPlaceCheck() {
 		$x   = 0;
 		$max = 0;
 		while ($x < $i) {
-			$levels                 = explode(",", $place_list[$x]);
-			$parts                  = count($levels);
+			$levels = explode(I18N::$list_separator, $place_list[$x]);
```
fisharebest: The `$list_separator` can only be used to **create** strings from lists.
It cannot be used the other way around.  For example, east asian languages do not use "western" puncutation, and the list separator may be a space or even an empty string.  This usage appears to be an internal (i.e. GEDCOM format) string, so the literal `', '` is probably correct here.

ddrury: Bugger! - I'll fix it

ddrury:
In functions setPlaceIdMap & setLevelMap, I've reverted to the old explode by ',' as by tracing the function calls $parent starts out as an array of place parts which is then imploded with ',' and as far as I can see is only used internally (I hope)

fisharebest:
I've merged in some more of this.  Can you rebase against the latest master?  Thanks!

fisharebest:
I've merged in a load more.  I think just the API / key stuff left.  Can you rebase again?

app/Module/GoogleMapsModule.php
```diff
@@ -3147,19 +3160,18 @@ private function placesEdit() {
...
-				Filter::get('svzoom'),
+				(float) Filter::post('sv_latiText'),
+				(float) Filter::post('sv_longText'),
+				(float) Filter::post('sv_bearText'),
+				(float) Filter::post('sv_elevText'),
+				(float) Filter::post('sv_zoomText'),
 				$placeid,
 			));
-			$controller->addInlineJavascript('window.close();');
+			header('Location: ' . Filter::post('destination', null, 'index.php'));
```
fisharebest: This change isn't working for me.  I get a "headers already sent" error, due to the earlier `$controller->pageHeader();`.

I've submitted a fix for this.

app/Module/GoogleMapsModule.php
```diff
@@ -4692,8 +4704,10 @@ function initialize() {
 			//--------------------------------------------------------------------------------------
 			var street = new google.maps.ImageMapType({
 				getTileUrl: function(coord, zoom) {
-					var X = coord.x % (1 << zoom);  // wrap
-					return 'https://cbk0.google.com/cbk?output=overlay&zoom=' + zoom + '&x=' + X + '&y=' + coord.y + '&cb_client=api';
+					//var X = coord.x % (1 << zoom);  // wrap
+					//return 'https://cbk0.google.com/cbk?output=overlay&zoom=' + zoom + '&x=' + X + '&y=' + coord.y + '&cb_client=api';
+					//return "http://tile.openstreetmap.org/" + zoom + "/" + X + "/" + coord.y + ".png";
+					return '';
```
fisharebest: This seems to be creating an empty overlay `street`, which we add to the map a few lines further down.

Removing it doesn't seem to break anything.  Do you know what it is trying to do?  Perhaps this is (was) the code that created the "blue lines" when drag the "pegman" over a road?

ddrury: I actually sent you an email a few weeks ago asking you about this. I can only find a few references to cbk0.google.com and they were about obtaining metadata - it seems that this url has completely gone now. I have no idea at all what it was doing and that is why I left it commented out just in case someone else could shed any light.

fisharebest: let's get rid of it...

fisharebest: Gone...

fisharebest: I've fixed the blue lines overlay.
Only 2 lines of code.  Thank-you google ;-)

ddrury: Clever boy :-)

fisharebest:
I've merged in the rest of the changes.

* Is the "v=3" parameter needed?  I don't see it in the API docs, and it works without it.
* I updated the URL for the API script to match the documentation.
* I finally worked out what the $levelo stuff was doing - (nothing!)
* FYI, I (temporarily) commented out the API key option.  We'll need a non-technical explanation (plus instructions on how to get a key?)  Do we know the limits for non-key use?  I can't find it online.

ddrury:
> Is the "v=3" parameter needed?

See https://developers.google.com/maps/documentation/javascript/versions#version-rollover-and-version-types

> Do we know the limits for non-key use?

See https://developers.google.com/maps/faq#usage-limits

This was a big and complex PR - I apologise for putting you through so many hoops - next time (and there is one*) I'll not try and "eat the elephant in one bite".

* I've simplified the HTML for the individual tab, removing all tables bar the one for the events sidebar. BTW One of the dimensions options is unused (width I think offhand) should it be removed?

fisharebest:
Thanks.  Your "google-fu" is obviously better than mine ;-)

PR #958, opened 2016-06-05
Google maps - simplify HTML on individual tab

ddrury:
Uses "display: flex" - tested on Firefox & Chrome latest versions and IE 8 - 11 & edge (desktop & mobile versions using edge's emulation function - thought ie 8&9 should fail but they don't) - Also note IE 8 & 9 display a Googlemaps api warning about unsupported browser.

Quite easy to revert to floats if preferred.


ddrury:
Oops - I've not merged properly - I'll do it again

Issue #983, opened 2016-06-12
Undefined index in new Google Maps code


fisharebest:
I'm seeing lots of "undefined index: events" errors in line 2039 of the GM module.

$unique_places[$index]['events'] .= $evtStr;

~~Possibly~~ triggered by an individual with no place location data...

fisharebest:
@ddrury - can you take a look at this one?

ddrury:
Ok

fisharebest:
This fixes the bug - but I think I see two issues.

1) places without co-ordinates get a flag a 0N 0W.

2) where an individual has more than one event in the same place, but the events have specific co-ordinates, then the last co-ordinate overwites the first. e.g. a person born and buried in the same village, with exact co-ords for the birth and burial will only get one flag - at the place of burial.

PR #985, opened 2016-06-12
 Googlemaps bug fix "undefined index: events"

ddrury:
Fix for issue #983 

provide a basic set of parameters for a location that will be overwritten by a detailed set when coordinates exist for that location


fisharebest:
This fixes the bug - but I see two issues.

1) places without co-ordinates get a flag a 0N 0W.

2) where an individual has more than one event in the same place, but the events have specific co-ordinates, then the last co-ordinate overwites the first.  e.g. a person born and died in the same village, with exact co-ords for the birth and death will only get one flag - at the place of death.

ddrury:
Ok
I'll try to have another look tomorrow

Regards
David

PR #990, opened 2016-06-13
Googlemaps bug fix "undefined index: events"

ddrury:
Further change to ensure that places derived from facts where the co-ordinates are stored in the Gedcom do not get over-ridden


ddrury:
The diff of this change is very confusing but in essence I've extracted the code that produces the $unique_places into a function and ensured that places from facts with embedded co-ordinates are given a unique index value.

BTW whay does github say that "This branch has conflicts that must be resolved"

fisharebest:
> BTW why does github say that "This branch has conflicts that must be resolved"

Your PR has probably changed a line of code that was also changed in the master branch.

You need do something like `git rebase origin/master` and then manually fix the conflicts.  Conflicts will be marked in the code using `<<<<<<`, `======`, and `>>>>>>`

ddrury:
Hang fire on this - I think it can be simplified

ddrury:
OK, maybe not simplified, but will sensibly group and identify multiple events with identical co-ordinates regardless of where those co-ordinates are stored

fisharebest:
FYI, can you check your PhpStorm config.  You are removing spaces that should probably remain.

e.g. 9681a6a7e0082b49ca32918ae4619ad1f8b4b2c5

ddrury:
I'm not sure where to fix that - I'm assuming you mean for example line 
2069 "functiongetMarkerImage(iconColor){"
the space between the closing bracket of the function call and the left 
brace, my settings do set that but in this case it's in a string so I 
don't know when it got removed.

Regards
David



fisharebest:
>I'm not sure where to fix that 

Perhaps:
Settings -> Editor -> Code Style -> JS -> Spaces

ddrury:
That setting was incorrect so hopefully phpStorm is clever enough  to 
apply it in a string

Thanks

Regards
David

PR #992, opened 2016-06-14
Pedigree map simplify html - remove table

ddrury:
A simple one! (and the last one - for now)

### DIFF
--- a/app/Module/GoogleMapsModule.php
+++ b/app/Module/GoogleMapsModule.php
@@ -852,34 +852,20 @@ class GoogleMapsModule extends AbstractModule implements ModuleConfigInterface, 
 		//<!-- end of count records by type -->
 		//<!-- start of map display -->
 		echo '<div id="pedigreemap_chart">';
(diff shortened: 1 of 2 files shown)


ddrury:
Whoops big mouth - there is currently a css conflict with the name accordion - investigating

fisharebest:
> A simple one!

Yay!

Before you do any more, I want to refactor the CSS.

classes should be used for styling (even if there is just one of the item, such as side_bar).
ids should be used to identify specific instances of identically styled elements.

So, many of the IDs should be classes.  Also, classes such as `link` aren't very helpful, and are likely to clash elsewhere.  I'm going to use a `gm-` namespace and more descriptive names.

ddrury:
OK,

I did have a further thought whilst laying in bed last night, seeing as 
you've accepted "display: flex" it may make the css for the tab pop up 
infowindow & pedigree map side bar simpler to use it there as well.

I'll have a tinker but not submit anything until you've refactored stuff 
(if its indeed practical)

Regards
David

PR #1006, opened 2016-06-18
Show old style sources in facts

tronsmit:
This attempts two things:

1. If any old style sources to facts were imported from a GEDCOM file, this will display the complete source_description record. Currently only the first line (level 2) is used.

2. The source_description is filtered through formatText to allow markdown. This is important to those who implement history files across multiple trees. In this case it is not guaranteed that sources have the same reference number or even that they exist. The latter case results in a page load error just due to the presence of another tree's histo file. One solution would be to use old style source citations. Even though they are not encouraged in the GEDCOM speciication this would be acceptable as they will not form part of a GEDCOM export.  Allowing markdown with old style sources in history files will make it possible, for example, to use clickable links to sources.

The only drawback of my implementation (2) is that it results in the source details being displayed on a new line underneath the "Source:" label.

### DIFF
--- a/app/Functions/FunctionsPrintFacts.php
+++ b/app/Functions/FunctionsPrintFacts.php
@@ -590,11 +590,14 @@ class FunctionsPrintFacts {
 		$nlevel = $level + 1;
 
 		// -- Systems not using source records [ 1046971 ]
+		// -- The old style is not supported when entering sources, but if found in the GEDCOM then display them fully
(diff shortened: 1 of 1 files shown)


fisharebest:
Thanks Tronsmit.  I have merged this - but made some changes.  The `getCont` and `getSubRecord` functions are part of some legacy code that I am trying to remove.  I think the only place it is still used is the reports.

I've replaced them with some regular expression matching.

PR #1008, opened 2016-06-18
Googlemaps change missed in last commit

ddrury:
Pedigree map: side bar icon & relationship text should be a link to that person's pedigree map. (lost in last commit)


ddrury:
withdrawn - error

PR #1009, opened 2016-06-18
Googlemaps change missed in last commit

ddrury:
Pedigree map: side bar icon & relationship text should provide a link to that person's pedigree map. (lost in last commit)


ddrury:
BTW was it a deliberate decision to remove the persons image from the popup info-window of the google maps individual tab?

PR #1022, opened 2016-06-21
Refactor Google Maps pedigree map sidebar html

ddrury:
- Rearrange pedigree map html to simplify on click management
- Item 1 provides opportunity to simplify CSS by using display: flex


fisharebest:
Merged.  FYI, I replaced the `id` attributes with `data` attributes.

ddrury:
Unfortunately you've missed a few instances where we change classes 
based on the id (eg line 1181 amongst others)

Regards
David



fisharebest:
Oops.  Thanks.

We can select elements using `.gm-ancester[data-target=12]`, but for the time being, I have just reverted the commit.

PR #1032, opened 2016-06-24
Google Maps enhancement

ddrury:

Uses a small javascript library [Overlapping Marker Spiderfier](https://github.com/jawj/OverlappingMarkerSpiderfier) to expand markers in a spiral centred on the co-ordinates when the marker is clicked - implemented for the Individual tab and the pedigree map.


fisharebest:
The markers have tool-tips (the name of the individual).
For the "multiple" markers, we see the tool-tip of just one of the names.

Is there a way to give different tool-tips to these multiple markers.  Maybe "5 individuals"?
If not, then would it be less confusing not to have them at all?

PS - there is a `.php_cs` config file for php-cs-fixer.  `php-cs-fixer fix` would tell you that there needs to be a blank line before `return` statements.

ddrury:
> For the "multiple" markers, we see the tool-tip of just one of the names.

I was already investigating the possibility of having a different icon for the groups (where the tooltip would clearly need removing) - apparently it can be done but will need a bit of work on my part.

> PS - there is a .php_cs config file for php-cs-fixer. php-cs-fixer fix would tell you that there needs to be a blank line before return statements.

I just tried this and it trampled all over the code - clearly I did something wrong :-(

fisharebest:
> I just tried this and it trampled all over the code - clearly I did something wrong :-(

Odd.  When I ran it, it just added two missing blank lines.

ddrury:
Will be resubmitted from different branch

PR #1035, opened 2016-06-25
Make historical facts persistent

### LINKED ISSUE #1031: History facts
In app/Module/IndividualFactsTabModule.php
line 139 perhaps should be _input id="checkbox_histo"_...
instead of _input id="checkbox_rela_facts"_...

And it would be nice that the « unchecked » option (for both _histo_ and _relatives_)  is activated by default.

tronsmit:
fixes #1031


fisharebest:
Thanks!

PR #1056, opened 2016-07-05
Google Maps enhancement

ddrury:

Uses a small javascript library Overlapping Marker Spiderfier to expand markers in a spiral centred on the co-ordinates when the marker is clicked - implemented for the Individual tab and the pedigree map.

**THIS IS NOT PRODUCTION READY**

Unfortunately (for you) I ended up rewriting functions pedigreeMap() & pedigreeMapJavascript().

The javascript used to handle the marker clustering has only fully been implemented for the pedigree map (I have a feeling there is currently a lot of duplicated javascript between the individual tab & the pedigree map). It now handles in a fairly intuitive way the explode / implode of the icons. The only thing extra I'd like to have done is to maintain the icons in their expanded state when an infowindow popup is open but a bug in the library sometimes fails to render the correct marker icon so I allow the icons to collapse to their clustered state.

If you like the implementation I'll address the Individual Tab code


ddrury:
Major bug loading scripts fails for place hierarchy

PR #1057, opened 2016-07-06
Google Maps enhancement

ddrury:

Uses a small javascript library Overlapping Marker Spiderfier to expand markers in a spiral centred on the co-ordinates when the marker is clicked - implemented for the Individual tab and the pedigree map.

**THIS IS NOT PRODUCTION READY**

Unfortunately (for you) I ended up rewriting functions pedigreeMap() & pedigreeMapJavascript().

The javascript used to handle the marker clustering has only fully been implemented for the pedigree map (I have a feeling there is currently a lot of duplicated javascript between the individual tab & the pedigree map). It now handles in a fairly intuitive way the explode / implode of the icons. The only thing extra I'd like to have done is to maintain the icons in their expanded state when an infowindow popup is open but a bug in the library sometimes fails to render the correct marker icon so I allow the icons to collapse to their clustered state.

If you like the implementation I'll address the Individual Tab code


ddrury:
... And whilst we're at it, have you any idea why we store two lots of co-ordinates? - one for the map and the other for the streetview.

fisharebest:
> have you any idea why we store two lots of co-ordinates?

This was Brian's design.  I guess that the pin/flag would go at the centre of the town, but the streetview could be anywhere - e.g. a famous landmark or a "Welcome to XXX" sign at the edge of town.

fisharebest:
> THIS IS NOT PRODUCTION READY

Are you telling me not to merge this yet?

BTW - removing the shadow markers would have been good as a separate PR, and would make it easier to focus on the main changes.

ddrury:
> Are you telling me not to merge this yet?

Only from the point of view that I haven't done any more work on the individual tab code. If you're happy for me to submit a later PR to make the clustering/declustering of the icons look the same as the pedigree map then you can merge this now.

ddrury:
Hang fire - more work needs doing - some click sequences produce wieird results

ddrury:
I think I've now sorted out the various click sequences.
Re Pedigree map

- I'd appreciate your view on removing the click functionality from the sidebar (for simplicity comment out line 1065)

- Could you look at lines 789 & 790 (Am I missing something here - $person->canShowName() is the right way to go isn't it?)

fisharebest:
> $person->canShowName() is the right way to go isn't it?

Yes - but I'm not sure we should be displaying a count of private individuals at all.  Elsewhere, we go out of our way not to leak the existance of private data.

app/Module/GoogleMapsModule.php
```diff
@@ -832,403 +856,217 @@ private function pedigreeMap() {
...
-
-		$js = '
-		// this variable will collect the html which will eventually be placed in the side bar
-		var gm_ancestors_html = "";
-		// arrays to hold copies of the markers and html used by the side bar
-		// because the function closure trick doesnt work there
-		var gmarkers = [];
-		var index = 0;
-		var lastlinkid;
+	private function pedigreeMapJavascript($marker_data = array(), $link_lines = array()) {
```
fisharebest: The parameters are always provided, so don't need the default `= array()`.
Add parameter types to PHPDoc.

app/Module/GoogleMapsModule.php
```diff
@@ -233,7 +230,7 @@ public function getBoxChartMenu(Individual $individual) {
 	 * A form to edit the module configuration.
 	 */
 	private function config() {
-		$controller = new PageController;
+		$controller = new PageController();
```
fisharebest: Did PhpStorm do this for you, or are you proposing to change the code style.

I note that we are currently inconsistent in this.  I *think* I prefer `new foo;` to `new foo();`.

But if we are going to reformat all of these, it should probably happen in a separate PR.

ddrury:
> I'm not sure we should be displaying a count of private individuals at all

Thats easy to remove

> The parameters are always provided, so don't need the default = array().

I've been working towards using the same javascript for the individual tab & the pedigree map and put this in for defaults - you're right they're not necessary - I'll remove them

> Did PhpStorm do this for you (code style)

Yes - I'll revert them

app/Module/GoogleMapsModule.php
```diff
@@ -1077,158 +917,144 @@ function createMarker(point, name, html, mhtml, icontype) {
...
+		var plines;
+		var colors     = ['#000000', '#FF0000', '#0000FF', '#00FF00', '#FFFF00', '#00FFFF', '#FF00FF', '#C0C0FF', '#808000'];
+		jQuery.each(<?php echo json_encode($link_lines); ?>, function(index, item) {
+			line   = [
+				new google.maps.LatLng(item.parent_lat, item.parent_lon),
+				new google.maps.LatLng(item.child_lat, item.child_lon)
+			];
+			plines = new google.maps.Polygon({
+				paths:         line,
+				strokeColor:   colors[item.generation],
```
fisharebest: If `generation` is more than the size of the colors array, this will error.
Perhaps use something like `colors[item.generation % colors.length]`

ddrury: Done

ddrury:
I'm going to withdraw this for now - I'd like to rationalise some of the Javascript first and then re-present this.

PR #1066, opened 2016-07-14
markerImage class removed in Google Maps api v3.11

ddrury:


app/Module/GoogleMapsModule.php
```diff
@@ -2346,12 +2394,11 @@ private function printHowManyPeople($level, $parent) {
 	 * @param string   $linklevels
 	 */
 	private function printGoogleMapMarkers($place2, $level, $parent, $levelm, $linklevels) {
+		echo 'var icon_type = null;';
 		if (!$place2['lati'] || !$place2['long']) {
-			echo 'var icon_type = new google.maps.MarkerImage();';
```
fisharebest: The changes in this function look suspicious.  i.e. it's not the same changes you were making elsewhere.  `icon_type` is becoming a string, not an object?

You've removed the declaration of `point`, but still use it in the next line.  Is this right?

ddrury:
The logic is correct. Originally a markerImage object was created but never really used - the icon_type.image (which was just the url to the image) was extracted to create a new object in function createMarker. I now don't bother creating an object and just pass in the url as a string (I've just renamed variable icon_type to icon_url to make it a bit clearer)

PR #1073, opened 2016-07-22
Googlemaps problem with streetview co-ordinates

ddrury:
For the Individual tab

If map co-ords are stored in the fact, currently function getPlaceData() returns empty streetview co-ordinates. We could use the place co-ordinates for the streetview but it is possible that they refer to somewhere not on a street, therefore this mod obtains the streetview co-ords from the placelocation table irrespective of the source of the map co-ords.

(this does sometimes result in the streetview display not quite matching the place location, but the alternative is a blank window)


ddrury:
More investigation needed

PR #1074, opened 2016-07-22
Google Maps enhancement

ddrury:
Sorry this is a big one - I couldn't see any way to split this into smaller chunks
1. Simplified the Pedigree Map class switching that indicates the visited state of a node
2. Extract the javascript for the Individual Tab and Pedigree Map into a separate file, make common and use for both maps - Includes the Overlapping Marker Spiderfier library for clustering markers located at the same (or very close to same) co-ordinates.

This is stage one - next step is to extract the javascript for the Administrative and Place Hierarchy maps (I don't think these will be straightforward)


ddrury:
Greg,

For some time I've been thinking that using a function that shows modern street images on a site that is fundamentally historic in nature is pointless and add that to the fact that currently my time is limited means that I've decided not to progress any further with this work. This mod as it stands is perfectly functional but while the javascript for the individual tab and pedigree map have been extracted into a separate file that for the place hierarchy & admin functions are still embedded.

As always the decision on whether to use this PR is yours and I'll leave it here for a while longer.

fisharebest:
I agree 100% about streetview.

It also prevents us from switching easily to another mapping provider such as OSM.

ddrury:
I'm closing this as with all the recent changes it wouldn't slot in without further work. I'll keep the branch locally just in case...

PR #1082, opened 2016-08-04
Move Zend_Mail to SwiftMailer & Remove Zend Framework

jon48:
Just a suggested PR, so please do not hesitate to reject it.
Given the expressed will to go away from the Zend framework, that the email is the only feature still using it, and that SwiftMailer is the underlying library used in Symfony/Laravel, I thought it would be a good exercise to move the Mail class to SwiftMailer, and to completely remove the dependency on Zend.

I have not tried rewriting the Mail class, or add tests (not very testable as it is, but I am not sure how much effort it is worth, as it may be completely reviewed when/if moving to a more modern architecture/framework), just replacing the Zend methods by the SwiftMailer ones.
I have tested plain & authenticated mode. I have as well tested partly Gmail over TLS, but had to work a bit around some issues with certificates (PHP 5.6+ being stricter - for good reasons), so further testing would be necessary on a completely working installation.


fisharebest:
Thanks Jon.  This is exactly the change I wanted to make here!

PR #1092, opened 2016-08-28
Stay in the current tree when logging out

JustCarmen:
When logging out a user is always redirected to the homepage of the default tree in stead of to the homepage of the tree the user logged out from. I think the latter should be the default behaviour.

Also see: https://www.webtrees.net/index.php/en/forum/help-for-ver-1-7-7/31527-solved-default-tree#60645

### DIFF
--- a/logout.php
+++ b/logout.php
@@ -15,6 +15,13 @@
  */
 namespace Fisharebest\Webtrees;
 
+/**
+ * Defined in session.php
+ *
+ * @global Tree $WT_TREE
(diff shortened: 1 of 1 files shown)


fisharebest:
This code will fail if no tree exists (e.g. during install / setup, before a tree is created), as`$WT_TREE` will be `null`.

I've submitted an alternative solution.

JustCarmen:
Tnx.

PR #1134, opened 2016-12-04
Fix PHP7.1 error

ddrury:
Fixes error

>  [] operator not supported for strings

I'm sure there are more!

### DIFF
--- a/app/Stats.php
+++ b/app/Stats.php
@@ -634,23 +634,20 @@ class Stats {
 	 */
 	public function totalSurnames($params = array()) {
 		if ($params) {
-			$qs       = implode(',', array_fill(0, count($params), '?'));
-			$opt      = "IN ({$qs})";
-			$vars     = $params;
+			$opt      = 'IN (' . implode(',', array_fill(0, count($params), '?')) . ')';
 			$distinct = '';
 		} else {
 			$opt      = "IS NOT NULL";
-			$vars     = '';
 			$distinct = 'DISTINCT';
 		}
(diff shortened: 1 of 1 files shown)


fisharebest:
Are you sure this is the

fisharebest:
I guess you are refering to this: http://php.net/manual/en/migration71.incompatible.php

AIUI, it means you cannot convert a string into an array using `[]`, and this won't work any more:

```
$x = 'hello';
$x[] = 'world';
print_r($x); // array ('hello', 'world')
```

Instead, `$x` must always be an array before you can use `$x[]` to append to it.

I'll need to find a PHP7.1 install and test this.

FYI, the continuous integration on travis-ci.org is already running our tests against PHP7.1.  Admittedly, we only have about 5% coverage at the moment.

app/Stats.php
```diff
@@ -634,23 +634,20 @@ public function totalRepositoriesPercentage() {
...
 	public function totalSurnames($params = array()) {
 		if ($params) {
-			$qs       = implode(',', array_fill(0, count($params), '?'));
-			$opt      = "IN ({$qs})";
-			$vars     = $params;
+			$opt      = 'IN (' . implode(',', array_fill(0, count($params), '?')) . ')';
 			$distinct = '';
 		} else {
 			$opt      = "IS NOT NULL";
-			$vars     = '';
```
fisharebest: I think the fix is to use `$vars = [];`.

fisharebest:
Ah, yes.  See my comment on the code.

It is a bug, but the fix is much simpler.

PR #1136, opened 2016-12-05
Feature/proposal census

jon48:
2 proposals:

- In `CensusColumnNationality`, add a suggested nationality rule if no birth place is found: use the census country as a default (+ small improvement to reuse an existing function to extract the country from the place). I appreciate this rule may work better for populations that do not have an important international immigration history (like Europe), maybe not so for countries like US.

- In `CensusColumnRelationToHead`, translate the term _head_ (otherwise it display in English, whereas other relationships to the head are translated in the user's locale). I noted that for German Census used a derived class `CensusColumnRelationToHeadGerman` to translate the term (but it is fixed as well). I am not sure what is the best approach: consistency of the language between _head_ and other relationships, or use of the local Census language term for _head of family_ (but there may be a better way than rewriting the whole `generate` method in sub classes).
(As an aside note, I tried to incorporate better coverage for the `CensusColumnRelationToHeadTest`, but PhpUnit does not play well with the I18N implementation, as we cannot mock static class, and it does not have a bootstrap script to load the minimal session data)

### DIFF
--- a/app/Census/CensusColumnNationality.php
+++ b/app/Census/CensusColumnNationality.php
@@ -17,6 +17,7 @@ namespace Fisharebest\Webtrees\Census;
 
 use Fisharebest\Webtrees\Date;
 use Fisharebest\Webtrees\Individual;
+use Fisharebest\Webtrees\I18N;
 
 /**
  * The nationality of the individual.
@@ -40,9 +41,11 @@ class CensusColumnNationality extends AbstractCensusColumn implements CensusColu
 			}
 		}
 
-		$place = explode(', ', $place);
-		$place = end($place);
-
+		// By default, we can assume the nationality is the census's one
+		if(empty($place)) $place = $this->place();
+		
+		$place = $this->lastPartOfPlace($place);
+		
 		if ($place === 'England' || $place === 'Scotland' || $place === 'Wales') {
 			return 'British';
 		} else {
--- a/app/Census/CensusColumnRelationToHead.php
+++ b/app/Census/CensusColumnRelationToHead.php
@@ -17,6 +17,7 @@ namespace Fisharebest\Webtrees\Census;
 
 use Fisharebest\Webtrees\Functions\Functions;
 use Fisharebest\Webtrees\Individual;
+use Fisharebest\Webtrees\I18N;
 
 /**
  * Relationshiop to head of household.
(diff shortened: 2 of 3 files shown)


mpwt:
Hi, I made the same sugestion for "head" some weeks ago or I was surprised about the hard coded word too. But it is right to let it untranslated because the census assistant is a transcription of real census forms and there is written a word in english, french or german etc. This word should not be translated.
Therefore I understand that all different censuses have to use a different relation to head file.

fisharebest:
> I was surprised about the hard coded word too.

If a French user enters a German census, then always we want "head" to be in German.
We do not want to translate this into the user's language (French).

Perhaps an alternative solution is that each census should specify the language it uses.

We then switch to that language while processing the columns.

So, a German census specifies 'de', and we call `I18N::init('de')` before we start...

fisharebest:
Or perhaps each census defines the text it uses for "Head of household", and the "relation to head" function uses that?

mpwt:
:) First I was surprised and made my suggestion like jon48. After I understand you (Greg) and the census assistant solution. :)

Please, do not make to much work for this single table cell. Your last suggestion sounds perfect but the user can overwrite all table cells in the moment of using the census assistant if any automatically filled cell is not perfect.

fisharebest:
> Please, do not make too much work for this single table cell.

I think it is also a problem for the relationship column.

It will be in the user's language - not the census langauge.

fisharebest:
> 2 proposals:

Two things in one issue/PR are always difficult to manage.

I've merged the nationality change.

I added a lookup to convert country to nationality - but I'm not sure if this will work.  We may need to translate the nationality...

jon48:
> Two things in one issue/PR are always difficult to manage.

Sorry for that. When I committed I thought I could open 2 different pull request from the two commits, but apparently no.

> I've merged the nationality change.

Thanks for that and for the additions, 

As for the `CensusColumnRelationToHead` question, it indeed raises few questions for a simple cell: my initial concern was that when entering data for a French census in a French interface, it is a bit odd to see  the value _head_, while other relationships are translated.
I will try to have a play with your different suggestions to keep the census's language for the values.

fisharebest:
Relationship names are now translated according to the language of the census.

PR #1155, opened 2016-12-27
Set min zoom level for all map types

ddrury:
For all map instances, ensure default value for min zoom level is set even if module setting is missing. 

Despite this fix it is still possible to get more than one world displayed when the width to height ratio of the map div is very large (very wide browser windows).

I suppose we could calculate the minzoom based on that ratio but I'm not sure it's worth the effort. This suffices for most practical purposes

### DIFF
--- a/app/Module/GoogleMapsModule.php
+++ b/app/Module/GoogleMapsModule.php
@@ -61,6 +61,14 @@ class GoogleMapsModule extends AbstractModule implements ModuleConfigInterface, 
 	const SCHEMA_SETTING_NAME     = 'GM_SCHEMA_VERSION';
 	const SCHEMA_MIGRATION_PREFIX = '\Fisharebest\Webtrees\Module\GoogleMaps\Schema';
 
+	const GM_MIN_ZOOM_MINIMUM = 1;
+	const GM_MIN_ZOOM_DEFAULT = 2;
+	const GM_MIN_ZOOM_MAXIMUM = 14;
+
+	const GM_MAX_ZOOM_MINIMUM = 1;
+	const GM_MAX_ZOOM_DEFAULT = 15;
+	const GM_MAX_ZOOM_MAXIMUM = 20;
+
 	/** @var Individual[] of ancestors of root person */
 	private $ancestors = [];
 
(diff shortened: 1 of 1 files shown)


fisharebest:
> even if module setting is missing

Settings should always be present?  It is set in one of the migration scripts https://github.com/fisharebest/webtrees/blob/master/app/Module/GoogleMaps/Schema/Migration5.php

Perhaps this is broken for some people/installations/upgrade paths?

+1 for replacing magic numbers with constants!

There's another magic number `6`, as in `MAX_ZOOM - 6`.

Perhaps get rid of it by having separate constants to represent the minimum/maximum/default value of both parameters?  e.g. something like

```
const GM_MIN_ZOOM_MINIMUM = 1;
const GM_MIN_ZOOM_DEFAULT = 2;
const GM_MIN_ZOOM_MAXIMUM = 14;

const GM_MAX_ZOOM_MINIMUM = 1;
const GM_MAX_ZOOM_DEFAULT = 15;
const GM_MAX_ZOOM_MAXIMUM = 20;
```

ddrury:
> Settings should always be present

OK, I did it as a bit of defensive programming - would you prefer that the defaults be removed?
I'll extend the constants as suggested and as we have a max_zoom preference, maybe it should also be set on each map instance (probably not as important as min_zoom) what do you think?

ddrury:
changes made as requested

fisharebest:
We have a lot of code related to zoom levels, which leads to a complicated UI in places.
I find it tedious to choose a zoom-level for each of the places in the GM module.

I wonder if we can simplify/automate this?

Perhaps calculate the zoom by looking at all the other places with the same parent place.
Then use a bounding rectangle that includes all of them.

ddrury:
Yeah, I agree, I think it should always be the case that on 
initialisation we set the min/max zoom levels and use the content to 
determine the actual zoom level (bounding box as you say)

I'd started to do this with the PR I recently withdrew. There is a huge 
amount of javascript "almost but not quite" duplication in the current 
code and it gets a bit impenetrable at times.

I'll have another rummage and see what falls out. I'd really like to get 
rid of the StreetView stuff - any chance?

BTW I've written an OpenStreetMap module using leaflet. Have a look at 
http://webtrees.drury.me.uk/individual.php?pid=I274&ged=drury.ged . The 
admin pages aren’t anywhere near complete yet!

Regards
David



fisharebest:
> I'd really like to get rid of the StreetView stuff - any chance?

Me too.  I'm off work for a week, so should have lots of time...

#1158

Amgine0:
OpenStreetMap module envy! pleaseohpleaseohplease?

Issue #1158, opened 2016-12-28
Remove Google Street View


fisharebest:
There are many criticisms of the Street View module.

* It shows the current view of the world, whereas genealogy is concerned with the historic view.  The current view often bears no relation to the historic view.
* Each time Google update their images, it invalidates existing coordinates/angles/elevations
* It ties us into Google Maps, and prevents us using other mapping providers, such as Open Street Map.
* The user interface is not intuitive, and the code behind it is convoluted and difficult to maintain.

I propose to remove StreetView for the 1.8.0 release

HonkXL:
In my opinion this is a good idea. I don't know any person using Street View here (but maybe they exist). Supporting Open Street Map would be great!

PR #1160, opened 2016-12-28
Remove all references to StreetView

ddrury:
1st step. No other functionality has been touched. There is some redundant code in places e.g. for the individual events map the streetview tab for the marker infowindow has been removed but the events tab is still there.


fisharebest:
I guess we both spent the evening doing the same thing - see #1158 ;-)

Let me review these changes to see if you found something that I missed...

fisharebest:
> Let me review these changes

Bar a couple of whitespace differences, it looks like we made identical changes!

PR #1198, opened 2017-01-30
Update Vagrantfile to use ubuntu/xenial64 with PHP 7 and required extensions

mnoorenberghe:
* wily64 gives a 404 and I think didn't support PHP 5.6
* Some PHP extensions were missing
* make and (x)gettext were missing
* The document root was fixed to be /var/www/html, not /var/www
* Path to PHP 7's php.ini was fixed

With the above changes I was able to run the setup wizard

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -10,7 +10,7 @@ Vagrant.configure(2) do |config|
 
   # Every Vagrant development environment requires a box. You can search for
   # boxes at https://atlas.hashicorp.com/search.
-  config.vm.box = "ubuntu/wily64"
+  config.vm.box = "ubuntu/xenial64"
 
   # Disable automatic box update checking. If you disable this, then
   # boxes will only be checked for updates when the user runs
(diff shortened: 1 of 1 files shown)


fisharebest:
The Vagrantfile is set to use the OLDEST version of PHP that we support.

I generally develop with the newest version, but it is necessary to use the old verison when running `composer update`.  Otherwise we'll get dependencies that won't run on old PHP.

fisharebest:
I added the Vagrant definition because `composer` did not have an option such as `--target-php-version=5.6`.

However, I see that the latest version *does* have this option: https://getcomposer.org/doc/06-config.md#platform

```
"config": {
    "platform": {
        "php": "5.6"
    }
}
```

Yay!  If this does what I think it does, then we no longer **need** Vagrant at all.

But standard dev environment is always useful.  What are your thoughts on https://laravel.com/docs/5.4/homestead

BTW - We should also add the libraries (e.g. `php/mbstring`) to `composer.json` and backport it to the 1.7 branch, which needs to support PHP5.3.

PR #1222, opened 2017-02-28
Fix issue 1221

jviksell:


mirono:
It is better to re throw the exception but without the sensitive information, and not to hide it

PR #1240, opened 2017-04-30
Distinguish main blocks from side blocks for easy reference

JustCarmen:
The new layout doesn't make it easy to distinguish the main part of the homepage/mypage from the side part. As a theme developer I need a reference to those parts for styling purposes (for example a smaller font for the side). At the moment the separate parts can only be referred to as .wt-main-container > row > col-sm-8 or .wt-main-container > row > col-sm-4 which is not very helpful and may lead to unwanted side effects. I don't think one should ever refer to grid classes directly for this kind of stuff.

I hope you will accept this change. Of course feel free to rename the classes to something else.

P.S. I am not sure if this is the way to go in this stage of development. If you rather want this kind of stuff just being reported as issue let me know.

### DIFF
--- a/index.php
+++ b/index.php
@@ -97,7 +97,7 @@ echo '<div class="row">';
 
 if ($blocks['main']) {
 	if ($blocks['side']) {
-		echo '<div class="col-sm-8">';
+		echo '<div class="col-sm-8 wt-main-blocks">';
 	} else {
 		echo '<div class="col-sm-12">';
 	}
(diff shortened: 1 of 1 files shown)


fisharebest:
> I hope you will accept this change.

I had already added these to the documentation (https://github.com/fisharebest/webtrees/blob/develop/themes/webtrees/css-1.7.8/style.css#L742-L746) but forgot to add the classes.

One thing.  If the main blocks (or the side blocks) are empty, should we also give a class to the other block (`col-sm-12`)?

> I am not sure if this is the way to go in this stage of development.

Issues and PRs are welcome!

JustCarmen:
> One thing. If the main blocks (or the side blocks) are empty, should we also give a class to the other block (col-sm-12)?

I didn't do it because I was in doubt. But rethinking it I think there might be situations where this can become handy. I think this part could also be called 'wt-main-blocks' or should it have a separate name?

> Issues and PRs are welcome!
Thanks. Good to know. What you have done so far looks really good!

PR #1245, opened 2017-05-13
Activate datatable for module table

JustCarmen:
The module table wasn't functioning as datatable. This changes fixes that. It also uses default paging and filter mode. 

There is a script available to get bootstrap4 layout in datatables, see:

https://datatables.net/examples/styling/bootstrap4.html

I think you should implement that.


fisharebest:
Thanks for the information about the new release of datatables.

PR #1247, opened 2017-05-25
Show configured family events in the individual box in pedigree chart…

jcnventura:
I wanted to be able to check if I already have a marriage record directly in the pedigree charts. Unfortunately even adding "MARR" to the "Other facts to show in charts" settings in admin_trees_config.php didn't display anything.

This is because, the MARR event actually belongs to the individual's family and not to the Individual object. I have added some code to add family events for these $opt_tags in the individualBoxFacts() function.

This code is working fine with webtrees 1.7.9, other than the fact that it often exceeds the individual's box max-height attribute. But this already happens anyway if you have multiple events in the WT_EVENTS_BIRT and WT_EVENTS_DEAT lists.

### DIFF
--- a/app/Theme/AbstractTheme.php
+++ b/app/Theme/AbstractTheme.php
@@ -886,11 +886,17 @@ abstract class AbstractTheme {
 		}
 		// Show optional events (before death)
 		foreach ($opt_tags as $key => $tag) {
+			$events = array();
 			if (!preg_match('/^(' . WT_EVENTS_DEAT . ')$/', $tag)) {
-				$event = $individual->getFirstFact($tag);
-				if (!is_null($event)) {
-					$html .= $event->summary();
-					unset($opt_tags[$key]);
+				$events[] = $individual->getFirstFact($tag);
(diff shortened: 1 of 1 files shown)


fisharebest:
> I wanted to be able to check if I already have a marriage record directly in the pedigree charts.

In all the charts, there is a "zoom" icon in the top right corner.

If you click this, you will see all the facts/events for the individual - including the marriage.

The information you want is available with one click.

> other than the fact that it often exceeds the individual's box max-height attribute

The boxes only have room for two facts.  I think it better to show birth/death than birth/marriage.

jcnventura:
The zoom button is useful, but on a 5 generation pedigree, clicking 31 times on it to quickly check if I have a marriage date is too much.

As I said, I've configured the site to show me MARR in other events. Nothing changes from the current functionality in the default configuration. It simply fixes one of the functions you already allow anyway. The 'bug' is that the MARR event can't be shown because the event discovery doesn't extend to the family objects.

PR #1252, opened 2017-06-19
Fix #1251 new module settings not saved

### LINKED ISSUE #1251: Saving module settings (setSetting vs setPreference) - 1.8.0-dev
It took me some time to figure out why some module settings weren't saved into the database anymore.

It turned out to be the difference between the old setSetting code and the new setPreference code. The old code checked if the module setting already existed. If not the setting was inserted into the database, otherwise it was updated.

The new getPreference code assumes the database setting already exist. It will only update an existing setting to a new value.

Is this intended? And if so, why and what is the best way to insert the setting the first time?

JustCarmen:
See #1251


fisharebest:
Thanks!

PR #1347, opened 2017-09-07
Medialist markup

JustCarmen:
I've made the cards inside the medialist all the same height. Further I've added card-header and card-footer markup. The latter are not absolutely necessary. I realise this is a matter of taste, so don't hesitate to change these to default card-blocks if that is what you prefer. What is important, is that the middle block must have the class card-body so it will be act as a flexbox which takes up all space by pushing the header block to the top and the footer block to the bottom.

Below a preview (webtrees theme)


JustCarmen:
Sorry wrong branch. Made a new pull request.

Issue #1270, opened 2017-07-02
pdf thumbnails return a 500 error


JustCarmen:
I have always used custom thumbnails for my pdf's. Uploaded with the extension pdf to match the main file. After the recent changes in the media area this no longer works. The thumbnails are not recognized anymore and return a 500 error.

fisharebest:
Hi Carmen,

I am working on two features to help this.

1) Automatically creating .JPEG files from the first page of a .PDF document.
This requires that the server has the php-imagick extension loaded.

2) Implement the GEDCOM feature that allows a media object to contain many media files.  For example, a media object could contain an audio file (an interview), a text file (a transcript of the interview) and an image (picture of the interviewee).

A "batch update" can easily be written for (2) which adds existing thumbnails linked to non-image files.

JustCarmen:
Ok, fine this sounds really good. The new features are certainly very welcome. And it is no problem to do a batch update on existing thumbnails.

fisharebest:
It is possible to create thumbnails from PDF files - but it requires the php-imagick library.

JustCarmen:
> A "batch update" can easily be written for (2) which adds existing thumbnails linked to non-image files.

Could you please explain this in more detail. I tried to rename a pdf thumbnail (which really is a jpeg image, but formerly needed to have the extension pdf to serve as a thumbnail for a pdf document) in the thumbs folder to 'jpg' but that doesn't work.

fisharebest:
> It is possible to create thumbnails from PDF files - but it requires the php-imagick library.

php-imagick also requires that `gs` is installed on the server, and that PHP has permission to execute processes on the server.  So, it won't work on >99% of servers.

fisharebest:
> I have always used custom thumbnails for my pdf's. Uploaded with the extension pdf to match the main file.

Can you tell me exactly what you did here?

e.g. if you have a file `xxx.pdf`, then what is your thumbnail called?

* `xxx.jpg` ?
* `xxx.pdf.jpg` ?
* `../thumbs/xxx.pdf.jpg` ?
???

JustCarmen:
> Can you tell me exactly what you did here?
e.g. if you have a file xxx.pdf, then what is your thumbnail called?

The thumbnail is called xxx.pdf while it is actually a jpg file. So the extension has been changed. These thumbnails are saved in the thumbs folder: .../thumbs/xxx.pdf

[...]

PR #1348, opened 2017-09-07
Medialist cards same height + card-header and card-footer markup.

JustCarmen:
I've made the cards inside the medialist all the same height. Further I've added card-header and card-footer markup. The latter are not absolutely necessary. I realise this is a matter of taste, so don't hesitate to change these to default card-blocks if that is what you prefer. What is important, is that the middle block must have the class card-body so it will be act as a flexbox which takes up all space by pushing the header block to the top and the footer block to the bottom.

Below a preview (webtrees theme)

### DIFF
--- a/medialist.php
+++ b/medialist.php
@@ -170,13 +170,14 @@ if ($action === 'submit') {
 
 	<div class="card-deck row mb-4 mt-4">
 		<?php foreach (array_slice($medialist, ($page - 1) * $max, $max) as $media): ?>
-			<div class="col=xs-12 col-sm-6 col-lg-4">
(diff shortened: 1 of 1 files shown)


fisharebest:
Thanks!

JustCarmen:
Yw :)

Fa10175:
I just tested this new presentation.
Information such as File Name, Size is accessible after clicking on the document name.
On the other hand if the document is of type jpg no problem if the document is of type pdf one has like message "error 500" before we had the symbol of a file pdf file
please see how we can no longer have "error 500" in this case

fisharebest:
@Fa10175 there is already a separate issue for this #1270

PR #1349, opened 2017-09-08
Pedigree chart arrows

JustCarmen:
This commit adds the arrows back into the pedigree chart. The placement of the arrows is done by using some bootstrap flexbox classes. Therefore I needed an extra variable to determine if the arrow should be placed next to, below or at the top of the person box.  I am not sure if the naming of this variable is correct according to webtrees standards, but you can rename it of course.

With this solution no extra css is required. However, I didn't remove the current styles from css because I am not sure if this will give conflicts elsewhere.

### DIFF
--- a/app/Controller/PedigreeController.php
+++ b/app/Controller/PedigreeController.php
@@ -102,20 +102,20 @@ class PedigreeController extends ChartController {
 		case self::PORTRAIT:
 			//drop through
 		case self::LANDSCAPE:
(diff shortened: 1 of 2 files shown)


fisharebest:
Looks good.  Can you fix the indentation (tabs/spaces).

makitso:
image
 markup not right.

JustCarmen:
I only have putted the arrows back in the chart.  I didn't do the popovers. But you are right, these need some attention.

makitso:
Ok, I will fix it.

PR #1353, opened 2017-09-09
Search forms additional styling

JustCarmen:
I did some additional styling on the search forms. Added the bootstrap form-controls to the fields and made input-groups where necessary (general form and replace form).

Further I've moved the buttons select all/select nothing to a imho better position. Placing them in the box with the tree selection makes it more clear these buttons belong to these checkboxes. See the preview below:

**Old**:

**New**:

I've also found some validation errors in the advanced search form (missing closing div and a closing div placed at the wrong position) which I've fixed.
Further I've removed all references to the old autocomplete function. I wasn't sure whether to implement select2 fields or not, so I haven't done that at the moment.

Another issue is the keyboard icon is not functioning. I don't know how to solve it.

### DIFF
--- a/search.php
+++ b/search.php
@@ -39,14 +39,14 @@ function checknames(frm) {
 		lname = frm.lastname.value;
 		place = frm.place.value;
 
-		if (year == "") {
+		if (year === "") {
 			if (fname.length < 2 && lname.length < 2 && place.length < 2) {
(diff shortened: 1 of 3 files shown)


makitso:
Search bar issue?
image

JustCarmen:
Strange, I see this:



But maybe this is due to the facts in your screenshot only to trees are shown and hence no select all/nothing buttons.

I will investigate this.

PR #1356, opened 2017-09-10
Search forms additional styling - a little fix

JustCarmen:
Style to set the padding on the wt-page-options-value moved to the common theme. All themes need this style to make sure the input and select fields in the same row will get the same height.

### DIFF
--- a/themes/_common/css-2.0.0/style.css
+++ b/themes/_common/css-2.0.0/style.css
@@ -170,6 +170,10 @@
 	text-align: center;
 }
 
+.wt-page-options-value {
+	padding: 0.25rem 0.75rem;
+}
+
 .wt-footer-container {
 }
 
--- a/themes/clouds/css-2.0.0/style.css
+++ b/themes/clouds/css-2.0.0/style.css
@@ -749,7 +749,6 @@ a > .wt-icon-arrow-up:hover::before {
 	color: #006;
 	background-color: #ecf5ff;
 	border: thin solid #ccc;
-	padding: 0.25rem 0.75rem;
 }
 
 .wt-page-content {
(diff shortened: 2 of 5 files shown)


fisharebest:
> to make sure the input and select fields in the same row will get the same height

Where is the height wrong?

JustCarmen:
In fact it is only wrong in the minimal theme (maybe my description wasn't correct). But then this style is added to all themes and should be moved to the common theme, right?

fisharebest:
> moved to the common theme

I am not certain that I want to keep this "common" file.

I added it for some work-arounds in bootstrap-4-alpha, etc.

I wonder if these are still needed.  For example, bootstrap-4-beta has many utility classes such as `d-flex`.  It may be better to add these to the HTML?

makitso:
> I am not certain that I want to keep this "common" file.
Ahhhh, I have been moving a lot of common style stuff out of the themes and into the common style.css. So, is this a bad thing???

JustCarmen:
>  am not certain that I want to keep this "common" file.

Are you sure you want to remove the common file? I thought it was added to make theming easier. If you want to modify a style you don't have to edit six stylesheets, but just one. 

> For example, bootstrap-4-beta has many utility classes such as d-flex. It may be better to add these to the HTML?

I think so, yes. I think you should use bootstrap classes where possible. If one want to overwrite a certain class for a custom theme that is always possible with normal css.
But it would be even nicer if we get the possibility to add some parts in our custom themes where we can change the html to override certain classes. But I think that is a little bit off-topic.

JustCarmen:
This is an old obsolete PR which should be closed.

PR #1454, opened 2017-11-02
Place update

magicsunday:
Instead of juggling around the gedcom place string, it would be easier/clearer to use the Place object instead. This also allows usage of the method to extract parts of a place or the short name used in charts. And it removes the required knowledge of the gedcom standard to handle place strings.


app/Census/CensusColumnFatherForeign.php
```diff
@@ -33,7 +33,7 @@ class CensusColumnFatherForeign extends AbstractCensusColumn implements CensusCo
 	public function generate(Individual $individual, Individual $head = null) {
 		$father = $this->father($individual);
 
-		if ($father && $this->lastPartOfPlace($father->getBirthPlace()) !== $this->place()) {
+		if ($father && $this->lastPartOfPlace($father->getBirthPlace()->getGedcomName()) !== $this->place()) {
```
fisharebest: Can we just use `getBirthPlace()->lastPart()` here?

If yes, the function `$AbstractCensusColumn::lastPartOfPlace()` is no longer needed?

app/Census/CensusColumnFatherBirthPlace.php
```diff
@@ -34,7 +34,7 @@ public function generate(Individual $individual, Individual $head = null) {
 		$father = $this->father($individual);
 
 		if ($father) {
-			return $this->notCountry($father->getBirthPlace());
+			return $this->notCountry($father->getBirthPlace()->getGedcomName());
```
fisharebest: We now have `Place::lastPart()`.
For consistency, perhaps it is better to have `Place::notLastPart()` and remove `AbstractCensusColumn::notCountry()`?

app/Family.php
```diff
@@ -307,7 +307,7 @@ public function getAllMarriageDates() {
 	/**
 	 * Get a list of all marriage places - for the family lists.
 	 *
-	 * @return string[]
+	 * @return array|Place[]
```
fisharebest: Just `@return Place[]` ?

app/GedcomRecord.php
```diff
@@ -1039,14 +1039,14 @@ public function getAllEventDates($event_type) {
 	 *
 	 * @param string $event_type
 	 *
-	 * @return array
+	 * @return array|Place[]
```
fisharebest: Just `@return Place[] `?

app/Individual.php
```diff
@@ -446,7 +446,7 @@ public function getBirthPlace() {
 			}
 		}
 
-		return '';
+		return new Place(null, $this->tree);
```
fisharebest: The constructor takes a string as its first argument, so perhaps: 

`return new Place('', $this->tree);a

app/Individual.php
```diff
@@ -485,7 +485,7 @@ public function getDeathPlace() {
...
```
fisharebest: `null` =>  `''`

app/Individual.php
```diff
@@ -540,7 +540,7 @@ public function getAllBirthDates() {
 	/**
 	 * Gat all the birth places - for the individual lists.
 	 *
-	 * @return string[]
+	 * @return array|Place[]
```
fisharebest: Just `@return Place[]` ?

app/Individual.php
```diff
@@ -572,7 +572,7 @@ public function getAllDeathDates() {
 	/**
 	 * Get all the death places - for the individual lists.
 	 *
-	 * @return string[]
+	 * @return array|Place[]
```
fisharebest: Just `@return Place[]` ?

app/Place.php
```diff
@@ -210,7 +222,7 @@ public function getReverseName() {
 	 *
 	 * @param Tree $tree
 	 *
-	 * @return string[]
+	 * @return array|Place[]
```
fisharebest: Just `@return Place[]` ?

app/Place.php
```diff
@@ -246,7 +258,7 @@ public static function allPlaces(Tree $tree) {
 	 * @param string  $filter
 	 * @param Tree    $tree
 	 *
-	 * @return Place[]
+	 * @return array|Place[]
```
fisharebest: Just `@return Place[]` ?

fisharebest:
I've merged this change.

Thanks

PR #1459, opened 2017-11-05
Include extended BDM facts (eg Baptism, Burial etc.)

ddrury:
I find it strange that we don't include the 'extended' BDM facts (baptism, burial etc.) in this module; so I've used the constants defined in session.php to build the complete list of facts to search for. I've also included an option that allows for the original BDM facts (default) or the extended list to be selected.


ddrury:
Withdrawn - try to refactor using views

PR #1461, opened 2017-11-05
Fixed user messages block id, added toggle to checkboxes of messages …

magicsunday:
…list


fisharebest:
Toggle is good - if all are the same.

But if you have selected one item, and then click "all", the one item is unselected.

Perhaps toggle the first one, and then make the rest the same as the first?

PR #1512, opened 2017-11-23
Update Readme.MD

ddrury:
Updated required sections covering php version and browser compatibility and standardise use of "directory" over "folder" (would prefer other way round but that would require a text change in code for consistency)


fisharebest:
> standardise use of "directory" over "folder" (would prefer other way round but that would require a text change in code for consistency)

We've had a long-standing policy to use "folder".  https://wiki.webtrees.net/en/Terminology

What's the problem here?

ddrury:
My error, I thought the text was "Clean up data directory". I'll make the change

ddrury:
Buggered up the commit - I'll resubmit

PR #1581, opened 2018-01-16
Readme.MD updated to reflect current requirements and facilities

ddrury:
Updated PHP version and required modules, possibility of using MariaDB, browser compatibility, references to control panel pages, describe automatic upgrade facility and reference ANSEL conversion utility

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,50 +76,53 @@ the project even stronger and better.
 
 To install **webtrees**, you need:
 
-* A webserver. Apache and IIS are the most common types. There are no requirements
-  to use a specific type or version.
+* A webserver. Apache, NGINX and IIS are the most common types. There are no
+ requirements to use a specific type or version.
 * Approximately 65MB of disk space for the application files, plus whatever is
   needed for your media files, GEDCOM files and database.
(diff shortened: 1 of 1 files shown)


fisharebest:
The MySQL version requirement was added before 1.0.0 was released, and is the lowest version that supports stored procedures.  Since we never used stored procedures, we can actually use older versions of MySQL.  I don't have an easy way to test, but webtrees will probably work with much older versions of MySQL.

ddrury:
Do you want me to remove the reference to a mysql version?

Something else, does webtrees still require about 65M of server space?

Regards
David

fisharebest:
On 17 January 2018 at 10:05, David Drury <notifications@github.com> wrote:
> Do you want me to remove the reference to a mysql version?

If you could.  I've removed the MySQL version check from the code.

> Something else, does webtrees still require about 65M of server space?

It's currently running at about 67MB, so yes, about that.

Greg
-- 
Greg Roach, Stony Stratford, Bucks, England, greg@subaqua.co.uk
"Don't eat animals - it's not good for them and they don't like it"

Issue #1572, opened 2018-01-06
RTL Top 10 given name block table sizes - 2.0.0


melizaa:
In the rtl Top 10 given-name blocks the male and female tables have different sizes. This does not look good on the Right section blocks, where the female and male data prints one above the other.

![2018-01-05_19-31-58](https://user-images.githubusercontent.com/5202996/34636099-2a0cbb7e-f24f-11e7-9f49-6c24eb75895c.jpg)

fisharebest:
The layout of these has been improved.  On narrow screens (and on the right-hand side), the two tables are centered vertically.  Otherwise they are centered horizontally.

PR #1591, opened 2018-01-23
Updated datatables to responsive bootstrap4 style

magicsunday:


magicsunday:
This may also fix #1572

fisharebest:
This plugin is interesting, but I am not sure we need it for this simple table?

Also, I do not think it works with assistive technology:
I cannot use it with a keyboard.  So, it will not work with voice-control.
I do not think it works with screen readers.

screen shot 2018-01-24 at 20 38 37

I guess it could be useful for the control panel, where we **need** to show lots of columns and data???

Currently, we force the site owner to choose tables (look good on desktop) or lists (look good on mobile). 

Must we have this choice?

Perhaps we can look for a simpler way to show the data.  For example, we could show the counts using [badges](https://getbootstrap.com/docs/4.0/components/badge/), with colors for male/female.

I am sure there are other options.  Do you have any thoughts or ideas?

magicsunday:
Currently the tables on the dashboard page breaks the layout if you change the resolution of the output device. This pull request was intended to correct this. I think this solution is a little better than the existing one. Whether it is optimal, you can of course discuss.

We could also omit "datatables.net-responsive-bs4" and wrap all tables with "table-responsive" classes (then they would be vertically scrollable).

magicsunday:
Please take a look at it: https://github.com/magicsunday/webtrees/tree/top-surnames-badge-list

I moved the stuff out of FunctionsPrintLists in a new template, but currently did not changed much of the logic, which should be done too, to keep the template clean.

magicsunday:
@fisharebest Got time to look at it?

magicsunday:
@fisharebest I updated the style of the surnames list with badges. What do you think?

screenshot-2018-2-9 webtrees

PR #1593, opened 2018-01-23
Added "table-responsive" class

magicsunday:
... to hide overflow and allowing scrolling vertically in table


fisharebest:
According to the [documentation](https://getbootstrap.com/docs/4.0/content/tables/#responsive-tables), the `table-responsive` class needs to be on the *parent* of the `<table>`.

I tried adding it there, but could not see a difference...

magicsunday:
You are right. It should be on the parent. I moved it.

PR #1594, opened 2018-01-23
Modified On this Day module

ddrury:
Include extended BDM facts (eg Baptism, Burial etc.). The output is generated with views.

BTW the function FunctionsPrintLists::eventsList has a comment on line 1603 that says sorting by anniversary date is unnecessary as the data is already in that order. This doesn't seem to be the case - see the attached image (fixed in this pr)

otd


ddrury:
Cancelled - a better solution is forthcoming

PR #1626, opened 2018-02-21
Fixes #1622

### LINKED ISSUE #1622: ErrorController::noRouteFound
After login as admin the ErrorController::noRouteFound is triggered and redirects (line 57) me to "admin_trees_manage.php" without any error message.

magicsunday:
@fisharebest Please take a look at the following changes. This fixes for me the admin login to a private tree.

If no tree is available at all and the user is no admin he gets the mentioned message "This user account does not have access to any tree" displayed. If admin he get redirected to "admin_trees_manage.php".

### DIFF
--- a/app/Http/Controllers/ErrorController.php
+++ b/app/Http/Controllers/ErrorController.php
@@ -57,7 +57,7 @@ class ErrorController extends BaseController {
 			return new RedirectResponse(Html::url('admin_trees_manage.php', []));
 		}
 
-		return $this->viewResponse('errors/no-tree-access', []);
+		return $this->viewResponse('errors/no-tree-access', [ 'title' => '' ]);
 	}
 
 	/**
--- a/app/Session.php
+++ b/app/Session.php
@@ -53,11 +53,7 @@ class Session {
 	 * @return mixed
 	 */
 	public static function get($name, $default = null) {
-		if (isset($_SESSION[$name])) {
-			return $_SESSION[$name];
-		} else {
-			return $default;
-		}
+		return $_SESSION[$name] ?? $default;
 	}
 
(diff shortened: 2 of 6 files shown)


fisharebest:
The change to `session.php` fixes the problem for me.

I have merged these changes.

The changes to `index.php` (`// Force login`) seem to duplicate the logic in `ErrorController::noRouteFound()`.  Am I missing something?

fisharebest:
OK - I see what I am missing. Redirect to admin page, not home page.

fisharebest:
> Redirect to admin page, not home page.

I am sure I saw this happen.  But I cannot reproduce it.

magicsunday:
The problem is:

    Session::get('GEDCOM', $first_tree_name);

won't return the first_tree_name, as the session array contains an empty "GEDCOM" entry. Session::get checks only if the value isset. I don't know if this could safely changed to "empty" or if something else will break.

After changing: 

    $from_session = Session::get('GEDCOM', $first_tree_name);
	
to:

    $from_session = Session::get('GEDCOM');
    $previous_tree_name = !empty($from_session) ? $from_session : $first_tree_name;

it seems fine.

fisharebest:
> as the session array contains an empty "GEDCOM" entry

Aha!

In `session.php`, we only write non-empty values to `$_SESSION`.
In `index.php`, we always write it - even if it is empty.

So, the actual bug is that we write an empty value.

magicsunday:
You are right. 

    if ($tree_name) {
        Session::put('GEDCOM', $tree_name);
    }

seems to work too.

fisharebest:
Merged.  Thanks.

PR #1639, opened 2018-03-09
Openstreetmaps module first release

ddrury:
~~**NOT PRODUCTION READY AS IT STANDS**~~

Created with the assumption that it's a core module, ~~however it has a couple of local classes not currently loaded with composer~~. The Schema 3 update has been commented out for now. The Individual tab, Pedigree map and Place lists use routes & views, the admin pages still have embedded HTML

Uses leaflet.js and a few associated plugins. It can use all the providers listed [here](http://leaflet-extras.github.io/leaflet-providers/preview/index.html) (a subset that I find suitable can be found in the file modules_v3/openstreetmap/providers.xml)

Tested on all current desktop browsers & minimal testing on Android (seems OK) - I don't have access to any IOS devices.

mimics all functions of Googlemaps module and in addition provides geoJSON import/export. 

~~**Note1** the map code for the place hierarchy is independent of placelist.php - the map is just inserted in the appropriate place.~~
**Note2** the csv export function now actually produces a comma separated file (Googlemaps actually uses a semicolon as the separator). The csv import function can handle commas & semicolons. 
**Note3** ~~the placelocation zoom field is not used in any of the user pages. On the admin add/edit page it is shown and saved should it be altered. The place location can be accurately set by using the map zoom buttons and dragging the marker to the appropriate spot, therefore the zoom field could quite easily be removed, however we may need to set sensible defaults somehow - eg. zoom for USA and Malta need to be quite different but are both countries (i.e level 0)~~ Had to revert to using zoom when there is only one marker to display - I can't find a solution that matches every instance - open to suggestions
**Note4**. The leaflet markercluster add on has a function zoomToShowLayer which I use with the user maps. When a sidebar item is clicked the popup attached to the associated marker is opened. This works fine on the Place hierarchy but for some reason on the Pedigree map & Individual tab the map is zoomed to it's maximum value - ~~I can't figure out why (maybe to do with the relative proximity of all the markers?)~~ This occurs when all the markers in a cluster have exactly the same coordinates, therefore the bounds of that cluster equate to a point and the function zooms in repeatedly until those bounds exceed the maps displayed area (which they never will) stopping only when the maps maxZoom level is reached - so ... working as designed


fisharebest:
Hi Dave - I'm in the middle of a push to convert the remaining code to use the http/controller/view structure.  If I don't look at this right away, it's because I'm busy elsewhere.

ddrury:
Understood

ddrury:
I'm closing this as I've split it into parts, hopefully to make it easier to digest

PR #1644, opened 2018-03-13
Fixed initial sort order of top given/surname tables

magicsunday:


fisharebest:
One of these tables has already been moved to a template (`resources/views/tables/surname`).

I copied moved the change to the new code.

PR #1647, opened 2018-03-13
Updated log/sign in forms to responsive format

magicsunday:
@fisharebest 

Please take a look at the updated login form. Its based on the common bootstrap login forms. Labels only displayed to screen readers otherwise hidden.


fisharebest:
screen shot 2018-03-26 at 14 43 37

I don't think it is a good idea to remove labels and rely on placeholders.  There are lots of online resources to back this up.  e.g.  https://developer.paciellogroup.com/blog/2011/02/html5-accessibility-chops-the-placeholder-attribute

It would fail accessibility testing at my employer.

magicsunday:
Labels are only hidden by the bootstrap class "sr-only".

PR #1679, opened 2018-04-08
OpenStreetMap module part1

ddrury:
mimics Googlemaps


ddrury:
Parts 1, 2 and 3 are demonstrations of functions. Part 4 brings it all together and contains a few optimisations not present in the previous parts

PR #1680, opened 2018-04-08
Openstreetmap module part2

ddrury:
convert Pedigree map to route/view


ddrury:
Rethinking strategy - will resubmit soon

PR #1681, opened 2018-04-08
Openstreetmap module part3

ddrury:
Convert Place hierarchy to route/view


ddrury:
Rethinking strategy - will resubmit soon

Amgine0:
<lurks attentively>

PR #1722, opened 2018-05-11
also checking for a $tree that is false

mwesten:
If no trees are available yes (after a fresh install) the $tree variable is "false"(bool). A check for a null value was done, but not for a "false", thus resulting in errors.


fisharebest:
If no tree is defined (all are private), then `$tree` *should* be null.

The culprit is `current()` in this line:

https://github.com/fisharebest/webtrees/blob/master/index.php#L52

fisharebest:
I've fixed the problem which caused `$tree` to be set to `false` instead of `null`.

PR #1724, opened 2018-05-13
Osm more typehinting

ddrury:


ddrury:
withdrawn to fix conflicts

PR #1755, opened 2018-06-08
fix user admin page link in the user verify notification letter

naudhr:
In the letter a moderator gets to review new user's profile there is a typo in the url so they can't get to the page unless manually fix `admin_users-edit` to `admin-users-edit`.


fisharebest:
Thanks!

PR #1762, opened 2018-06-17
Tree error check message fix

naudhr:
If there are no errors have been found during the tree check, a creepy message «', I18N::translate('No errors have been found.'), '» appears.


fisharebest:
Thanks

PR #1763, opened 2018-06-17
A user registration fix

naudhr:
While composing an e-mail, there occurs an exception:

Uncaught Error: Call to undefined method Fisharebest\Webtrees\Tree::getTitleHtml() in resources/views/emails/register-notify-text.php:4

So a user gets their e-mail, but the administrator doesn't. So does not `User::create` routinge get called, user e-mail verification fails; as a tip of the cake, the user gets «Hello administrator...»·on the screen.

### DIFF
--- a/resources/views/emails/register-notify-html.php
+++ b/resources/views/emails/register-notify-html.php
@@ -4,7 +4,7 @@
 </p>
 
 <p>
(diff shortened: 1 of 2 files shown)


fisharebest:
This variable still needs to be escaped, so we need to replace `$tree->getTitleHtml()` with `e($tree->getTitle())`.

I removed the function `getTitleHtml()` (and similar ones) because escaping output is a function of the views, not a function of the objects.

Can you update this.  Thanks.

PR #1777, opened 2018-07-09
Incorrect famid used in getPrimaryChildFamily

jon48:
The logic in `getPrimaryChildFamily` wrongly assumes that the key of the returned array from `getChildFamilies` is the Family Xref (whereas it is just the numercal index). As a consequence, `$famid` does not contain the correct value, and the checks are flawed (and will certainly fall back to the first family).

Looking at the code history, the bug seems quite old, but I guess that it has been a case rare enough for an individual to have multiple FAMC records, and the first not being the primary one, that it has not been reported before.


fisharebest:
Thanks.

PR #1843, opened 2018-08-22
Use pclzip 2.8.2.1, to fix usage with PHP 7.1 and 7.2

jcnventura:
Use an improved version of pclzip that is compatible with PHP 7.1+.


mpwt:
For work with the new lutana/pclzip you have to change the path in vendor/composer/autoload_classmap.php and autoload_static.php too. Thanks.

fisharebest:
@mpwt - thanks

PR #1848, opened 2018-08-24
Update functions.php

magicsunday:
Added an double underscore function to the helper methods in functions.php to simpler utilize the translation method. This seems very common in various other libs and saves some typing.

### DIFF
--- a/app/Helpers/functions.php
+++ b/app/Helpers/functions.php
@@ -16,6 +16,7 @@
 declare(strict_types=1);
 
 use Fisharebest\Webtrees\Html;
+use Fisharebest\Webtrees\I18N;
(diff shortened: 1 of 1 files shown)


magicsunday:
Total occurrences:

translate = 5356
translateContext = 1092
number = 318
plural = 61
digits = 27
percentage = 9

In my opinion it's only necessary to cover "translate" and "translate Context". The others are negligible. The shorthand versions do not have to exist for all methods.

fisharebest:
The plan is to migrate to Laravel.  Laravel uses the `__()` function - but it handles placeholders differently.

laravel: `__('Hello :name', 'fish')`

webtrees: `translate('Hello %s', 'fish')`

PR #1852, opened 2018-08-27
Update composer.json

magicsunday:
Added missing dependencies.


magicsunday:
ext-curl is required by Select2 and AutocompleteController
ext-intl is required by I18N (Class Collator)
ext-zip is required by AdminUpgradeController

PR #1854, opened 2018-08-27
Use fully qualified global function calls

magicsunday:
Use fully qualified global function calls to improve performance:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/app/Bootstrap4.php
+++ b/app/Bootstrap4.php
@@ -34,7 +34,7 @@ class Bootstrap4 extends Html
         if (empty($items)) {
             return '';
         } else {
-            return '<span class="badge badge-secondary">' . I18N::number(count($items)) . '</span>';
+            return '<span class="badge badge-secondary">' . I18N::number(\count($items)) . '</span>';
         }
     }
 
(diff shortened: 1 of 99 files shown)


fisharebest:
I agree that it is good to be explicit about global/namespaced functions.

But we cannot do this for *all* functions, as it will prevent us from creating mock-functions in tests.  For example, see `ini_get()` and `microtime()` in [this test](https://github.com/fisharebest/webtrees/blob/master/tests/app/Services/TimeoutServiceTest.php).

There are two ways to specify global functions.  We can import them, or use a fully-qualified name.

```
use function strpos;

strpos(...);
```

```
\strpos(...);
```

We import classes.  I think it is clearer and more consistent to import functions.  What do you think?

magicsunday:
The functions "ini_get" and "microtime" are not extended to the fully qualified named version. There are only some optimized functions which takes advantage of some special optimizations.

https://github.com/php/php-src/blob/f2db305fa4e9bd7d04d567822687ec714aedcdb5/Zend/zend_compile.c#L3872

But according to https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/ this seems to apply to all internal functions.

Personally I dont like all the slashes in the source code. I thinks its may be easier to read the source code im we use the ``use`` statement.

PR #1856, opened 2018-08-27
Fixed PHPdoc blocks

magicsunday:


fisharebest:
Some of these changes look OK.

You added `@package` tags to some of the classes.  I don't think these are needed.

You added `@throws \Exception` tags to some of the functions.  I don't see that these functions can throw exceptions.

magicsunday:
- I removed the @package

- The `@throws \Exception` bubbles up from a previously called function and indicates the exception is not processed, e.g. StatisticsChartController::chartCustomChart
-- calls `checkModuleIsActive`
-- this calls `Module::getActiveCharts` and then `getActiveModulesByComponent`
-- the Database::execute throws an possible \Exception

I think, even if the exception is caught at the end, the docblock should be complete, indicating that an exception can be thrown in a method.

magicsunday:
I splitted the commits in one fixing missing dockblocks etc and one fixing the missing throw tags.

fisharebest:
> I splitted the commits in one fixing missing dockblocks etc and one fixing the missing throw tags.

Thanks.  I have merged the DocBlock commit.

IMHO, we should only use `@throws` for exceptions that are thrown in the function (not in functions that are called indirectly.)

magicsunday:
Ok

PR #1857, opened 2018-08-27
Update File.php

magicsunday:
Check result of mkdir and throw an exception im directory creation fails.

### DIFF
--- a/app/File.php
+++ b/app/File.php
@@ -75,7 +75,14 @@ class File
                 self::mkdir(dirname($path));
             }
             try {
-                mkdir($path);
+                if (!mkdir($path) && !is_dir($path)) {
+                    throw new \RuntimeException(
+                        sprintf(
+                            'Directory "%s" was not created',
+                            $path
+                        )
+                    );
+                }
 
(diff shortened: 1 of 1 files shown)


fisharebest:
I can't check this right now, but I think that if `mkdir()` fails, then it will raise an `E_WARNING`.

We set `error_reporting()` to `E_ALL`, so this will already raise an exception?

magicsunday:
This fix was adopted from https://github.com/kalessil/phpinspectionsea/blob/master/docs/probable-bugs.md#mkdir-race-condition

magicsunday:
I thinks its better to check the result of the function. Before the functions returned true even if mkdir failed.

fisharebest:
The `File` class is being removed.  Most of it has gone.

It will be replaced with `Flysystem`.

This will give two advantages.

1) testability - we will be able to use a mock-class to test file operations.
2) we can use either the "local" adpater or the "SFTP" adapter, to remove the need to `chmod 777 /data/`
3) we can store our data files anywhere - not just on our webserver.

PR #1858, opened 2018-08-27
Update AdminTreesController.php

magicsunday:
Renamed internal variable "tree" as this is already passed as method parameter.

### DIFF
--- a/app/Http/Controllers/AdminTreesController.php
+++ b/app/Http/Controllers/AdminTreesController.php
@@ -1606,10 +1606,17 @@ class AdminTreesController extends AbstractBaseController
             }
         }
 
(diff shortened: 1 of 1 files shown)


fisharebest:
Just remove the parameter?

The parameters for the http-controller actions are created automatically by the  `Resolver` class.  No other code changes would be needed.

magicsunday:
The parameter `$tree` is used in line 1602ff and 1624.

fisharebest:
The `$tree` parameter is used to redirect back to the same tree after the synchronisation is complete.

Moving the line which creates the URL is a simpler solution.  38d96dad3

PR #1861, opened 2018-08-27
Update LoggedInUsersModule.php

magicsunday:
Removed always true


fisharebest:
Thanks.  This was added for testing.

PR #1863, opened 2018-08-27
PHP fixes

magicsunday:
Various PHP fixes

### DIFF
--- a/app/Date/CalendarDate.php
+++ b/app/Date/CalendarDate.php
@@ -97,7 +97,7 @@ class CalendarDate
     public function __construct($date)
     {
         // Construct from an integer (a julian day number)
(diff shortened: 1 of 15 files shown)


magicsunday:
- getPreference should not return strict "string" if the default value may differ from it.
- either allow mixed or force default value to be string too

fisharebest:
I have merged 6 of the 8 commits.

The preferences are always strings - they are stored as strings in the database.  The default should have the same type as the data.

For the functions in loops, we *need* the `microtime()` function in the loop, as it returns a different value each time.  In any case, this logic is now moved to a new class (`TimeoutService`)

magicsunday:
Uhh, I have not seen this microtime stuff.

PR #1864, opened 2018-08-27
Added PHP7 return types to methods according doc block

magicsunday:


fisharebest:
> according doc block 

I hope the docblock is correct ;-)

fisharebest:
Oops.  Lots of things broken.  Application won't start.

magicsunday:
Uh :(

magicsunday:
The getPreference method is fixed in this commit: https://github.com/fisharebest/webtrees/pull/1863/commits/0cee83a35d1ab5e6973971301a2dcb95b8560c33

fisharebest:
The preferences are stored in the database as strings.
They are retrieved as strings.
So, I think it is clearer if we always save them as strings.

PR #1870, opened 2018-08-29
Moved HTML out of class and into template files

magicsunday:
Please have a look at this. I think the HTML should not reside in the classes. So I moved it into template files (Not all, because it's too much for the moment. 🙄). Maybe the template files should names .phtml instead of .php to better clearify the intention?


magicsunday:
Please have a look at this. I think the HTML should not reside in the classes. So I moved it into template files. Maybe the template files should names .phtml instead of .php to better clearify the intention?

PR #1871, opened 2018-08-29
Static methods called in a non static way

magicsunday:
Some static methods were called using `$this->`


fisharebest:
All these functions did not need to be `static` - so the fix is to update the functions, not the calls.

magicsunday:
Ok.

PR #1904, opened 2018-09-14
Strict types

### LINKED ISSUE #1900: Use declare(strict_types=1) in all PHP files?
Just a question. Should we use ``declare(strict_types=1)`` declaration to force strict type checking?

magicsunday:
Fixes #1900 

I added `declare(strict_types = 1);` to all classes except `resources/*` (this should be done in a separate task, maybe even rename the files to `.phtml`). I also fixed the errors which encountered me while I was testing the strict type declaration (hopefully I got them all).


fisharebest:
Code style.

PSR-12 isn't issued yet, but I am trying to follow it.  See https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#3-declare-statements-namespace-and-import-statements

The existing `declare` statements already follow this standard.

I see the problems you have found in the two commits labelled "fix strict errors".  For some of these, I think there are perhaps better solutions.  For example, in `printSosaNumbera()`, the code is using the same parameter for two things.  It may be better to have two parameters, etc.

I will look at these first.

magicsunday:
I moved the `declare` statements after the file level docblock and also removed the whitespaces inside them.

fisharebest:
Thanks.  I'll merge this into a local branch, and use it for testing/development.

I'll try to find/fix as many issues as possible before I merge the PR.

fisharebest:
These have been merged in.

PR #2047, opened 2018-12-02
Fix reselecting batch update plugin.

Bonbadil:
Hello,

When you change option of a batch update plugin, the page reload but the plugin isn't correctly reselected because of a \ in excess. And if you reselect plugin, you loose the chosen option. Making plugins' option unusable.


fisharebest:
Thanks.

PR #2060, opened 2018-12-13
Update Stats.php

magicsunday:
Fixes missing quotes in query, resulting in empty response. 

https://dev.webtrees.net/demo-dev/index.php?route=statistics&ged=demo#tab-other-statistics

=> First event, last event

### DIFF
--- a/app/Stats.php
+++ b/app/Stats.php
@@ -3023,7 +3023,7 @@ class Stats
             'CENS' => I18N::translate('census added'),
         ];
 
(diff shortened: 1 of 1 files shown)


fisharebest:
I know know MySQL *can* use `"` for strings, but `'` is the SQL standard, and we use single quotes everywhere else.

Can you switch from `"` to `'`.

fisharebest:
Just realised I can edit commits in the PR myself

PR #2086, opened 2019-01-03
Admin Users wrong name shown

ddrury:
Wrong name shown in full name column


fisharebest:
@ddrury - you may have noticed that I have been rewriting the database access code to use `illuminate/database`.

(This is a database-agnostic interface, which lets us use databases other than MySQL.  In particular, it lets us use in-memory SQLite databases - which means we can now run automated tests on code that needs to access the database.)

I have added a test for this feature "user and real names both shown on the admin page".  This should prevent any future code changes from breaking this feature.  A very small step - but one in the right direction!

ddrury:
A lot of work - whats that chinese saying - "A journey of a thousand miles begins with a single step"

PR #2156, opened 2019-01-30
2059

magicsunday:
I created a pull request due the massive changes in the last days. Please take a look at the changes. Mainly I moved all the stuff from Stats.php to separate repository classes. I also started changing the database queries to use illuminate db and not plain SQL anymore. Some statistics now have templates to render its stuff. Its not ready yet. Things to do:
- move every piece of HTML to a template
- replace outdated google charts with chart.js or the new google chart implementation
- family and individual repository needs additional work

Maybe not all of the statistics are needed anymore? Especially the comma (white space) separated ones could be removed, I think?


fisharebest:
Thanks for this!  I will try to review and merge it later today.

I will also move and rename some of the classes.

I am moving business logic such as this to the `/app/Services/` folder.

Eventually, the `/app/` folder will just contain classes that represent entities such as individuals, dates, places, etc.

PR #2193, opened 2019-02-08
Update View.php

magicsunday:
Use a hash to prevent adding the same content to the stack more than once. 

May there any reason against this? Everything seems to work well.

If using "#getAllTagsTable#" and the methods called in the block are adding javascript, the code is added multiple times to the footer.


magicsunday:
This is already included through the chart update.

PR #2204, opened 2019-02-12
Update AbstractGoogle.php

magicsunday:
Fix division by zero.

The greek tree throws errors on the statistics page.
https://dev.webtrees.net/demo-dev/index.php?route=module&module=statistics_chart&action=Chart&ged=mythology


fisharebest:
FYI, the unit tests include a test for these - it runs all the statistics with an empty tree.

However, we can't run this test until we finish converting all the SQL to use the new DB adapter...

magicsunday:
Ah ok. Good to know.

Issue #2120, opened 2019-01-19
Module system in 2.x


ric2016:
Following the suggestion in #1950, this ticket collects some high-level aspects to consider for the module system in webtrees 2.x.

* Installation

Are modules to be installed in the same way as before, i.e. simply by dropping a directory in the proper location, or is something else preferable? The current approach is probably fine.

* Themes vs modules
 
Is it a good idea to merge these? The functionality seems rather different usually, what are the benefits of having these two concepts integrated? On the other hand, a module provides all kinds of different functionality already (tab, block, chart ...), so a theme can be seen as just one more of these.

* Protection against errors

Should we attempt to catch exceptions and fatal errors (which is apparently possible in php 7) from modules in order to present a page that is not completely broken in these cases? If this is easily possible, it seems like a good idea, but ultimately the site admin has to handle these cases anyway. Note that we probably cannot keep a module from displaying errors or stack traces within its own output area (such as a tab) anyway.

* Separation of modules

Is it important to enforce a unique internal module name? Should activated duplicated modules work independently? I cannot think of a use case where it makes sense to have the same module more than once, other than for testing purposes. Still, it seems like a good idea to use unique names. The easiest way to achieve this is probably to manage them outside of the module (e.g. use the directory name and pass it to the module as a variable), rather than to have the module implement a `getName` method (in which case we'd have to reject modules with non-unique names).

* Naming

A module should be able to name all of its parts independently. Currently the name of a module's tab cannot be different from the module's (displayed) name itself, this restriction seems unnecessary. It may also make sense to make all names configurable in general (tab name, chart name etc), so that a system with two independent modules which happen to use the same name e.g. for a tab can disambiguate these.

* Hooks

Should we additionally provide some kind of hook system, where modules can provide smaller UI parts or other functionality - E.g. not a complete tab or chart, but some additional output in a facts & events block? I'm currently using this approach to integrate several of my own custom modules. As long as this is just about additional content to display, it can probably be handled already by extending the respective view (once views are used everywhere, and sufficiently broken down).

* Lists

This is the only remaining main hierarchy where it's currently not possible to add entries via custom modules. We should allow this, similar to charts, reports etc, and consequently turn the existing lists into modules.

To be continued - We may have to split this up into multiple tickets though.

fisharebest:
> Are modules to be installed in the same way as before, i.e. simply by dropping a directory in the proper location, or is something else preferable? The current approach is probably fine.

The current installation process has two steps. 

[...]

PR #2230, opened 2019-02-17
Lists as Modules

ric2016:
As suggested in Issue #2120.

Some implementation notes:

For now, I have kept and adapted existing classes and views (e.g. ListController, views), this could be cleaned up later.

Setting the access levels on the new list modules affects (as usual) other modules using the lists, see e.g. TopSurnamesModule (if IndividualListModule isn't accessible, top surnames are still displayed, but without links to the lists). For the list access level settings, I had to migrate the respective database table (Migration42.php).

Note that in this initial implementation, several UI elements still create lists directly, e.g. source-page.phtml via families-table.phtml. I'm not sure if these should be routed via the new modules as well.


fisharebest:
> For the list access level settings

Some types of module have access levels - block, chart, menu, report, sidebar, tab
Others are simply enabled/disabled.

I wasn't expecting lists to have access levels.  My expectation was that this feature would simply allow users to

* disable unused lists (e.g. "shared notes")
* provide new types of list - perhaps lists of certain facts or events 
* provide replacement versions of existing lists

Since the contents of the list are already filtered for privacy, I'm not sure there is a need to set an access level for them.

I will be very busy next week, and so it may take me some time to review this.

> Note that in this initial implementation, several UI elements still create lists directly, e.g. source-page.phtml via families-table.phtml. I'm not sure if these should be routed via the new modules as well.

That's an interesting question.  More thought is needed.

ric2016:
> I wasn't expecting lists to have access levels.

I guess it's debatable whether this feature is actually very useful. In any case, it doesn't affect most aspects of the implementation (e.g. even without access levels, TopSurnamesModule still has to check whether IndividualListModule is enabled, and so on). Since most other 'main' components have access levels, it probably doesn't hurt to keep them here as well for consistency.

> Since the contents of the list are already filtered for privacy, I'm not sure there is a need to set an access level for them.

Similar arguments could be made e.g. for charts - why do we allow access levels there?

(As long as they don't introduce too much additional complexity (both in develoment and in administration), I'm generally always in favor of additional configuration options.)

magicsunday:
Please do not use closures all over the place.

ric2016:
> Please do not use closures all over the place.

I used them in this case to preserve the existing function signatures and implementations as much as possible.

I have to ask though: Is this general webtrees development policy, and if so, is there some kind of guide for this? It's sometimes hard to do the right thing here ...

magicsunday:
This is just my personal preference as closures passed to public methods make code more difficult to read and unterstand.

fisharebest:
> Similar arguments could be made e.g. for charts - why do we allow access levels there?

Perhaps we should remove access levels from charts?

Perhaps the requirement for access levels should depend on the specific module, not on the type of module.  `ModuleHasAccessLevelInterface`?

I can see some modules that are only useful for certain users.  For example:

* the "Missing data report" is only useful to users with "Edit" role.
* the "Pending changes report" is only useful to users with "Moderate" role.

In both these cases, the "access level" does not allow us to select "Editors" or "Moderators" - the acces level is based on reading the data, not writing the data.

Perhaps 

> Please do not use closures all over the place.

I have not had time to look at the code in detail, but a couple of comments.

`$router` can return `null` - but we pass the return value to `e()` - which can only take a `string`.

It is not obvious why we use `$router` instead of calling `ModuleListInterface::listUrl()` directly?

But I have only had a *very* quick look at the code.  I hope to have more time at the weekend.

ric2016:
Regarding the access levels: In my view, they are useful regardless of the character of the module and its data. If the site admin e.g. wants to hide specific charts, lists etc from visitors, for whatever reason (in order to entice them to sign up maybe?) - why shouldn't he be able to configure this?

ric2016:
> `$router` can return `null` - but we pass the return value to `e()` - which can only take a `string`.

I missed a few null checks - I'll amend the pull request.

> It is not obvious why we use `$router` instead of calling `ModuleListInterface::listUrl()` directly?

I thought it would be cleaner this way (abstracting the url generation out of the views) - the views shouldn't contain any (complex) logic with regard to how to generate urls in my opinion, and this was intended as a step in that direction; while at the same time preserving the existing function signatures and implementations as much as possible. (It also makes the views more flexible, as they could be used outside a ModuleListInterface context)

fisharebest:
> I thought it would be cleaner this way (abstracting the url generation out of the views)

I think you can still do this by passing the module instead of `$router`.

```
$module = $module_service->findByInterface('IndividualListModule::class)->first();
echo printTagCloud($tree, $module, ....);

...

function printTagCloud(Tree $tree, ?ModuleListInterface $module, ....) {
    if ($module instanceof ModuleListInterface) {
       echo '<a href="' . module->listUrl($tree, ['surname' => $surrname]) ."' . $surname . '</a>'
    } else {
       echo '<span>' . $surname . '</span>';
    }
}
```

ric2016:
> I think you can still do this by passing the module instead of `$router`.

Sure, that also works. I still think this isn't optimally abstract, but it's not that important overall. I'll make this change everywhere `$router` is used now.

ric2016:
I have updated the PR as suggested.

fisharebest:
Thanks for this.  I have merged it in.

FYI, I decided to replace the `component enum (...)` with `interface varchar()` in the `wt_module_privacy` table.

It's a little simpler, and I think a little more flexible.

PR #2235, opened 2019-02-19
Update list charts to use google chart api

magicsunday:
This is the last part of the chart update. It replaces the image charts below the individual/family list with the svg based one.


fisharebest:
Thanks!

PR #2241, opened 2019-02-20
Fixes issue 2240, failing to render custom module templates

magicsunday:


magicsunday:
Adminpanel fails now :-|

PR #2242, opened 2019-02-20
Fixes issue 2240, failing to render custom module templates

magicsunday:
Fixes issue 2240, failing to render custom module templates #2241

### DIFF
--- a/app/View.php
+++ b/app/View.php
@@ -18,7 +18,10 @@ declare(strict_types=1);
 namespace Fisharebest\Webtrees;
 
 use Exception;
+use Fisharebest\Webtrees\Module\ModuleCustomInterface;
+use Fisharebest\Webtrees\Module\ModuleInterface;
 use Fisharebest\Webtrees\Module\ModuleThemeInterface;
+use Fisharebest\Webtrees\Services\ModuleService;
 use function ob_end_clean;
 use Throwable;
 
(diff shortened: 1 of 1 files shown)


fisharebest:
I have some ideas for this in a local branch, which use namespaces.

Modules would have namespace for their own views.  e.g. `mymodule::path/to/view`
The core code would use a "global namespace".  e.g. `::path/to/view`.

This would prevent conflicts - if two modules both created a view with the same name, then one would see the view of the other.

This gets registered with something like `$module->registerViewPath()`

Modules would be able to replace views in the core code (and in other modules!) by specifying replacements.  e.g.

```
public function replaceViews() {
    return [
        '::some/core/module' => 'mymodule::replacement/for/the/view';
    ];
}
```

Using functions instead of simple search paths also allows module to choose views at run-time.

For example, they could supply different views for different themes or admin/user/visitor.

The namespace idea comes from https://laravel.com/docs/5.7/packages#views

magicsunday:
Sounds good. Please add. 😄

fisharebest:
The issue I had not yet resolved was the access levels.

The access level is part of the core code - not part of the module.

So, a module does not know if it can/will be used - and therefore does not know if it should replace existing views.

I left this work "on hold" so I could think about the way that modules interact with access levels...

fisharebest:
This functionality should now be provided by https://github.com/fisharebest/webtrees/commit/3e4bf26f9f16d92152484c56e3629888fb6019d5

magicsunday:
If I register a namespace in a module, shouldn't this be enough, that every subsequent call to view() uses this module specific namespace first, so I'm not require to always include the namespace?

    $this->viewResponse(
        $this->name() . '::template-name',
        ....
    );

PR #2247, opened 2019-02-21
Varius fixes to the statistics

magicsunday:
Due the recent changes to use the illuminate db component, some statistics have been broken.

I changed calculating the century from:

    ROUND((d_year - 50) / 100) AS century

back to:

    FLOOR((d_year / 100) + 1) AS century

as the rounding was faulty and the 21st century was missing.

FLOOR seems to be a valid standard SQL method according to the following page:
https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators?hl=de#floor

### DIFF
--- a/app/Statistics/Repository/IndividualRepository.php
+++ b/app/Statistics/Repository/IndividualRepository.php
@@ -97,8 +97,8 @@ class IndividualRepository implements IndividualRepositoryInterface
         }
 
         $rows = $query
-            ->groupBy(['n_id', 'n_givn'])
-            ->select(['n_givn', DB::raw('COUNT(*) AS count')])
+            ->groupBy(['n_givn'])
+            ->select(['n_givn', DB::raw('COUNT(distinct n_id) AS count')])
             ->pluck('count', 'n_givn');
 
         $nameList = [];
@@ -109,9 +109,9 @@ class IndividualRepository implements IndividualRepositoryInterface
                 // Exclude initials and particles.
                 if (!preg_match('/^([A-Z]|[a-z]{1,3})$/', $given)) {
                     if (\array_key_exists($given, $nameList)) {
-                        $nameList[$given] += (int) $count;
+                        $nameList[$given] += $count;
                     } else {
-                        $nameList[$given] = (int) $count;
+                        $nameList[$given] = $count;
                     }
                 }
             }
@@ -483,6 +483,7 @@ class IndividualRepository implements IndividualRepositoryInterface
             ->whereNotIn('n_surn', ['', '@N.N.'])
             ->select('n_surn')
             ->groupBy('n_surn')
+            ->orderByRaw('count(n_surn) desc')
             ->take($number_of_surnames)
             ->get()
             ->pluck('n_surn')
@@ -642,6 +643,7 @@ class IndividualRepository implements IndividualRepositoryInterface
     public function statsBirthQuery(bool $sex = false, int $year1 = -1, int $year2 = -1): array
     {
         $query = DB::table('dates')
+            ->select(['d_month', DB::raw('COUNT(*) AS total')])
             ->where('d_file', '=', $this->tree->id())
             ->where('d_fact', '=', 'BIRT')
             ->whereIn('d_type', ['@#DGREGORIAN@', '@#DJULIAN@'])
@@ -653,15 +655,13 @@ class IndividualRepository implements IndividualRepositoryInterface
 
         if ($sex) {
             $query
+                ->select(['d_month', 'i_sex', DB::raw('COUNT(*) AS total')])
                 ->join('individuals', function (JoinClause $join): void {
                     $join
                         ->on('i_id', '=', 'd_gid')
                         ->on('i_file', '=', 'd_file');
                 })
-                ->groupBy('i_sex')
-                ->select(['d_month', 'i_sex', DB::raw('COUNT(*) AS total')]);
-        } else {
-            $query->select(['d_month', DB::raw('COUNT(*) AS total')]);
+                ->groupBy('i_sex');
         }
 
         return $query->get()->all();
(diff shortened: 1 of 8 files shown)


magicsunday:
Otherwise

   (((d_year - (d_year % 100)) / 100) + 1) AS century

seems to work too. On my system I don't get any round problems while dividing 100.

fisharebest:
The change to illuminate/database was to enable support for MySQL, SQLite, Postgres and SQLServer.

`FLOOR` does not work in SQLite - which webtrees uses in its unit testing.

(This is why `FLOOR` was changed to `ROUND`)

> as the rounding was faulty

Faulty how?  Isn't this right?

```
mysql> select round((1900 - 50) / 100), round((1999 - 50) / 100);
+--------------------------+--------------------------+
| round((1900 - 50) / 100) | round((1999 - 50) / 100) |
+--------------------------+--------------------------+
|                       19 |                       19 |
+--------------------------+--------------------------+
```

magicsunday:
Changed it to (((d_year - (d_year % 100)) / 100) + 1) AS century

fisharebest:
`_year % 100` won't work with SQLite either...

magicsunday:
1901-2000 is 20th century, while 1900 is 19th, 2018 is 21st century

magicsunday:
> 
> 
> `d_year % 100` won't work with SQLite either...

damn :)

fisharebest:
I will need to check, but I thought that the statistics functions always returned the century starting from `0`, and the charts/etc. were responsible for displaying a century starting from `1`.

But if that has changed with the rewrite to statistics, then just change `- 50` to `+ 50` ?

magicsunday:
OMG. This was too obvious to see. 🤦‍♂️

magicsunday:
Change 50 to 49

select round((1899 + 49) / 100),
       round((1900 + 49) / 100),
       round((1901 + 49) / 100),
       round((1949 + 49) / 100),
       round((1950 + 49) / 100),
       round((1999 + 49) / 100),
       round((2000 + 49) / 100),
       round((2001 + 49) / 100);

fisharebest:
`+ 49` ??  Not `+ 50` ??

1800 and 1899 are in the 19th century.

```
mysql> select round((1800 + 49) / 100), round((1899 + 49) / 100);
+--------------------------+--------------------------+
| round((1800 + 49) / 100) | round((1899 + 49) / 100) |
+--------------------------+--------------------------+
|                       18 |                       19 |
+--------------------------+--------------------------+

mysql> select round((1800 + 50) / 100), round((1899 + 50) / 100);
+--------------------------+--------------------------+
| round((1800 + 50) / 100) | round((1899 + 50) / 100) |
+--------------------------+--------------------------+
|                       19 |                       19 |
+--------------------------+--------------------------+
```

fisharebest:
Did the 21st century start in 2000 or 2001 ?

fisharebest:
https://en.wikipedia.org/wiki/21st_century

magicsunday:
1800 is the 17th century from 1701 to 1800
https://en.wikipedia.org/wiki/18th_century

app/Statistics/Repository/IndividualRepository.php
```diff
@@ -97,8 +97,8 @@ private function commonGivenQuery(string $sex, string $type, bool $show_tot, int
         }
 
         $rows = $query
-            ->groupBy(['n_id', 'n_givn'])
-            ->select(['n_givn', DB::raw('COUNT(*) AS count')])
+            ->groupBy(['n_givn'])
+            ->select(['n_givn', DB::raw('COUNT(distinct n_id) AS count')])
```
fisharebest: I can't work out what this fixes.  I see the same results with old/new code.

app/Statistics/Repository/IndividualRepository.php
```diff
@@ -109,9 +109,9 @@ private function commonGivenQuery(string $sex, string $type, bool $show_tot, int
                 // Exclude initials and particles.
                 if (!preg_match('/^([A-Z]|[a-z]{1,3})$/', $given)) {
                     if (\array_key_exists($given, $nameList)) {
-                        $nameList[$given] += (int) $count;
+                        $nameList[$given] += $count;
                     } else {
-                        $nameList[$given] = (int) $count;
+                        $nameList[$given] = $count;
```
fisharebest: Surely these need to be cast from strings to integers.

The code that uses them calls `I18N::number()` which requires an integer.

magicsunday: The value is already an integer

app/Statistics/Repository/IndividualRepository.php
```diff
@@ -642,6 +643,7 @@ public function commonSurnamesListTotals(
     public function statsBirthQuery(bool $sex = false, int $year1 = -1, int $year2 = -1): array
     {
         $query = DB::table('dates')
+            ->select(['d_month', DB::raw('COUNT(*) AS total')])
```
fisharebest: Do these changes fix a problem - or do they just remove the `else`?

app/Statistics/Repository/IndividualRepository.php
magicsunday: On my site it returns 

screenshot_2019-02-21 statistiken

and after the fix it returns the same result as in 1.7

screenshot_2019-02-21 statistiken 1

app/Statistics/Repository/IndividualRepository.php
magicsunday: removes teh else

app/Statistics/Repository/IndividualRepository.php
fisharebest: I think it depends on the PDO driver.

I just tested it, and for SQLite, all columns (even numeric ones) are returned as strings.

magicsunday: :(

app/Statistics/Repository/IndividualRepository.php
fisharebest: You're right.  For my test data (royal family) there was no difference.  But I see it in other data.

fisharebest:
Thanks!

app/Statistics/Repository/IndividualRepository.php
magicsunday: There may be other locations which need a casting too when used with SQLite.

PR #2248, opened 2019-02-22
Moved color interpolation to separate class, renamed helper to servic…

magicsunday:
…e, use snake_case in classes

- Adding some class comments.


fisharebest:
@magicsunday 

There are many functions which do two or three different things - depending on a parameter.  A good example is:

```
public function statsMarrQuery(bool $first_marriage = false
```

This should really be two different functions - e.g. `statsMarriageQuery()` and `statsFirstMarriageQuery()`

Would you like to update these, or shall I?  I want to avoid us both working on the same code.

magicsunday:
> Would you like to update these, or shall I? I want to avoid us both working on the same code.

Feel free to change it.

PR #2250, opened 2019-02-22
SetupController fails to create a new database

magicsunday:
SetupController fails to create a new database


fisharebest:
There is new setup wizard that adds initial support for SQLite.

There is more to do, but it conflicts with this PR.

So, I have pushed the initial work so that the MySQL setup wizard will run.

PR #2287, opened 2019-03-01
Fixed calling static methods dynamically

magicsunday:


fisharebest:
These methods should not be static.

We need to call them via an object (= a "service" = "business logic") which is passed in.

So the fix is to change the method, not the caller.

PR #2309, opened 2019-03-06
Fixes #2307

### LINKED ISSUE #2307: 2.0 - Languages do not show up in Control Pannel / Languages
image

Recompiling languages makes no difference  composer webtrees:mo

magicsunday:
Fixes #2307

### DIFF
--- a/app/I18N.php
+++ b/app/I18N.php
@@ -470,7 +470,7 @@ class I18N
     {
         $locales = [];
 
-        foreach (glob(WT_ROOT . 'resources/lang/*',GLOB_NOSORT | GLOB_ONLYDIR) as $file) {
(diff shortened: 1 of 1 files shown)


xmlf:
Thank you!
It works very well.

fisharebest:
I just pushed a slightly different fix for this.

fisharebest:
Thanks for finding the problem.

magicsunday:
glob sort the files, this is already done afterwards.

fisharebest:
`glob()` sorts by filename - `it`, `fr`, ...

We need to sort by the local language name - "Italiano", "franćais", ...

magicsunday:
Yes but Sorting of glob could be omitted.

fisharebest:
> Yes but Sorting of glob could be omitted.

Of course!

It is not needed on any of the calls to `glob()` so I have updated all of them.

PR #2318, opened 2019-03-09
Fix unintended ajax request url

### LINKED ISSUE #1709: Descendants sidebar - Wrong behavior
In the sidebar descendants list, clicking twice on the very last entry creates an embedded copy of the current individual's page.

It seems like the last link defaults to the current page and, when triggering the expansion, is interpreted as a query.

The bug appears in both demo-dev and demo-stable. It does not occur if the last entry has the indication "no children".

ric2016:
Fixes #1709.

### DIFF
--- a/app/Module/DescendancyModule.php
+++ b/app/Module/DescendancyModule.php
@@ -191,7 +191,18 @@ class DescendancyModule extends AbstractModule implements ModuleSidebarInterface
 
         return
             '<li class="sb_desc_indi_li">' .
-            '<a class="sb_desc_indi" href="#"><i class="plusminus icon-minus"></i>' .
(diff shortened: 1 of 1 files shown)


fisharebest:
This fix works by fetching a modified URL - which returns an empty response.

A simpler fix is to detect this condition, and not fetch the resource. 42aaa3e8e

It took me a while to understand the cause of this.  Thanks for identifying the problem and PR.

ric2016:
A simpler fix is obviously preferable (I wasn't sure how this was originally intended to work).

fisharebest:
I didn't write the original code - so I don't know how it was intended to work either ;-)

PR #2335, opened 2019-03-16
Update statistics.phtml

magicsunday:
Changed breakpoints, so that statistics wont overlap.


fisharebest:
Removing the `-4` and `-8` makes the columns equal width.

Would it be better to change the breakpoint instead?  e.g. `col-lg-4` and `col-lg-8`?

magicsunday:
Yes this may be better.

fisharebest:
Thanks.  I think that looks a bit better.

PR #2354, opened 2019-03-20
Added missing "for"-attribute to the label element

magicsunday:
This was a previous pull request (https://github.com/fisharebest/webtrees/pull/1850). I changed it accordingly to your suggestions.


fisharebest:
The Bootstrap4 class has been deprecated for a long time, and has now been removed.

It has been replaced with views/templates.

I have added similar logic to the new code.

Issue #1377, opened 2017-09-30
Source Record DATA Issues (3/3): Edit Feature Requests


ric2016:
Some background:

According to the GEDCOM specification, events may be assigned to source records, e.g. as follows:

```
0 SOUR @Sx@
1 DATA
2 EVEN BIRT, DEAT
3 DATE FROM 1787 TO 1852
2 EVEN MARR
3 DATE FROM 1815 TO 1852
```

indicating that the source contains Births and Deaths from 1787 to 1852, and Marriages from 1815 to 1852. To clarify, this refers to source records, not to source citations! Unfortunately, source citations use the same tags in a different structure (e.g. 1 SOUR 2 DATA 3 DATE), which is potentially confusing and may have contributed to existing bugs.   

Feature Requests:

In webtees 1.7.x, the 'DATA' structure may only be set when creating new sources (and then only a single 'EVEN'), apart from editing the raw Gedcom. In webtrees 2.x, it may only be set via editing the raw Gedcom.

It would be useful to have editing support for

1. adding multiple 'EVEN' when creating a new source;
2. adding 'DATA' to an existing source record;
3. adding additional 'EVEN' to existing 'DATA'.

If we had 2. and 3., we could even leave the modal dialog for creating new sources as it is, and only allow to add the 'DATA' structure to existing sources.

ric2016:
2. is already possible if 'DATA' is set in the tree preferences (under "Facts for source records").

ric2016:
Not all changes from the PR have been merged - therefore, this still isn't fixed.

PR #2381, opened 2019-04-14
Fix for #1377, second attempt

ric2016:
Allows adding multiple EVENs and NOTEs, when editing a SOUR:DATA fact.
Allows adding NOTE when adding a SOUR:DATA fact.

The changes in `edit-fact.phtml` are from a previous pull request (they had not been merged).
On this occasion, I have adjusted `add-fact.phtml` accordingly.
I'm aware all of this (SOUR:DATA:EVEN) is a rather exotic part of the specification, and probably doesn't have high priority - I think after this we'll be done with this finally.

### DIFF
--- a/resources/views/edit/add-fact.phtml
+++ b/resources/views/edit/add-fact.phtml
@@ -12,6 +12,20 @@
 
     <?php FunctionsEdit::createAddForm($tree, $fact) ?>
 
+    <?php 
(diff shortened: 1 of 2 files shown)


fisharebest:
Thanks.  Again, sorry it has taken me so long to find the time to review this.

PR #2414, opened 2019-05-07
Fixed duplicate entries in census place selector

magicsunday:
Instantiated objects could not be compared with "in_array". This results in duplicate entries in the selector.

### DIFF
--- a/app/Census/Census.php
+++ b/app/Census/Census.php
@@ -75,9 +75,17 @@ class Census
                 $census_places = [];
                 break;
         }
-
+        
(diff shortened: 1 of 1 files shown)


fisharebest:
Would it work to change the third parameter of `in_array` to `false`?

fisharebest:
The third parameter was added recently, at the request of some code-analysis tools.

I added `true`, when I meant `false` :-(

magicsunday:
Maybe add some comment, why "false" is used here.

PR #2466, opened 2019-06-22
Adjust css classes for pending edits

ric2016:
When editing names and gender of an individual, the pending changes are not highlighted, because there is no css for "old" and "new". I assume "wt-old" and "wt-new" is intended here?


fisharebest:
Thanks

PR #2474, opened 2019-06-27
Add test for coverage class Age

sibador:
This is a small contribution to increase the coverage of unit tests
I added test in the Age class.


fisharebest:
This class was added accidently - and is not used.

I think we might need a "DateDifference" class - which will be similar to this.

So, I'll keep the Age class, and use it as the basis for future development.

PR #2554, opened 2019-08-24
Skip media objects without title while checking for duplicates

serl:
Hello,
this page crashes a little for me, given that I have 1395 media object, and none of them has a title.
I see that the same fix has been done in version 2!


fisharebest:
Thanks

PR #2658, opened 2019-10-20
load modules and catch exceptions on include and boot.

### LINKED ISSUE #2657: Faulty modules will break webtrees
As faulty modules can break webtrees, I will create a PR which will add some error handling and disable modules, if they are faulty.

bmhm:
Fixes #2657 

You can leave this open if you would like me to implement some more of the functionality specified in the TODOs.

### DIFF
--- a/app/Services/ModuleService.php
+++ b/app/Services/ModuleService.php
@@ -20,6 +20,7 @@ declare(strict_types=1);
 namespace Fisharebest\Webtrees\Services;
 
 use Closure;
+use Exception;
 use Fisharebest\Webtrees\Auth;
 use Fisharebest\Webtrees\Contracts\UserInterface;
 use Fisharebest\Webtrees\I18N;
@@ -210,6 +211,7 @@ use Fisharebest\Webtrees\Webtrees;
 use Illuminate\Database\Capsule\Manager as DB;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
+use RuntimeException;
 use stdClass;
 use Throwable;
 
@@ -606,7 +608,7 @@ class ModuleService
     }
 
     /**
-     * Load a module in a static scope, to prevent it from modifying local or object variables.
+     * Load a custom module in a static scope, to prevent it from modifying local or object variables.
      *
      * @param string $filename
      *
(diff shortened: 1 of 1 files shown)


app/Services/ModuleService.php
```diff
@@ -606,15 +607,44 @@ public function setupLanguages(): Collection
...
      *
      * @return mixed
      */
     private static function load(string $filename)
     {
-        return include $filename;
+        try {
+            return include $filename;
+        } catch (Exception $loading_exception) {
+            self::disable_module($filename);
```
fisharebest: We don't need to disable the modue (by renaming it).
It did not load, and is therefore already disabled.

I'm sure module developers will get very annoyed if the code renames their module every time they make a coding error or typo.

Instead, we could simply display the error.

```
FlashMessages::addMessage(e($ex->getMessage()), 'danger');
```

I don't think the full stack-dump is needed.  Errors here are likely to be class/trait/mismatches and will be clear from the error message.

app/Services/ModuleService.php
```diff
@@ -783,7 +813,16 @@ public function bootModules(ModuleThemeInterface $current_theme): void
                 continue;
             }
 
-            app()->dispatch($module, 'boot');
+            try {
+                app()->dispatch($module, 'boot');
+            } catch (Exception $boot_exception) {
+                $module->setEnabled(false);
```
fisharebest: Perhaps just display the error, as above?

app/Services/ModuleService.php
bmhm: Sure, I'd put this in:
```
FlashMessages::addMessage(I18N::translate('Module %s errored while loading.', $module_name), 'danger');
FlashMessages::addMessage($loading_exception->getMessage(), 'danger');
```

… or really just the message?

fisharebest: > really just the message

In reality, I would add a message telling the user that this module isn't compatible with webtrees, and that they should upgrade/delete it.

But I reviewed this over my morning coffee, and that was all I had time to type.... ;-)

bmhm:
@fisharebest implemented two messages in https://github.com/fisharebest/webtrees/pull/2658/commits/86539327dfaf8464638d4dbdfb8323c7142a8d88.

`pre-hook-commit` executed.

fisharebest:
You added two `try/catch` blocks.

The first is good.  It stops the module being created, and the rest of the code will not see it.

I am not certain about the second.

If the call to `boot()` fails, then a message is displayed and the code continues.
But if the boot fails, then the rest of the module will probably also fail and we will just get another error further on...

bmhm:
> You added two `try/catch` blocks.
> 
> The first is good. It stops the module being created, and the rest of the code will not see it.
> 
> I am not certain about the second.
> 
> If the call to `boot()` fails, then a message is displayed and the code continues.
> But if the boot fails, then the rest of the module will probably also fail and we will just get another error further on...

THAT WAS THE WHOLE POINT :dagger: 

Now, if a module calls `app('nonexistent')`, the page will not show a nice error, just `text/html`.

with the catch block, the module will not boot correctly, granted. But the error generated afterwards will only affect the Tab or the HTML Block.

Do you really prefer breaking a complete website instead of only a tab? :(

PR #2669, opened 2019-10-21
Show upload error for media files, too. Fixes #2655.

### LINKED ISSUE #2655: Fail to upload large document
I'm trying to add a pdf document (size 81 MB) to an event as a media object.
The upload counter reaches 38-40%, then zeros out and start again. It will again reach 38-40%. And will zero again. Thirs time it will bring an error page in the "add media" screen.

Version is 1.7.10

Thanks,
Miron

bmhm:


fisharebest:
I've just reverted this.

The code doesn't work for many reasons.

* undefined namespace (`Functions::class` not imported)
* undefined variable `$upload` (should be (`$uploaded_file`)

It doesn't actually work.  This request is made as an AJAX call, and the client side doesn't catch/display the error.

bmhm:
Then why not remove  `FlashMessages::addMessage(I18N::translate('There was an error uploading your file.'));` from the try-catch-block at the end of the method as well?

fisharebest:
FYI, `composer.json` contains several code-analysis scripts.

Several of these are bundled together as `pre-commit-hook`.

At the end of my `.git/hooks/pre-commit`, I have added this line

`composer webtrees:pre-commit-hook`

It will prevent you from committing code that fails style-checks, unit tests or static analysis.

bmhm:
> FYI, `composer.json` contains several code-analysis scripts.
> 
> Several of these are bundled together as `pre-commit-hook`.
> 
> At the end of my `.git/hooks/pre-commit`, I have added this line
> 
> `composer webtrees:pre-commit-hook`
> 
> It will prevent you from committing code that fails style-checks, unit tests or static analysis.

Please add this to https://github.com/fisharebest/webtrees/pull/2659 before merging :)

Issue #2665, opened 2019-10-20
github - Bad routing for all Search options.


makitso:
Either with search from top right menu or any search from Primary Icons.

![image](https://user-images.githubusercontent.com/5147603/67160265-e5f4fc00-f303-11e9-9b1a-8a502b3298bc.png)

dbq-andersons:
Similar behavior with the Media Objects search. Routes back to "My Page" if logged in or the default tree if not logged in.

Issue #2664, opened 2019-10-20
github - Bad routing on Reports / Ancestors / html and PDF


makitso:

fisharebest:
@makitso - are you still seeing this error?  If so, can you post any errors/details.

makitso:
Yes, still have the error:
Select Report / Ancestors

::1 - - [22/Oct/2019:16:29:37 -0700] "GET /public/favicon-32.png?v=1549812510 HTTP/1.1" 200 2003 "http://localhost/index.php?route=%2Ftree%2Ftree1%2Freport-setup&xref=I0001&report=ahnentafel_report" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.120 Safari/537.36"

Select continue

:1 - - [22/Oct/2019:16:29:45 -0700] "OPTIONS * HTTP/1.0" 200 126 "-" "Apache/2.4.41 (Ubuntu) (internal dummy connection)"
::1 - - [22/Oct/2019:16:30:02 -0700] "GET /index.php?route=report-run&tree=tree1&report=ahnentafel_report&varnames%5B%5D=pid&vars%5Bpid%5D=I0001&varnames%5B%5D=maxgen&vars%5Bmaxgen%5D=3&varnames%5B%5D=sources&vars%5Bsources%5D=on&varnames%5B%5D=notes&varnames%5B%5D=occu&varnames%5B%5D=resi&varnames%5B%5D=children&varnames%5B%5D=pageSize&vars%5BpageSize%5D=US-Letter&output=HTML HTTP/1.1" 302 348 "http://localhost/index.php?route=%2Ftree%2Ftree1%2Freport-setup&xref=I0001&report=ahnentafel_report" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.120 Safari/537.36"
::1 - - [22/Oct/2019:16:30:02 -0700] "GET /index.php?route=%2F HTTP/1.1" 302 362 "http://localhost/index.php?route=%2Ftree%2Ftree1%2Freport-setup&xref=I0001&report=ahnentafel_report" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.120 Safari/537.36"
[...]

PR #2673, opened 2019-10-21
Fixes for #2665 #2664

gachowy:
Fixes for issue in forms with GET (#2665 and #2664)

In the case of 'action' that has additional GET parameters, it was overridden by form data. In forms there is a hidden input 'route' which should work when route was defined without `route()` method. When we are using `route()` with parameters, and placing it into value of that input, and we don't have pretty URLs, than `route()` can add additional 'index.php` to final URL.

Solution prepared, was additional, optional parameter for `route()` method, to get just raw route, without domains/index.php etc.

### DIFF
--- a/app/Helpers/functions.php
+++ b/app/Helpers/functions.php
@@ -158,10 +158,11 @@ function response($content = '', $code = StatusCodeInterface::STATUS_OK, $header
  *
  * @param string  $route_name
  * @param mixed[] $parameters
+ * @param bool    $skip_url Flag for rendering routes without absolute URL - for usage in GET forms
  *
  * @return string
  */
-function route(string $route_name, array $parameters = []): string
+function route(string $route_name, array $parameters = [], $skip_url = false): string
 {
     $request          = app(ServerRequestInterface::class);
     $router_container = app(RouterContainer::class);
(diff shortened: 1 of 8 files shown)


fisharebest:
Hi @gachowy 

I'm actually working on these issues right now.

I'm using the same approach that I used on the charts.  Forms are now submitted using POST, which redirects back to the page with the correct URL.

This allows "pretty URLs" to be maintained - and allows parameters to be included in the URL.

For example, the two phonetic searches could have the algorithm parameter in the URL.  e.g.

/trees/demo/search-phonetic/Russell
/trees/demo/search-phonetic/DaitchM

This isn't possible with the current approach of using forms with GET.

bmhm:
@gachowy are you interested in joining the slack chat?

fisharebest:
PS,

Another reason for removing the `<input name="route">` field is that it links many of our templates to the implementation of our routes.  If we change the routing implementation, we'd need to change all of these.

The approach I'm using means that the routing logic is confined to just two functions.  One creates routes, and one parses routes.

gachowy:
@bmhm Hey. Of course :)
Currently, i've lot work as CTO in my company, but i've got some time, to commit in some open source projects on evenings. And secondly, geneology was my passion ;)

fisharebest:
@gachowy - thanks for taking the time to contribute.

As described previously, there is a better solution, which I have now implemented.

Therefore I shall close this PR.

Issue #2682, opened 2019-10-23
Git - Pedigree Map cannot change individual


ddrury:
Not sure if this is a knock-on from https://github.com/fisharebest/webtrees/commit/353080f8c238896c32674e7f73a3448a714cb0ac but changing the individual on the form and clicking view shows the original person.

ddrury:
Shouldn't the form in pedigree-map/page.phtml have an action?

fisharebest:
> Shouldn't the form in pedigree-map/page.phtml have an action?

We POST back to the same URL.

This post request then generates an updated URL and redirects back to it.

In the "old webtrees" style, we could just use GET with lots of hidden `<input>`s.

With pretty URLs, we can't, as we need to update the URL, not the query parameters.

PR #2684, opened 2019-10-23
Fix: PedigreeMap

ddrury:
Only single generation shown: $generations not being set correctly in pedigreeMapModule::getPedigreeMapFacts()


fisharebest:
I think this was already fixed in #2682 /  11b6f9c

The `generations` parameter is part of the URL, and definitely an "attribute" and not a query-parameter.  It just wasn't being updated.

ddrury:
Well it doesn't work for me!

Regards
David

fisharebest:
> Well it doesn't work for me

Ah.  Two related problems.

1) the `generations` parameter in the main URL not updated (I fixed this).
2) the `generations` parameter in the callback not being set (I missed this).

Issue #2681, opened 2019-10-23
Error while loading statistic charts


magicsunday:
After initial page load the charts are not generated. Best seen after switching tabs.

https://dev.webtrees.net/demo-dev/module/statistics_chart/Chart/demo

The firefox console contains the following error:

`Must call google.charts.load before google.charts.setOnLoadCallback`
```
    jo https://www.gstatic.com/charts/loader.js:289
    callbackPieChart https://dev.webtrees.net/demo-dev/module/statistics_chart/Chart/demo#tab-tab-48ebde000198465de8a86b5856f54597 line 1 > scriptElement:4
    <anonym> https://dev.webtrees.net/demo-dev/module/statistics_chart/Chart/demo#tab-tab-48ebde000198465de8a86b5856f54597 line 1 > scriptElement:29
    L https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
    We https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
    append https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
    html https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
    Z https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
    html https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
    load https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
    l https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
    fireWith https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
    D https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
    t https://dev.webtrees.net/demo-dev/public/js/vendor.min.js?v=1570700530:1
```

fisharebest:
Seems to be fixed by PR #2688

PR #2688, opened 2019-10-24
GH-2681: Refactored statistics charts, prevent multiple google chart …

magicsunday:
…loading, Fixed initial loading


fisharebest:
Thanks!

Issue #2721, opened 2019-10-31
github - Individual Tabs - facts & events


ddrury:
Clicking on a source (not the expand + icon) produces the following error
```
Undefined variable: tree …\resources\views\source-page.phtml:98
#0 …\resources\views\source-page.phtml(98): Fisharebest\Webtrees\Webtrees::Fisharebest\Webtrees\{closure}(8, 'Undefined varia...', 'D:\\Server\\nginx...', 98, Array)
#1 …\app\View.php(186): include('D:\\Server\\nginx...')
#2 …\app\View.php(282): Fisharebest\Webtrees\View->render()
#3 …\app\Helpers\functions.php(217): Fisharebest\Webtrees\View::make('source-page', Array)
#4 …\app\Http\ViewResponseTrait.php(51): view('source-page', Array)
#5 …\app\Http\Controllers\SourceController.php(97): Fisharebest\Webtrees\Http\Controllers\AbstractBaseController->viewResponse('source-page', Array)
#6 …\app\Http\Middleware\WrapHandler.php(79): Fisharebest\Webtrees\Http\Controllers\SourceController->show(Object(Nyholm\Psr7\ServerRequest))
#7 …\vendor\oscarotero\middleland\src\Dispatcher.php(136): Fisharebest\Webtrees\Http\Middleware\WrapHandler->process(Object(Nyholm\Psr7\ServerRequest), Object(Middleland\Dispatcher))
#8 …\vendor\oscarotero\middleland\src\Dispatcher.php(118): Middleland\Dispatcher->handle(Object(Nyholm\Psr7\ServerRequest))
#9 …\app\Http\Middleware\Router.php(123): Middleland\Dispatcher->dispatch(Object(Nyholm\Psr7\ServerRequest))
#10 …\vendor\oscarotero\middleland\src\Dispatcher.php(136): Fisharebest\Webtrees\Http\Middleware\Router->process(Object(Nyholm\Psr7\ServerRequest), Object(Middleland\Dispatcher))
#11 …\app\Http\Middleware\BootModules.php(62): Middleland\Dispatcher->handle(Object(Nyholm\Psr7\ServerRequest))
#12 …\vendor\oscarotero\middleland\src\Dispatcher.php(136): Fisharebest\Webtrees\Http\Middleware\BootModules->process(Object(Nyholm\Psr7\ServerRequest), Object(Middleland\Dispatcher))
[...]

PR #2722, opened 2019-10-31
Fix missing parameters

ddrury:
Fixes https://github.com/fisharebest/webtrees/issues/2721 and a similar problem with Control Panel - Manage Family trees - Import


fisharebest:
Thanks

PR #2786, opened 2019-11-19
Add facility to have name accordion closed on entry

ddrury:
I appreciate you don't like any extra preference options but I thought this may be useful to save a bit of real estate on the page


ddrury:
Bugger, the change to the en-US language file was StyleCI - I didn't mean to commit it - sorry

PR #2795, opened 2019-11-25
Github - Places hierarchy - a few tidyups & improved accessibility

ddrury:
- Minor tidy to some of the views
- created new view for the marker popup as using the markup from the sidebar proved awkward to get a decent popup display.
- improved accessibility for the sidebar (using a coloured border to indicate places without valid co-ords wasn't much good for people suffering from colour-blindness)
- Made assumption that for now the flags would be placed in a sub-folder of data. (really made this change to ensure the markup worked properly)
- Fixed layout for RTL pages

### DIFF
--- a/app/Http/Controllers/PlaceHierarchyController.php
+++ b/app/Http/Controllers/PlaceHierarchyController.php
@@ -45,6 +45,7 @@ use function assert;
 class PlaceHierarchyController extends AbstractBaseController
 {
     private const MAP_MODULE = 'openstreetmap';
+    private const FLAG_PATH = 'data/';  //Assume the flags are in a sub-folder of data
 
     /** @var SearchService */
     private $search_service;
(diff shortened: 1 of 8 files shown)


fisharebest:
Flags are almost certainly going to end up being stored in the database.
The application can only write to `/data/`, and we can only serve static assets from `/public/`.

We could serve them the same way we handle media files.  But most SVGs are tiny, so it might be better to serve them inline.

Also, all user-data is actually written to an abstract/remote filesystem (which could be dropbox, S3, google drive, etc.).  It just happens to default to using the local `/data/` folder.

I've put this off, as I also want to update the place-location system to store localised place names (as per the webtrees/geodata repository on github).  I plan to do both together - after 2.0 is released.

I'll have a look at this PR in the next day or so.

BTW - `col-xxx` classes are block-level and usually applied to `<div>`s.  Is there a significance in changing them to `<span>`s?

ddrury:
> BTW - col-xxx classes are block-level and usually applied to `<div>`s. Is there a significance in changing them to `<span>`s?

Hangover from an experiment - I'll them change back tomorrow

ddrury:
Greg,

Can I include the functional changes (excluding Flags) from this PR, in the one I'm preparing re extra-markers?

With regard to the flags, I appreciate your long term goals and that the current "solution" offends the structural integrity of the application, but I figured the for people currently using flags in 1.7, a simple shift of the places folder from the googlemaps module to somewhere else (public?) would be a reasonable short term solution. After all, whatever you do long term, you'll either need to provide a migration path or users will have to start from scratch.

PR #2801, opened 2019-11-29
Github - Mapping - change icon package

ddrury:
Here are all the relevant files changes during the shift from package Leaflet.BeautifyMarker to package leaflet.extra-markers.

As well as replacing the above package, a longstanding problem on the place hierarchy map was revealed whereby places without valid co-ordinates had an invisible marker placed at co-ords (0,0), with the new package these became visible. We now no longer send data for invalid places to the mapping view

When I initially wrote this mapping code I intended that place-map.phtml would be general purpose and used for all maps, therefor the number of parameters passed in had to cover every eventuality, now that we have different views for each map the parameters for each can be simplified

I hope you have more luck that me in compiling the new package!!


fisharebest:
I think I have managed to split the commit into the separate changes, and merge them (except for the icon library).

I found a "dirty hack" to use the existing beautifymarker library with FA5.

PR #2806, opened 2019-12-01
Maps missed bits

ddrury:
Renamed the Place Hierarchy views to indicate that fact and fixed a couple of things you missed in https://github.com/fisharebest/webtrees/pull/2801.


fisharebest:
I've merged this, and had another think about display of places without co-ordinates.

PR #2820, opened 2019-12-03
Map views bug fixes & tidy

ddrury:
tab.phtml - fix bug basedata.I18N no longer exists.
PedigreeMapModule - fix major bug - line/icon colours were changing for every person not every set of parents, rename misleading variable.
All views: change classes to remove some local css and tidy use of JS sidebar variable.


fisharebest:
Merged

PR #2863, opened 2019-12-11
Swap colorbox for lightgallery

ddrury:
A few notes,

Lightgallery seems to be actively supported, and whilst it doesn't support wheelzoom, it does have a zoom function which allows scroll/pan by mouse/finger of the zoomed image. [See the Lightgallery site](http://sachinchoolur.github.io/lightGallery/)

I've use Lightgallery's dynamic function so that duplicates can be removed from the list of images to be displayed. (A classic example of duplicates is on the families tab where an individual has a "main image" therefore that image appears for the individual and each family of which (s)he is a member)

The way I've set it up means that for tabbed pages only the visible images on the active tab are shown, however it is easy to revert to showing all images on the page (remove ':visible' from the selector).

Tested with Firefox, Chrome & Edge on desktop and Chrome on Android


ddrury:
Going to remove this at the moment - strange things happening when not logged in

PR #2864, opened 2019-12-11
Replace Colorbox with Lightgallery

ddrury:
Lightgallery seems to be actively supported, and whilst it doesn't support wheelzoom, it does have a zoom function which allows scroll/pan by mouse/finger of the zoomed image. See the [Lightgallery site](http://sachinchoolur.github.io/lightGallery/)

I've use Lightgallery's dynamic function so that duplicates can be removed from the list of images to be displayed. (A classic example of duplicates is on the families tab where an individual has a "main image" therefore that image appears for the individual and each family of which (s)he is a member)

Tested with Firefox, Chrome & Edge on desktop and Chrome on Android

To speed thing up a bit set the lightgallery thumb parameter to false


ddrury:
Greg,

Just noticed npm run production isn't inlining the css images. I've tried adding the line     `.copy("node_modules/lightgallery/dist/img/*", "public/css/images")` to webpack.mix.js and the images are copied but the css isn't being changed!

I think I've been here before but any suggestions?
 BTW if all else fails I could update `resources/css/_lightgallery.css` to over-ride the supplied css

fisharebest:
> I think I've been here before but any suggestions?

I'm pretty tied up with feedback/issues from the 2.0.0 release.

I'll have a look at this as soon as I get some time.

ddrury:
> I'm pretty tied up with feedback/issues from the 2.0.0 release.
> 
> I'll have a look at this as soon as I get some time.

Fully understood

ddrury:
I think I can see the problem but I'm not sure how to fix it!`postcss_image_inliner` has assetPaths set to `resources/css`. which means it doesn't include anything that is compiled into `public/vendor.css`. I checked leaflet and by chance it's relative path for urls is correct but its images aren't inlined either

ddrury:
See Issue https://github.com/fisharebest/webtrees/issues/2871 for solution to above problem

UksusoFF:
Seems https://github.com/fancyapps/fancybox more active supported.
Last accepted PR in lightGallery was in Jan 2019.

ddrury:
Following @UksusoFF's comment I've just done an installation of Fancybox, however in my opinion Lightgallery is the better product. On the desktop Fancybox only has one level of zoom (to the native size of the image) - can't tell on my tablet as the screen isn't big enough. performance isn't quite as good and facilities are fewer.

Nevertheless, if anyone is interested I'll push the code up

fisharebest:
> however in my opinion Lightgallery is the better product.

Then let us stick with lightgallery for the time being.

PR #2865, opened 2019-12-11
Fix a couple of validation errors

ddrury:


fisharebest:
The `<label></label>` breaks the display for me.  (I only tried in firefox).

Maybe add this to the `<input>` instead?

`aria-label="<?= I18N::translate('Media folder') >"`

ddrury:
changed as suggested

PR #2880, opened 2019-12-13
Add individuals w/o links to "unrelated" list

### LINKED ISSUE #2741: Results incomplete in "Find unrelated individuals"
_(I raised this issue in the Forum already [1] a couple of weeks ago, but this might be the more appropriate place for it.)_

I noticed that the "Find unrelated individuals" report (in "Control panel" -> "Manage family trees" -> "Genealogy data") does not include individuals without any family record, e.g. those added with the "add unlinked individual" option, or those imported via GEDCOM without any FAMS/FAMC info.

The issue is caused by the method of selecting individuals in UnconnectedPage.php [1], which is only considering individuals listed in the "link" table, so individuals without any links are never looked at. 

I think it should be enough to add an $individual_group with all individuals not mentioned in the "link" table right after looping through the $components, but I don't feel familiar enough with the code yet to do it myself.

----
[1] https://www.webtrees.net/index.php/en/forum/3-help-for-2-0-beta/33601-results-incomplete-in-find-unrelated-individuals
[2] https://github.com/fisharebest/webtrees/blob/96716c4735cd29561cdf33f324c111f41c544214/app/Http/RequestHandlers/UnconnectedPage.php

t-karcher:
Fixes #2741

### DIFF
--- a/app/Http/RequestHandlers/UnconnectedPage.php
+++ b/app/Http/RequestHandlers/UnconnectedPage.php
@@ -60,11 +60,20 @@ class UnconnectedPage implements RequestHandlerInterface
             $links = ['FAMS', 'FAMC'];
         }
 
-        $rows = DB::table('link')
+        // select individuals with at least one link
+        $linked_individuals = DB::table('link')
             ->where('l_file', '=', $tree->id())
             ->whereIn('l_type', $links)
-            ->select(['l_from', 'l_to'])
-            ->get();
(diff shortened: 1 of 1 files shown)


t-karcher:
Oops. Needs more testing first. Will reopen later.

fisharebest:
I think the solution is simply something like

$graph = select i_id from indivdiuals
left join links
where link is null

PR #2884, opened 2019-12-14
Adjust order of facts

ric2016:
The order of facts for a new individual is currently mainly alphabetic (with death events coming last).

If e.g. in tree preferences, Facts for new individuals are set to:
ALIA,BAPM,BIRT,BURI,DEAT
we get the facts in the edit dialog in exactly this order, which isn't nice.

This PR adjusts this order to:
1. Birth events, in the order [BIRT, others]
2. Other non-birth non-death events
(3. marriage fact, as before)
4. Death events, in the order [DEATH, others]


fisharebest:
I've just submitted a slightly different fix for this problem, which keeps all the sorting logic in one place.

It also reduces the amount of logic in the templates.

PR #2942, opened 2020-01-04
A new batch update plugin

JustCarmen:
This plugin helps finding Level 1 Primary Media Objects from webtrees 1 and place these objects in the gedcom before the other media objects to use these in webtrees 2 as preferred images in the caroussel, on charts etc.

As discussed on Slack. You asked me to make a pr but you also suggested to solve it in an alternative way by implementing this from the OBJE records - not from the INDI records.  This would allow us to delete the _PRIM as part of the process. I've thought about that but I don't see how it can be achieved if we also want to change the sort order in the INDI records too.  An object can be linked to multiple INDI records and one INDI record can have multiple _PRIM tagged objects.

But I agree it would be better to remove the _PRIM tag as part of the process. But you can not do this before the user pushes the update button while the sorting needs to be done in another record as the removing.

So this is what I came up with so far. It might not be ideal but it works. Please feel free to adjust it. I might learn from it :)

### DIFF
--- a/app/Module/BatchUpdate/BatchUpdatePrimaryMediaPlugin.php
+++ b/app/Module/BatchUpdate/BatchUpdatePrimaryMediaPlugin.php
@@ -1,0 +1,138 @@
+<?php
+
+/**
+ * webtrees: online genealogy
+ * Copyright (C) 2019 webtrees development team
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+declare(strict_types=1);
+
+namespace Fisharebest\Webtrees\Module\BatchUpdate;
+
+use Fisharebest\Webtrees\I18N;
+use Fisharebest\Webtrees\Tree;
+use Fisharebest\Webtrees\Individual;
+use Fisharebest\Webtrees\GedcomRecord;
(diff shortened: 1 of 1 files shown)


ddrury:
Hi Carmen,
I had the same thought, and came up with the following. I've been busy since Xmas and haven't fully tested it - eg. I've assumed only one `1 _PRIM` record per person - is this true? I also thought about deleting them from the media Obj but haven't got that far.

```
    private function update(GedcomRecord $record): string
    {
        $sorted = [];
        preg_match_all('/1 _WT_OBJE_SORT @(.*)@/', $record->gedcom(), $sort_id);
        preg_match_all('/1 OBJE @(.*)@/', $record->gedcom(), $media_id);

        // Is there a 1 _PRIM Y record
        foreach ($media_id[1] as $id) {
            $media_rec = Media::getInstance($id, $record->tree());
            $primary   = $media_rec->facts(['_PRIM']);
            if ($primary->isNotEmpty() && $primary->first()->value() === 'Y') {
                $sorted[] = $id;
                break;
            }
        }
        // add from _WT_OBJE_SORT lines
        // but first get rid of spurious sort lines (level 2?)
        $unused = array_diff($sort_id[1], $media_id[1]);
        // then check for media without a sort record
        $needed = array_diff($media_id[1], $sort_id[1]);
        $sorted = array_unique(array_merge($sorted, array_diff($sort_id[1], $unused), $needed));

        //delete old media related lines
        $new_ged = preg_replace('/\n1 .*?OBJE.*/', '', $record->gedcom());

        //add the newly sorted lines
        foreach ($sorted as $item) {
            $new_ged .= sprintf("\n1 OBJE @%s@", $item);
        }

        return $new_ged;
    }
```

JustCarmen:
> I've assumed only one 1 _PRIM record per person - is this true? 

In my own Gedcom I've found one case where two different _PRIM objects where linked to the same INDI record. However, this is not a rare situation since objects behave independent from INDI records. In the case two _PRIM objects were linked to the same person, the first in order was select as highlighted image. So I think we need to be aware of this situation and remove the _PRIM tag from both linked objects.

I don't have time to test your code right now, probably later this week.

JustCarmen:
The conversion of Primary Media Objects is include in the data-fix module set (commit ce42304a0e3b715b5830d587f136ecd5ca69f11b). So this PR can be closed now.

Issue #2991, opened 2020-01-20
V2 Query on vendor.css


ddrury:
When working on Lightgallery, I thought that some of the fonts were looking a bit "anaemic" so switching back to master and looking at vendor.css, I found that the body styling in bootstrap was being over-ridden by style.css in leaflet-geosearch (line 127).

Shouldn't bootstrap be the last file in the webpack.mix.js merge process so it's not over-ridden?

fisharebest:
IMHO, leaflet-geosearch certainly shouldn't be styling the `<body>` element.

However, the pinned issue at https://github.com/smeijer/leaflet-geosearch/issues
suggests that there is little point creating an issue.

> Shouldn't bootstrap be the last file in the webpack.mix.js merge process so it's not over-ridden?

Bootstrap may need to come before datatables-bs4 and typeahead-bs4.
But moving this package first might be an option.

Maybe there's alternative package that it still supported/maintained?

ddrury:
I've checked all the other files that comprise vendor.css and there is nothing else that interferes with bootstrap. There are a whole load of alternatives on the leaflet plugins page - I'll have a rummage

ddrury:
This looks like the best bet https://github.com/perliedman/leaflet-control-geocoder

PR #2992, opened 2020-01-21
V2 Update views/admin/location-edit.phtml - replace Geosearch

ddrury:
This should be installed after https://github.com/fisharebest/webtrees/pull/3088. as script in views/admin/location-edit would have changes introduced here over-written

- Replaced geosearch with leaflet-control-geocoder (Fixes issue https://github.com/fisharebest/webtrees/issues/2991)
- Added map reset button 
- Removed redundant code.

### DIFF
--- a/app/Http/Controllers/Admin/LocationController.php
+++ b/app/Http/Controllers/Admin/LocationController.php
@@ -328,7 +328,6 @@ class LocationController extends AbstractAdminController
             'parent_id'   => $parent_id,
             'lat'         => $lat,
             'lng'         => $lng,
-            'ref'         => $id,
             'data'        => $this->mapLocationData($id),
             'provider'    => [
                 'url'     => 'https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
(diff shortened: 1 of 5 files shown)


fisharebest:
Something up with my styling?  Can you post a screendump showing what it *should* look like?

Screen Shot 2020-01-21 at 12 09 36

ddrury:
I sometimes find I have to run `npm run production` twice to get everything in place. Compiling with sass seems to fix it.

Anyway here's a couple of screenshots
demo1
demo2

fisharebest:
> I sometimes find I have to run npm run production twice

Aha.  That does it.

ddrury:
Re npm run production. I _think_ the problem is sequencing in that the postCss line runs before mix.styles produces vendor.css therefore on first run the previous vendor.css is used. On 2nd run the recently produced vendor.css is compressed.

By using sass, all the css files are merged and minified in one action so it always works

ddrury:
Thats it, I'm done now!

PR #3008, opened 2020-01-24
Fix location data import bug

ddrury:
Import fails to write to the database when the _Add and Update_ option is selected

### DIFF
--- a/app/Http/Controllers/Admin/LocationController.php
+++ b/app/Http/Controllers/Admin/LocationController.php
@@ -734,28 +734,26 @@ class LocationController extends AbstractAdminController
 
         foreach ($places as $place) {
             $location = new Location($place['fqpn']);
-            $exists   = $location->exists();
+            $new   = !$location->exists();
 
-            if ($options === 'update' && !$exists) {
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Admin/LocationController.php
```diff
@@ -744,7 +744,7 @@ public function importLocationsAction(ServerRequestInterface $request): Response
                 $added++;
             }
 
-            if (!$exists || $options === 'update') {
+            if (!$exists || stripos($options, 'update') !== false) {
```
fisharebest: Does `$options` have a fixed list of possible values, or is it a completely variable string that might contain anything.

If the former, then it would be clearer to write `!$exists || $options === 'update' || $options === 'update2' || $options === 'do-update'`

ddrury:
You're a bit quick off the mark - Give me a couple of minutes and I'll 
shove a new version up. I've renamed $exists so I could get my brain 
around the logic better

Regards
David

ddrury:
Options can have the values "update", "addupdate" (which caused the initial problem) and "add"

PR #3068, opened 2020-02-07
Replace colorbox with lightgallery

ddrury:
fixed problem with webpack compiling assets - see note in webpack.mix.js
allow each instance to have own options
track upstream - follow js semistandard where possible


ddrury:
I'm closing this as it's been up for a long time so I assume you're not interested, nevertheless I shall keep it up to date for my own use.

PR #3070, opened 2020-02-08
Demonstrate usage of sass compilation...

ddrury:
I've just noticed something else that I guess can be classed as a benefit.
All the images referenced in vendor.min.css have been inlined. Apart from the sass compilation I can't see anything else that'd do this...

### DIFF
--- a/resources/sass/vendor.scss
+++ b/resources/sass/vendor.scss
@@ -1,0 +1,32 @@
+/**
+ * webtrees: online genealogy
+ * Copyright (C) 2020 webtrees development team
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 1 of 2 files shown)


fisharebest:
> All the images referenced in vendor.min.css have been inlined.

This is currently done explicitly using postCss in the current code.

Does this mean that `.sass()` automatically calls `.postCss()`?  I can't find any documentation.

However, not all images are inlined.  The leaflet-geosearch ones aren't.  I can't work out why not.

ddrury:
I'm confused, leaflet-geosearch doesn't have any images mentioned in it's css (or any in the module!). 

There is (was?) a problem with leaflet and webpack, see https://github.com/ghybs/leaflet-defaulticon-compatibility. It's old and I've not found anything that says it's still necessary.

> This is currently done explicitly using postCss in the current code.

Yeah, but the asset path used for .postCss() is resources/css and with sass we don't do anything in there. So it must either be sass itself or as you say sass calls .postCss() automatically. 

There is a bit here https://laravel.com/docs/6.x/mix#url-processing. that shows setting processCssUrls: false with sass so maybe that indicates it's enabled by default.

ddrury:
Tried setting processCssUrls false as above - didn't make any difference!

fisharebest:
> /* Ensure bootstrap comes last to avoid being overridden */

Can you elaborate on this.

We currently load bootstrap first.  What gets overrideen?

ddrury:
Nothing now but see https://github.com/fisharebest/webtrees/issues/2991. But wouldn't it make sense from a defensive programming viewpoint.

fisharebest:
Hi Dave - I've merged this in - but with a slight modification.

I use a vendor.css (instead of vendor.scss)

It makes it clearer that we are processing plain CSS.

It seems to work fine (no need for a double compilation run...).

But if I've misunderstood something, let me know.

Issue #3085, opened 2020-02-16
function app() and the IoC container


ric2016:
in functions.php, `app()` is documented as "Get the IoC container, or fetch something from it."

I find it a bit surprising that using app e.g.

`$factories = app(GedcomRecordFactories::class);`

does not return a singleton instance from the container, but instead creates a new object every time, which is often not helpful. Is that intentional? Singletons have to be created explicitly via
```
if (!Application::getInstance()->bound(GedcomRecordFactories::class)) {
  Application::getInstance()->singleton(GedcomRecordFactories::class);
} 
```

Should this perhaps be the default logic when using `app()` (without additional parameters)?

fisharebest:
> Is that intentional?

We are using the laravel container - and this is how is works.

>  Singletons have to be created explicitly via

There is  a simpler syntax to create a singleton

```
app()->instance(GedcomRecordFactoryInterface::class, new MyGedcomRecordFactory())
```

Factories are not usually singletons - because they have no state or resources.

You'd generally use bind your implementation to an interface.  e.g.

```
app()->bind(GedcomRecordFactoryInterface::class, `MyGedcomRecordFactory`.::class)
```

Now, when a method type-hints `GedcomRecordFactoryInterface`, the container will create a `MyGedcomRecordFactory`.

ric2016:
> We are using the laravel container - and this is how is works.

But we could implement app() in functions.php differently, or at least have another method there which returns singletons. I guess my question is, is there at all some use of app() within the code where we actually must not obtain a singleton?

A factory may not have been the best example (but why shouldn't a factory have state? In my case, it does - for caching). Using another example, `app(Statistics::class)`: What is the point of not using a singleton Statistics here? Every call constructs a new Statistics object, which seems like unnecessary overhead.

fisharebest:
> Every call constructs a new Statistics object, which seems like unnecessary overhead.

Creating objects in PHP is very fast.  My guess is that this overhead is so small, you can not measure it reliably.

> but why shouldn't a factory have state?

Global state is generally bad.  It means the output of a function depends on not
just the inputs to the function, but also *every previous call to the package*.

Stateless classes are easier to refactor, easier to test, easier to debug.

> In my case, it does - for caching

Caching should be done using the cache object.  (Most of the similar "caches" have already been converted).

It makes testing easier.  I can set up a "null" cache.

ric2016:
> Global state is generally bad.

Except when it's explicitly required.
[...]

Issue #3073, opened 2020-02-09
Module initialization order


ric2016:
In some of my custom modules, I have to replace certain webtrees classes, such as GedcomRecord. This isn't nice but generally works by modifying the class loader from the custom module.

However, classes that are already in use cannot be replaced (e.g. a recent modification of SiteMapModule.php causes GedcomRecord to be initialized there, apparently).

Can we load custom modules prior to core modules, in order to give them a chance to carry out such replacements? This would make webtrees more customizable.

This would be a simple change in `ModuleService.php` lines 540-541 (function 'all'), swapping the two function calls:

`return $this->customModules()
                ->merge($this->coreModules())`

fisharebest:
> isn't nice but generally works by modifying the class loader

:-)

I think you would have the same problem if another custom module used (and hence loaded) that class before you...

> swapping the two function calls:

That order was intentional.  Firstly, I thought that custom modules should be able to rely on the core modules being available/initialised.  Secondly, I thought it would make it harder for custom modules to break the application.

I will admit that I didn't think in too much detail about this, and all the consequences.
But at the moment, I would prefer not to change it.

I *think* the correct solution is to have a `GedcomRecordFactory` object, which is responsible for creating `GedcomRecord` objects.

e.g. replace

```
$indivdiual = Individual::getInstance(...);
```

with

```
$factory = app(GedcomRecordFactory::class);
$individual = $factory->getIndividualInstance(...);
```

We would also need to declare `IndividualInterface`, etc.

Then, you could register your own factory, which would create your own objects (which implement the Interface).

I'm not sure how easy it would be to convert all the code to this structure.

fisharebest:
> I'm not sure how easy it would be to convert all the code to this structure.

Perhaps not *too* difficult.

The existing `getInstance()` functions would simply need to be replaced with calls to the factory object.

fisharebest:
This is actually a change that I've been thinking about for a while, as it would make testing easier.

ric2016:
GedcomRecord was just one example. But I generally agree that it would be preferable to modify the webtrees core code, so that custom modules won't need these replacements in the first place.

Regarding the actual GedcomRecord extension, I use an approch where factories for specific types may be registered (so I guess it is overall a GedcomRecordFactoryFactory), like this:

`GedcomRecordExt::addFactory('_LOC', new SharedPlaceFactory(...));`

[...]

PR #3091, opened 2020-02-22
allow use of GedcomRecordFactory for specific types, cleaned up.

ric2016:
again, see #3073 (also #3085).

Use of internal cache in GedcomRecordFactories (taken from GedcomRecord) should probably be refactored to use main cache.

### DIFF
--- a/app/GedcomRecord.php
+++ b/app/GedcomRecord.php
@@ -67,10 +67,6 @@ class GedcomRecord
 
     protected const ROUTE_NAME = GedcomRecordPage::class;
 
-    /** @var GedcomRecord[][] Allow getInstance() to return references to existing objects */
-    public static $gedcom_record_cache;
-    /** @var stdClass[][] Fetch all pending edits in one database query */
-    public static $pending_record_cache;
     /** @var string The record identifier */
     protected $xref;
     /** @var Tree  The family tree to which this record belongs */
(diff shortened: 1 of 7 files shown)


fisharebest:
FYI, I am not 100% happy with the factory code in the develop branch.

Specifically, it prevents many other classes (e.g. app/Services/*) from using dependency injection.

ric2016:
Why are they prevented from using it? Do you mean remaining calls e.g. to Individual::getInstance, as in IndividualListService?

fisharebest:
> Why are they prevented from using it?

Because these objects can be created *before* the modules are booted - which means that they will use the default factory objects - and not the ones created by your module.

fisharebest:
For example, the DescendancyModule uses the SearchService - which needs IndividualFactory and FamilyFactory.

ric2016:
I see - Possible solutions:

* Boot the (custom) modules earlier - This may not be possible/desirable for other reasons?

* Do not inject dependencies in the Constructor, but only when they are actually used the first time. This is how it's currently done in SearchService - Is that a problem? I guess it's harder to keep track of dependencies this way.

fisharebest:
> * Boot the (custom) modules earlier
> * Do not inject dependencies in the Constructor, but only when they are actually used the first time.

There are two ways to obtain an IndividualFactory (or any other object).

1) Dependency injection.  This is the pattern used throughout the code.

2) Obtain from the container.  We generally use this only for external resources, such as filesystems and caches.

For objects that are created **after** we boot the modules, both will work exactly the same.

For objects that are created **before** we boot the modules, both will *appear* to work. But may give different factory objects.  There will be no difference for the core code - just for people using your module.

If we are creating factory objects to support your module, then there is a problem when the easy/default/obvious way to use them is incompatible with your module.

I really do like the factory methods.  It is good to separate the logic of indivduals/families/etc. from the database.  But I do not see a good way to allow modules to override them.

You have two requirements.

1) create new types of record (`_LOC` = SharedLocation).  What if two modules want to create a `_LOC` object?

2) override exising classes (`Individual`).  What if two modules want to modify `Individual` ?

Maybe a solution is to pass all new objects to the modules?

e.g.

```
$record = new GedcomRecord(...);

foreach ($modules as $module) {
    $record = $module->createRecord($record);
}
```

The module could contain this:

```
function createRecord(GedcomRecord $record): GedcomRecord
{
  if ($record type is `_LOC`) {
    return new SharedLocation(...);
  } else {
    return $record;
  }
}
```

ric2016:
These 'what if two modules modify the same part of the system' problems occur elsewhere as well, e.g. when a module changes a view. I don't think it's possible to handle them in general, other than 'first (or last) change wins'.

`foreach ($modules as $module) {
    $record = $module->createRecord($record);
}`

I don't think this would actually allow two modules creating e.g. _LOC objects to work together, unless they are already aware of the other module being potentially enabled. In general, the last module would be just as likely to simply replace the previous module's changes as with the current solution.

> There are two ways to obtain an IndividualFactory (or any other object).

I'm not sure I understand the difference - 'Dependency injection' is the case where a constructor argument is resolved via the container? So ultimately in both cases, the object is obtained from the container? 

Maybe just disallow dependency injection for factories then?

ric2016:
> There are two ways to obtain an IndividualFactory (or any other object).

Perhaps this should be addressed first more generally in the overall architecture - Is it a good idea to use the container for both purposes?

See e.g. [here](https://martinfowler.com/articles/injection.html#ServiceLocatorVsDependencyInjection) - Basically, we don't have either Service Locator or Dependency Injection, but both concepts mixed, and in particular both resolved via the same container.

(By the way, the PSR-11 spec advises against using the container for the service locator pattern, see [here](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container.md#13-recommended-usage))

If we want dependency injection in general, but a service locator for specific cases (external resources, such as filesystems and caches, as you pointed out; and maybe now the factories as well), should we perhaps make the distinction clearer e.g. by using separate containers?

PR #3093, opened 2020-02-25
Geographic data table add key ...

ddrury:
...  to indicate badge meanings.

Could probable do away with the badge-secondary (grey )and merge with badge info (blue-green)

### DIFF
--- a/resources/views/admin/locations.phtml
+++ b/resources/views/admin/locations.phtml
@@ -7,9 +7,15 @@ use Fisharebest\Webtrees\Webtrees;
 
 <?= view('components/breadcrumbs', ['links' => $breadcrumbs]) ?>
 
-<h1><?= $title ?></h1>
-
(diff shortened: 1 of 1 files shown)


fisharebest:
Thanks - but we need to replace the colour-coding entirely.

We are displaying information using **only** colour.
This is not accessible to users who are colour-blind.

Instead, I guess we need separate badges/icons/indicators/text/whatever. for each piece of information.

* number of child places (badge?)
* missing co-ordinates (warning icon?)
* unused places (some other icon?)

fisharebest:
Thanks - but we need to replace the colour-coding entirely.

We are displaying information using **only** colour.
This is not accessible to users who are colour-blind.

Instead, I guess we need separate badges/icons/indicators/text/whatever. for each piece of information.

* number of child places (badge?)
* missing co-ordinates (warning icon?)
* unused places (some other icon?)

PR #3119, opened 2020-03-17
Update messages.po

ramu98:
fix to line 936


fisharebest:
It is better to update the translations using https://translate.webtrees.net

PR #3141, opened 2020-03-27
Update messages.po

ramu98:


ramu98:
In the Korean messege.po file of Webtrees, most of the parts used in Korea have been translated. However, it is not approved because there have been errors since the last version. I am not a programmer. I can't find what is the problem. If there is a wrong LOGIC, please help us fix it.

fisharebest:
PLEASE DO NOT CREATE PULL-REQUESTS FOR PO FILES.

It creates problems to merge them.

All updates should be at translate.webtrees.net.

You can upload .PO files directly to translate.webtrees.net

There was an error merging this file.

The updates may be lost.

PR #3145, opened 2020-03-29
Allow sendmail command to be configured

dnrce:
Swift Mailer's default of `/usr/sbin/sendmail -bs` isn't always sufficient.

### DIFF
--- a/app/Http/RequestHandlers/EmailPreferencesAction.php
+++ b/app/Http/RequestHandlers/EmailPreferencesAction.php
@@ -63,6 +63,7 @@ class EmailPreferencesAction implements RequestHandlerInterface
 
         $params = (array) $request->getParsedBody();
 
+        Site::setPreference('SENDMAIL_COMMAND', $params['SENDMAIL_COMMAND']);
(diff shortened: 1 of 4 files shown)


fisharebest:
Do you think we should add validation to this parameter?

Allowing any value to be passed to `proc_*()` makes me nervous.

> Swift Mailer's default of /usr/sbin/sendmail -bs isn't always sufficient.

What changes do you need?  Different path?  Different executable name?  Different parameters?

dnrce:
In my own case I need `/usr/bin/sendmail -oi -t`.

Good point on security. Would it be a better approach to only make this configurable via `config.ini.php`?

fisharebest:
> Would it be a better approach to only make this configurable via config.ini.php?

Yes, I think so.  (See <https://12factor.net/config>)

fisharebest:
I have updated the documentation at https://webtrees.net/faq/email/

Can you test the latest code for me?  Thanks.

dnrce:
I would have updated but thanks for throwing the fix together so quickly! I have tested by applying this patch to my 2.0.3 installation and it works great.

dnrce:
:+1: on the updated documentation as well. Thanks!

PR #3149, opened 2020-03-29
Treat empty values in coordinates like NULL values

### LINKED ISSUE #3132: New place is not recognised as missing coordinates.
When I am adding a new place in gedcom, in Control panel/Geographic data is the place not displayed with yellow point, so there is no possibility to find places without coordinates.

I think it is because of automaticaly added zoom factor

image

Other place without any data:

image

v2.0.3
Ladislav

t-karcher:
Fixes #3132 

While strictly speaking the error is caused by the id() function in webtrees/app/Location.php (which initializes a new location with an empty string instead of NULL values), fixing this would not fix existing data, and would also not prevent data coming from elsewhere (e.g. via migration) to be formatted properly. So I think it's best to fix the symptoms (treatment of empty strings in coordinates) in this case instead of searching for and fixing all current and future root causes.

### DIFF
--- a/app/Http/Controllers/Admin/LocationController.php
+++ b/app/Http/Controllers/Admin/LocationController.php
@@ -214,16 +214,16 @@ class LocationController extends AbstractAdminController
         $prefix = DB::connection()->getTablePrefix();
 
         $expression =
-            $prefix . 'p0.pl_place IS NOT NULL AND ' . $prefix . 'p0.pl_lati IS NULL OR ' .
-            $prefix . 'p1.pl_place IS NOT NULL AND ' . $prefix . 'p1.pl_lati IS NULL OR ' .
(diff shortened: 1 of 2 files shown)


t-karcher:
@miqrogroove : Webtrees is not saving numbers in the database, but strings with prefix "E/W" or "N/S":

image

 So, "0" should never appear there (and if it does, it's worth a warning sign, I guess), and "E0" does not trigger the warning:

image

miqrogroove:
I was looking at the wrong variables then.  `pl_long` would not normally contain `"0"` that's correct.

miqrogroove:
I'm going to update #3326 to match this style.  These changes are in the same file, so they should be consistent.

fisharebest:
@t-karcher - I don't like the `empty()` function in PHP.  It has too many subtle behaviours, and is makes code difficult to follow.

So, it was easier to push an alternative - instead of asking you to change the code.

But thanks for the solution!

PR #3248, opened 2020-05-15
Translate type of media on media page

### LINKED ISSUE #3239: Type of media not translated
Type of media is translated in media_list, but not in media.
media

t-karcher:
Fixes #3239


fisharebest:
Thanks!

PR #3256, opened 2020-05-17
Bootstrap 4.5 fixes for F.A.B and Minimal themes

ddrury:


fisharebest:
I think I pushed the exact same change about 60 seconds before your PR...

PR #3287, opened 2020-05-22
Add a timestamp in the export filename

### LINKED ISSUE #3217: Export shouldn't overwrite
If i export a tree at server, it will be named as %tree%.ged. That would replace older files.

It better should be %TREE%%DATE%%TIME%.ged to avoid, that older backups are overwritten.

Like automaticly backups, while updates, are already doing it.

t-karcher:
Fixes #3217


fisharebest:
I've chosen a different solution in the fix for #3217

PR #3293, opened 2020-05-23
Better handling of default tags for unrecognised custom tags

### LINKED ISSUE #1685: Editing an event that is defined without a date shows date field twice - demo-dev
When I edit an individual- or family-event that exists in the Gedcom without a date, I see Date 2 times in the edit form.

If I edit an event that has a date or I add a new event, I see the date as I should, only once.

See the Not married fact of f47 in demo-dev.

t-karcher:
- List of default tags for unrecognized custom tags added to Config.php
- Check added in FunctionsEdit.php to prevent the addition of duplicate date tags (fixes #1685)

### DIFF
--- a/app/Config.php
+++ b/app/Config.php
@@ -443,4 +443,22 @@ class Config
             'MARR',
         ];
     }
+    
+    /**
+     * A list of facts that are provided in the create / edit form
+     * for all unrecognised custom tags
+     *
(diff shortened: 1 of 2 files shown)


fisharebest:
We now have a new system for handling tags, which solves this problem in a cleaner way.

Issue #3255, opened 2020-05-16
Places Map Sometimes Doesn't Work


miqrogroove:
When I hit an individual page for the first time during a session, the Places tab and map works normally.

If I switch to another tab, such as Album or Facts and events, and then refresh the page, the Places tab will not work correctly.  I get a blank map and have to refresh the page again with the Places tab active.

ddrury:
I can't reproduce this at all, When you say "will not work correctly" what happens? Any errors anywhere?

miqrogroove:
Every time I follow those steps, it looks like this:
![bug](https://user-images.githubusercontent.com/1371835/82148149-91c1e400-9820-11ea-8c52-1b698afa603f.png)

ddrury:
A few things to try in order. Using the Browser developer tools

- On the network tab, anything unusual? any error statuses? you should see a number of images received from <letter>.tile.openstreetmap.org are they OK
- On the console tab, there should be a line XHR GET (current page). If you expand it and click on the response tab - Response payload, does the result look like` webtrees-2.0\resources\views\modules\places\tab.phtml`?

If these are both OK, insert the following at line 99 of tab.phtml (3rd line of _buildMapData) `console.log(data);` you should then see in the console tab `Object { type: "FeatureCollection", features: (x) […] }` where x is the number of events for this person.

miqrogroove:
I'm still getting status 401 for `/public/site.webmanifest?v=2.0.3` in the Console and Network tabs.

Literally the only tile being loaded is `https://c.tile.openstreetmap.org/2/1/1.png` unless I refresh the page.

XHR GET is disabled by default for the Console tab.  After enabling that, I do see it fetched `/module/places/Tab/tree1?xref=...` and when expanded there are several mentions of `vendor.min.js?v=2.0.3:2`

miqrogroove:
And sorry but I don't know how to find the file you mentioned.

https://github.com/fisharebest/webtrees/search?q=mapdata+filename%3Atab.phtml&unscoped_q=mapdata+filename%3Atab.phtml

miqrogroove:
`{type: "FeatureCollection", features: Array(18)}`

I should point out that the blue sidebar is working in this situation.  It's only the map itself that's messed up.

miqrogroove:
Also could not find any "response tab".  Still uncertain if my DevTools is configured the same as yours.

ddrury:
[...]

PR #3294, opened 2020-05-24
Fix Places Map Load Failure

miqrogroove:
Because the XHR script sometimes runs prior to rendering the osm-map element, it is necessary to bind _buildMapData to a different event named shown.bs.tab, and then unbind it later so it won't run more than one time.

See https://github.com/fisharebest/webtrees/issues/3255

### DIFF
--- a/resources/views/individual-page.phtml
+++ b/resources/views/individual-page.phtml
@@ -153,7 +153,10 @@ use Illuminate\Support\Collection;
   "use strict";
 
   // Bootstrap tabs - load content dynamically using AJAX
-  $('a[data-toggle="tab"][data-href]').on('show.bs.tab', function () {
+  $('a[data-toggle="tab"][data-href][href!="#places"]').on('show.bs.tab', function () {
+    $(this.getAttribute('href') + ':empty').load($(this).data('href'));
+  });
+  $('a[href="#places"]').on('shown.bs.tab', function () {
     $(this.getAttribute('href') + ':empty').load($(this).data('href'));
   });
 
--- a/resources/views/individual-page.phtml
+++ b/resources/views/individual-page.phtml
@@ -153,11 +153,19 @@ use Illuminate\Support\Collection;
   "use strict";
 
   // Bootstrap tabs - load content dynamically using AJAX
-  $('a[data-toggle="tab"][data-href][href!="#places"]').on('show.bs.tab', function () {
-    $(this.getAttribute('href') + ':empty').load($(this).data('href'));
-  });
-  $('a[href="#places"]').on('shown.bs.tab', function () {
-    $(this.getAttribute('href') + ':empty').load($(this).data('href'));
+  $('a[data-toggle="tab"][data-href]').on('show.bs.tab', function () {
+      let target = $(this.hash + ':empty');
+      if (target) {
+          // Start the download immediately...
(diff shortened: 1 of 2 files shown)


miqrogroove:
Looks like this doesn't cover all situations so...

fisharebest:
To be sure I understand the problem:

Bootstrap uses CSS transition to hide/show the tabs.
This transistion lasts 150ms.

`.bs.tab.show` fires at the start of the transition.
`.bs.tab.shown` fires at the end of the transition.

webtrees uses the `.bs.tab.show` event as a trigger to load dynamic content.

The map relies on CSS properties that are not available/correct until the end of the transition.

On a fast server, it is possible to load and render the map before the transition is complete.


This is a problem that might affect other tabs in third-party modules.
Adding an event listener to `.bs.tab.shown` means that this code can only
be used inside a bootstrap tab.  IMHO, we should avoid this tight coupling.
I think we can currently embed "tabs" in any other page/site/application.


In #3255, you said

> `fitBounds()` is called from the wrong event.  You're using `show.bs.tab` for everything, but the `osm-map` doesn't render until `shown.bs.tab`

Perhaps a better solution is to load the tab content on `shown.bs.tab`?
It makes everything much simpler, and "just works".

Yes, it means that the tab will load up to 150ms later - but this is barely noticable.
Loading the map tiles takes much longer.

What do you think?

miqrogroove:
Yes, you've got the bug figured out.

My patch worked so well that I didn't want to slow down the maps by moving everything to the late event.

There are one or two other ways to handle the timing within `individual-page.phtml` but this still involves calling `_buildMapData()` separately from the XHR.  In effect, it is coupled or using a callback and not a "private" function anymore.  This would allow the map to be reused without specific references to bootstrap, but it changes the way the map is loaded.

Or, just move the one XHR to the late event and it will "just work".

miqrogroove:
Either way, the `#places` tab will have to be handled differently from all other tabs.

fisharebest:
> Either way, the #places tab will have to be handled differently from all other tabs.

I'd prefer to avoid special handling like this.

If a user created a copy of this module (with minor modifications), then it wouldn't work.

Javascript isn't my strongest skill - hence the ugly mix of jQuery and vanilla JS.

This works for me.  We start the download immediately, using a promise.
On the second event, we wait for the promise and insert the data.

What do you think of this?  Can you clean it up at all?

```
  // Bootstrap tabs - load content dynamically using AJAX
  $('a[data-toggle="tab"][data-href]').on('show.bs.tab', function () {
      let target = document.querySelector(this.hash + ':empty');
      // Only fetch data once.
      if (target !== null) {
          // Start the download immediately...
          let tab_content = fetch(this.dataset.href);

          // ...but don't insert it until the tab is ready to receive it.
          $(this).on('shown.bs.tab', async function () {
              await tab_content
                  .then(data => data.text())
                  .then(function (data) {
                      $(target).html(data);
                  });
          });
      }
  });
```

miqrogroove:
I like your version.  Tested and working.  You're not supporting Internet Explorer, correct?

fisharebest:
> You're not supporting Internet Explorer, correct?

There are probably many things that don't work in IE.

But it is a long time since anyone raised an IE compatibility issue, so I guess there aren't (m)any IE users left.

PR #3297, opened 2020-05-24
Genealogical numbering system support

divinity666:
Added support for the common genealogical numbering system of ancestors (ref. [https://en.wikipedia.org/wiki/Genealogical_numbering_systems]).

Numbering of individuals in the reports (especially the text reports) should be a common understandable system and not only an incrementing number system. This especially becomes obvious in trees where far away generations are no more filled with all ancestors. Anyway the numbering system proposed here also in this situation assures a reproducible numbering system if the same individual is tested.

### DIFF
--- a/app/Report/ReportParserGenerate.php
+++ b/app/Report/ReportParserGenerate.php
@@ -2605,7 +2605,7 @@ class ReportParserGenerate extends ReportParserBase
     }
 
         /**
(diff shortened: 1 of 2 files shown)


arbor95:
This pull request is not implemented/accepted for more than 3 years and so I think it should be removed/closed.

PR #3300, opened 2020-05-25
Show links to linking records in DeleteRecord.php

### LINKED ISSUE #3298: Improve confirmation page after deletion of individuals
Currently, this is what you see after deleting a spouse of a family:

image

While there's nothing wrong with it, I'd still like to suggest two improvements:
1) Don't route back to the deleted individual, but to the main page instead: We know it's deleted, so why go back there?
2) In the "link deleted" info message, it would be helpful to link the affected family, so I can immediately go back there and check the results / continue to work with it.

t-karcher:
Links added to all records that previously linked to the newly deleted individual or family.

Closes #3298 (after some reconsideration, I now agree that routing to the main page is not a good idea)

### DIFF
--- a/app/Http/RequestHandlers/DeleteRecord.php
+++ b/app/Http/RequestHandlers/DeleteRecord.php
@@ -79,12 +79,12 @@ class DeleteRecord implements RequestHandlerInterface
                             $new_gedcom = $this->removeLinks($new_gedcom, $linker->xref());
                             $relict->updateRecord($new_gedcom, false);
                             /* I18N: %s are names of records, such as sources, repositories or individuals */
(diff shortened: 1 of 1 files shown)


fisharebest:
I like this solution, but can you change some things:

1) Do not create a new translation string with embedded HTML.  Instead, create the link and insert it into the existing string.

2) You should add the class `alert-link`.  See the https://getbootstrap.com/docs/4.4/components/alerts for details

3) You need to escape the URL, using `e($z->url())`.  Otherwise `&` characters in the URL will not be encoded correctly.

fisharebest:
Merged.  Thanks!

PR #3306, opened 2020-05-26
Sort Charts menu entries

ric2016:
Sort Charts menu entries (just like Lists menu entries).
Without this, custom module charts appear at the top currently (as reported [here](https://github.com/vesta-webtrees-2-custom-modules/vesta_extended_relationships/issues/4)).

(the empty '->filter()' operation in ListsMenuModule doesn't seem to have any effect?)

### DIFF
--- a/app/Module/ChartsMenuModule.php
+++ b/app/Module/ChartsMenuModule.php
@@ -95,6 +95,9 @@ class ChartsMenuModule extends AbstractModule implements ModuleMenuInterface
         $submenus   = $this->module_service->findByComponent(ModuleChartInterface::class, $tree, Auth::user())
             ->map(static function (ModuleChartInterface $module) use ($individual): Menu {
                 return $module->chartMenu($individual);
+            })
+            ->sort(static function (Menu $x, Menu $y): int {
+                return $x->getLabel() <=> $y->getLabel();
             });
 
         if ($submenus->isEmpty()) {
--- a/app/Module/ListsMenuModule.php
+++ b/app/Module/ListsMenuModule.php
@@ -92,7 +92,6 @@ class ListsMenuModule extends AbstractModule implements ModuleMenuInterface
             ->map(static function (ModuleListInterface $module) use ($tree): ?Menu {
                 return $module->listMenu($tree);
             })
-            ->filter()
             ->sort(static function (Menu $x, Menu $y): int {
                 return $x->getLabel() <=> $y->getLabel();
             });


fisharebest:
`findByComponent()` already sorts modules.

I did try browsing your repository on github, but I couldn't find the definition for the "vesta symbol" (it is late and I am tired...)

So your module sorts first because you have some sort of emoji symbol at the front of the module name?

fisharebest:
If there are no objects of a given type, the list menu modules can return null instead of a menu object.

You do not need a "Repository list" option when there are no repositories.

So the `filter()` is for these.

I guess your test environment has at least one record of every type!

ric2016:
1. No, this has nothing to do with the 'Vesta' symbol. Even without this symbol, it comes first. See the screenshot in the original issue [here](https://github.com/vesta-webtrees-2-custom-modules/vesta_extended_relationships/issues/4)).

> findByComponent() already sorts modules

But not by label. The default sorting is used e.g. in the control panel, that's ok. For the list menu, we want to sort differently (apparently). Compare to the current sorting in ListsMenuModule.php. This should be consistent with ChartsMenuModule.php.

2. regarding filter(): I don't think that's how filter() works?
"The filter method filters the collection using the given callback, keeping only those items that pass a given truth test:"

Test: I remove all list modules. I remove the filter(). The List menu still doesn't show. This is handled by the check starting [here](https://github.com/fisharebest/webtrees/blob/fecbc3c6d765c35658db2293fa4d1197d252911d/app/Module/ListsMenuModule.php#L99).

fisharebest:
> "The filter method filters the collection using the given callback, keeping only those items that pass a given truth test:"

If no callback is provided, it removes null entries.

`->chartMenu()` always returns a `Menu` object.
`->listMenu()` returns either a `Menu` object or null.

I tested this.
I deleted the `->filter()` line.
I then deleted the only repository in my test site.
I now get this error, because the `RepositoryListModule->listMenu()` returns `null`.

```
Fatal error: Uncaught TypeError: Argument 2 passed to Fisharebest\Webtrees\Module\ListsMenuModule::Fisharebest\Webtrees\Module\{closure}() must be an instance of Fisharebest\Webtrees\Menu, null given in /Users/gr4376/Projects/webtrees/app/Module/ListsMenuModule.php:96
```

> But not by submenu label. The default sorting (by module name) is used e.g. in the control panel, that's ok. For the list menu, we want to sort differently (apparently). 

OK - I see it now.  I think all the core modules use the module-name as the menu name.

There is another place we must do this - the ReportMenuModule should also sort by menu name.
Can you update this one as well?

ric2016:
ok, now I understand the filter().

I'll create a new PR, including ReportMenuModule.

PR #3307, opened 2020-05-27
Sender mail address added to contact message body

### LINKED ISSUE #2899: github - Contact me email does not include senders email address
image

t-karcher:
Fixes #2899

### DIFF
--- a/resources/views/emails/message-user-html.phtml
+++ b/resources/views/emails/message-user-html.phtml
@@ -9,7 +9,7 @@ use Fisharebest\Webtrees\I18N;
     </p>
 
     <p>
-        <?= /* I18N: %s is a person's name */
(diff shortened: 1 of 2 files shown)


fisharebest:
Using parentheses like this is problematic.

1) You have LTR text inside them, and on an RTL page, one of the parentheses will be rendered LTR and the other RTL.  So you will see two `(` or two `)`.
2) Not all languages use latin punctuation.

If parethesis must be used, then they need to be part of a translation string.
This will allow translators to use different characters and also to insert hidden directional override characters.

Generally, I try to find a different construct.  Using dashes works well in many contexts - but possibly not this one.

t-karcher:
Good point. I'm not creative enough to do it without parantheses, so I now added them to the translation string as suggested.

PR #3314, opened 2020-05-31
Provide an override hook (for custom xrefs)

ric2016:
This adds an object type parameter to $tree->getNewXref(), so that custom modules may easily restore the option to use custom xref prefixes, as in 1.x. The functionality of the method itself is not modified.

(as suggested e.g. here: https://www.webtrees.net/index.php/en/forum/help-for-2-0/33978-identities-in-gedcom-file#74475)

By the way, is there a reason to have separate methods 'createIndividual', 'createFamily', 'createMediaObject'? It seems that their functionality is the same as 'createRecord'.

### DIFF
--- a/app/Functions/FunctionsImport.php
+++ b/app/Functions/FunctionsImport.php
@@ -1015,7 +1015,7 @@ class FunctionsImport
             ->value('m_id');
 
         if ($xref === null) {
-            $xref = $tree->getNewXref();
+            $xref = $tree->getNewXref('OBJE');
             // renumber the lines
             $gedrec = preg_replace_callback('/\n(\d+)/', static function (array $m) use ($level): string {
                 return "\n" . ($m[1] - $level);
--- a/app/Http/Controllers/AdminTreesController.php
+++ b/app/Http/Controllers/AdminTreesController.php
@@ -994,7 +994,7 @@ class AdminTreesController extends AbstractBaseController
         $xrefs = $this->duplicateXrefs($tree);
 
         foreach ($xrefs as $old_xref => $type) {
-            $new_xref = $tree->getNewXref();
+            $new_xref = $tree->getNewXref($type);
             switch ($type) {
                 case Individual::RECORD_TYPE:
                     DB::table('individuals')
--- a/app/Tree.php
+++ b/app/Tree.php
@@ -446,10 +446,16 @@ class Tree
     public function createRecord(string $gedcom): GedcomRecord
     {
         if (!Str::startsWith($gedcom, '0 @@ ')) {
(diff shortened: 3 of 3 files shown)


fisharebest:
> By the way, is there a reason to have separate methods 'createIndividual', 'createFamily', 'createMediaObject'? It seems that their functionality is the same as 'createRecord'.

There is (or was?) code that used the new objects immediately.
So, it was important to have a record with the correct type.

----------

I do not think that the functions `getNewXref()` and `createXxxx()` should be in the class `Tree`.

It is not the responsibility of the Tree to create new records.

Instead, perhaps, they belong in the record factory objects?

Then perhaps most of `class FunctionsImport` also belongs in the factory objects?

ric2016:
Oh, I did not mean to open this particular can of worms ;-)
We could use the opportunity to think about it. Although it's really a separate issue.

* I agree the `createXxxx()` methods could be moved to the factories.
* The `getNewXref()` method is global (over all trees), so it should be in `Site`maybe? (But how to override there from a custom module? - That would require a SiteService, or an XRefFactory)
* FunctionsImport is messy ...

Not sure if all this refactoring would be very useful right now in general.

fisharebest:
I do not want to add an API/hook if it will be removed/replaced in the future.

> Oh, I did not mean to open this particular can of worms ;-)

Too late ;-)

> FunctionsImport is messy ...

`FunctionsXXXX` (and the reports) are the only code that remains from phpGedView.
I hope to replace all of them this year.

>The getNewXref() method is global (over all trees), so it should be in Sitemaybe? 

IMHO, this function should be internal to the record factories.
We do not need it anywhere else.

To allow customisation, perhaps:

```
class AbstractGedcomRecordFactory {
  protected function xref($prefix = 'X', $suffix = '') {...}
}
```

Then you can create a new factory with a different prefix.

```
class MyIndividualFactory extends IndividualFactory {
  IndividualFactory::xref() {
    return parent::xref('I');
  }
}
```

ric2016:
In general I agree.

However, if the `getNewXref()`function is completely internal to the factory (and not part of its interface), custom modules cannot override only this part specifically.

This is the problem again where two custom modules want to adjust different aspects, but both have to customize the same factory. Even with the method in the interface, it's tricky:
Each module would have to play nice and in turn
* retrieve the current IndividualFactory (etc.) from the service locator
* implement a custom factory which delegates to this current factory for most methods, except those that the module actually intends to change
* set this custom factory at the service locator

Alternatively, we further split up the factories by specific functionality, and now require an `IndividualNextXrefFactory` etc?

fisharebest:
> IMHO, this function should be internal to the record factories.
> We do not need it anywhere else.

We do need it somewhere else - the "renumber GEDCOM" function requires it.

Issue #3291, opened 2020-05-23
Place List Shows World Map When Children Have No Coodinates


miqrogroove:
If you create a places such as Dallas, Texas, USA and Houston, Texas, USA and then specify coordinates for Texas, USA, when you then navigate to the Place List page for Texas, the map shows the entire world with no indication of where Texas is located.

fisharebest:
> If you create a places such as Dallas, Texas, USA and Houston, Texas, USA and then specify coordinates for Texas, USA, when you then navigate to the Place List page for Texas, the map shows the entire world with no indication of where Texas is located.

I did this.  I have this data.

```
mysql> select * from wt_placelocation;
+-------+--------------+----------+---------------+---------+---------+---------+---------+
| pl_id | pl_parent_id | pl_level | pl_place      | pl_long | pl_lati | pl_zoom | pl_icon |
+-------+--------------+----------+---------------+---------+---------+---------+---------+
|     1 |            0 |        0 | United States |         |         |       2 |         |
|     2 |            1 |        1 | Texas         | W100    | N31     |       5 |         |
|     3 |            2 |        2 | Dallas        |         |         |       2 |         |
|     4 |            2 |        2 | Houston       |         |         |       2 |         |
+-------+--------------+----------+---------------+---------+---------+---------+---------+
4 rows in set (0.00 sec)
```

I see this:

<img width="1200" alt="Screenshot 2020-06-05 at 22 35 11" src="https://user-images.githubusercontent.com/2306088/83924563-d801c780-a77c-11ea-9494-90e09bdfc583.png">

The zoom level is set in the database.  If I edit the location and change the zoom level from 5 to 8, I see this:

<img width="1173" alt="Screenshot 2020-06-05 at 22 39 03" src="https://user-images.githubusercontent.com/2306088/83924808-69713980-a77d-11ea-9b3c-d97a9f3d005f.png">

How do I reproduce the problem?

miqrogroove:
In that case, simply visit the Texas page instead of the United States page.  Its children have no coordinates and it won't even try to show you where Texas is located.

fisharebest:
> visit the Texas page instead of the United States page

Sorry - I thought I was!

As you might guess, I didn't create this module, and rarely use it.

miqrogroove:
In the situation where the current place has a zoom level saved, is it possible to use the saved zoom level instead guessing the "number of degrees"?

fisharebest:
[...]

PR #3324, opened 2020-06-06
Implement Saved Zoom Levels in Place List Pages

### LINKED ISSUE #3229: Geographic data: Some Maps display Incorrectly.
Greg as requested I have made a new WebTrees Test tree containing 17 individuals, and includes the problems with the maps I mentioned on the Forum.
https://webtrees.wilson-mciver.com/index.php?route=%2Ftree%2FWTtest

Search for Ann Greg or click here:-
https://webtrees.wilson-mciver.com/index.php?route=%2Ftree%2FWTtest%2Findividual%2FI6241%2FAnn-Gregg

If you click on her death place <Mirehouse, Keswick, Cumberland, England> the following map is displayed
 
image

**This is NOT the correct map**, I have accurately defined <Mirehouse>, under <Control panel / Geographic data / England / Cumberland / Keswick / Mirehouse>.
The following Map is the one that should be displayed here.
image

 If you click on her Burial place <St. Bega Church, Mirehouse, Keswick, Cumberland, England> the First Map above is correctly Displayed. 

=============================================
 

A similar situation exists for Roger Williamson
https://webtrees.wilson-mciver.com/index.php?route=%2Ftree%2FWTtest%2Findividual%2FI6232%2FRoger-Williamson
If you Click on <Baptism Torpenhow, Cumberland, England>, the following Map is displayed.
 
image

This isn’t correct as it doesn’t show <Torpenhow> the town which is nearby and is itself defined in the following map.
 
image

This is the Map I would expect to be shown.

I trust this is the information you required?
What did you want me to do with the <Places-Global.csv> file?

miqrogroove:
Fixes #3229 and improves on the patch from #3291

### DIFF
--- a/app/Http/Controllers/PlaceHierarchyController.php
+++ b/app/Http/Controllers/PlaceHierarchyController.php
@@ -301,10 +301,10 @@ class PlaceHierarchyController extends AbstractBaseController
                 'features' => $features,
             ],
             'parent' => [
-                'zoom'     => $parent->zoom(),
-                'latitude' => $parent->latitude(),
-                'longitude'=> $parent->longitude(),
-                'bounds'   => $parent->boundingRectangle(),
+                'zoom'      => $parent->zoom(),
+                'latitude'  => $parent->latitude(),
+                'longitude' => $parent->longitude(),
+                'bounds'    => $parent->boundingRectangle(),
             ]
         ];
     }
--- a/app/Http/Controllers/PlaceHierarchyController.php
+++ b/app/Http/Controllers/PlaceHierarchyController.php
@@ -236,16 +236,17 @@ class PlaceHierarchyController extends AbstractBaseController
         $features  = [];
         $sidebar   = '';
         $flag_path = Webtrees::MODULES_DIR . 'openstreetmap/';
-        $show_link = true;
+        $has_child = $places !== [];
         $is_mapped = false;
 
-        if ($places === []) {
-            $places[] = $placeObj;
-            $show_link = false;
-        }
+        // Include parent with its child list so we can see it even if they are in separate locations.
+        $places[] = $placeObj;
 
         foreach ($places as $id => $place) {
             $location = new PlaceLocation($place->gedcomName());
+
+            // Careful, $id != $placeObj->id()
+            $is_child = $place->id() != $placeObj->id();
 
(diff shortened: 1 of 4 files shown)


miqrogroove:
Just for brainstorming, this could be improved slightly if we have some way to test if the `$features` are not bounded by the saved zoom level.  I haven't tested that so far because it's not an issue on my site.

miqrogroove:
@fisharebest can you confirm for me that the Place->id() value is precached and not generating a query for each call?  I've placed that inside the controller loop and wasn't sure about the DB performance.

fisharebest:
> can you confirm for me that the Place->id() value is precached and not generating a query for each call?

Yes.  See https://github.com/fisharebest/webtrees/blob/2.0.5/app/Place.php#L112-L141

miqrogroove:
That looks like something different.

Let's say I'm calling `->id()` on 300 place objects in one page.  Is this going to run 300 queries, and then cache them?  Or is the id value already cached before I call on it?

fisharebest:
If there are 300 different places, then yes, this will be 300 different calls.

We'll also have 2x300 calls to count the individuals/families at each place.

miqrogroove:
Could you help suggest better code for my line 249?  It's only purpose is to compare place objects.  I'd rather not use the database at all.  😇

app/Http/Controllers/PlaceHierarchyController.php
```diff
@@ -236,16 +236,18 @@ protected function mapData(Place $placeObj): array
...
-            $show_link = false;
-        }
+        // Include parent with its child list so we can see it even if they are in separate locations.
+        $places[] = $placeObj;
 
         foreach ($places as $id => $place) {
             $location = new PlaceLocation($place->gedcomName());
 
+            // Careful, $id != $placeObj->id()
+            $is_child = $place->id() != $placeObj->id();
```
miqrogroove, line 249: Needs optimization.  Halp!  😳

miqrogroove: I think I've figured it out now.  Just using `gedcomName()` instead.

miqrogroove:
This should be all better now.  I didn't realize the ID call was so expensive.  Just using the gedcom string instead.

miqrogroove:
Some parts have been replaced by #3357.  Any remaining bugs will be covered by future patches from the same fork.

PR #3325, opened 2020-06-06
Fix Wrong Zoom Level on Places Tabs

### LINKED ISSUE #3235: Wrong Zoom Level in Places
This is something we resolved in 1.7 with Google Maps a while back, but seems to be a problem again in 2.0 with OSM.

When an individual only has multiple events in the same place, but only one place, the map is using a default zoom level of maximum zoomed in, rather than the zoom level saved in the geo data.  For most places, this results in a blank or mostly blank map that must be zoomed out for context.

miqrogroove:
Fixes #3235


miqrogroove:
See new PR.

PR #3326, opened 2020-06-07
Fix Unexpected Blank Map for NULL Values from v1.7

miqrogroove:
Fixes additional map failures described in https://github.com/fisharebest/webtrees/issues/3229#issuecomment-629739387

### DIFF
--- a/app/Http/Controllers/Admin/LocationController.php
+++ b/app/Http/Controllers/Admin/LocationController.php
@@ -349,30 +349,27 @@ class LocationController extends AbstractAdminController
      */
     private function mapLocationData(int $id): array
     {
-        $valid = $id !== 0;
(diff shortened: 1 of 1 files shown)


fisharebest:
The function `mapLocationData()` seems redundant.

It is only used in once - the edit location page.

That page has access to the `$location` object - so it can get the details from it directly.

Everywhere that we want a lat/lon and the information isn't present, we use the parent's lat/lon.  So, I have built this into `Location::latitude()` and `Location::longitude()`.

So, I will close this PR and push some alternative (and I hope simpler) code instead.

PS - I don't use the `empty()` in webtrees code.  It is far too ambiguous and confusing, and makes code difficult to read.  Instead use strict comparisons/typings.

PR #3343, opened 2020-06-11
 Fix incomplete logic to fetch parent coordinates

miqrogroove:
Corrects multiple problems found in https://github.com/fisharebest/webtrees/commit/e340fe65d66c30f8416fe6cc6a22bb4eb8a4bb1d

### DIFF
--- a/app/Http/Controllers/Admin/LocationController.php
+++ b/app/Http/Controllers/Admin/LocationController.php
@@ -295,8 +295,8 @@ class LocationController extends AbstractAdminController
                 $hierarchy     =  [];
             } else {
                 $hierarchy = $this->getHierarchy($parent_id);
-                $tmp       = new PlaceLocation($hierarchy[0]->fqpn);
-                $title     = e($tmp->locationName());
+                $location  = new PlaceLocation($hierarchy[0]->fqpn);
+                $title     = e($location->locationName());
             }
         }
 
--- a/app/PlaceLocation.php
+++ b/app/PlaceLocation.php
@@ -163,16 +163,13 @@ class PlaceLocation
      */
     public function getValidData()
     {
-        $tmp = $this;
-        $parent = $tmp->parent();
-        $pl_lati = (string) $tmp->details()->pl_lati;
-        $pl_long = (string) $tmp->details()->pl_long;
-        while (($pl_lati === '' || $pl_long === '') && $parent->id() !== 0) {
+        $parent = $this;
+        do {
             $tmp = $parent;
             $parent = $tmp->parent();
             $pl_lati = (string) $tmp->details()->pl_lati;
             $pl_long = (string) $tmp->details()->pl_long;
-        }
+        } while (($pl_lati === '' || $pl_long === '') && $parent->id() !== 0);
 
         return $tmp;
     }
--- a/app/PlaceLocation.php
+++ b/app/PlaceLocation.php
@@ -157,6 +157,26 @@ class PlaceLocation
     }
 
     /**
+     * Get the parent, if $this has no coordinates.
+     *
+     * @return object
+     */
+    public function getValidData()
+    {
+        $tmp = $this;
+        $parent = $tmp->parent();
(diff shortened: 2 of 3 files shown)


fisharebest:
I really want to stop using "zoom level".

Many reasons.  e.g. it depends on screen size. 
If you set the zoom level on a desktop, then many points will be off-screen on a mobile.
If you set the zoom level on a moble, then all the points will be in the centre of the screen on desktop.

Instead, we should specify a bounding rectangle and tell the map to fit that rectangle.

Perhaps it is time to drop the zoom column from the database and replace `pl_lati` and `pl_long` with floating point values.

This would make it so much easier to find a bounding rectangle based on child locations.
e.g. `select min(latitude), max(latitude), min(longitude), max(longitude) where parent_id=??`

miqrogroove:
> If you set the zoom level on a desktop, then many points will be off-screen on a mobile.

This one should not be a real problem because the displays that use multiple pins always fit the pins instead of using the saved zoom.

The one repeatable problem I've addressed in my other pull requests is that there are situations where the pins are in the same place or close to each other and the Leaflet logic is using max zoom instead of considering the saved zoom value at all.

In any case, where we edit the places using a single pin in the center of the map, I don't think switching to mobile is a huge concern.

> If you set the zoom level on a moble, then all the points will be in the centre of the screen on desktop.

So the difference I see here is that maybe when editing a place that was created on a mobile device I might think "wow this needs to be zoomed in one or two more levels so that I can fill the screen / not see so much surrounding area / etc".  Likewise, my solution where I limit the zoom level to the highest saved zoom level among multiple pins might result in a desktop map that is clustered slightly too much.

> Instead, we should specify a bounding rectangle and tell the map to fit that rectangle.

I think it would be appropriate to begin saving this in the database when editing places.  I just needs to be logical.  If "bounds" are the desired data rather than "zoom" we have to be careful how this is treated when changing devices and map aspect.  e.g. are the bounds going to get slightly larger every time someone switches devices and edits a place, because it has to fit the new screen shape?

> Perhaps it is time to drop the zoom column from the database and replace `pl_lati` and `pl_long` with floating point values.

How will this work without destroying all of the saved place views?  Were you thinking of devising a conversion between zooms and bounds?

> This would make it so much easier to find a bounding rectangle based on child locations.
> e.g. `select min(latitude), max(latitude), min(longitude), max(longitude) where parent_id=??`

Leaflet does this already.  I guess I'm not sure the context of what you mean though.

miqrogroove:
There's another bug in here.  Now that `$data` is gone, the form is reading the name of `$location` even when adding a new place.  I still need to add back another variable somewhere to make it work properly again.

miqrogroove:
Yet another regression:  Because the parent coordinates are being fetched by the PlaceLocaiton object in all situations, the Place Hierarchy list now displays all child pins at the parent's location when they actually have no defined location.  I'll work on this one too.

fisharebest:
>  Because the parent coordinates are being fetched by the PlaceLocation object in all situations,

Perhaps that isn't needed any more?

fisharebest:
PS - I just pushed some changes that I was working on this morning.

miqrogroove:
Man you are breaking things faster than I can patch them 😂

miqrogroove:
I fixed the earlier regression in #3324 since it affected those files directly.  

At this point my fork is substantially conflicted with the master branch.  I also won't be able to test that new code on my production site, so it will have to wait until I have free time to upgrade my beta image.

fisharebest:
> Man you are breaking things faster than I can patch them 😂

Sorry - I was on my lunchbreak and ran out of time for testing.

miqrogroove:
Now that 2.0.6 is released, is it correct to assume that I need to refactor all of these patches to work with https://github.com/fisharebest/webtrees/commit/9a9dfcf740849bfd5295d079a25083580a41c902 and the loss of all zoom info?

miqrogroove:
Some of these changes were replaced by #3357.  The remaining bugs will be addressed in new patches.

PR #3362, opened 2020-06-20
Always display saved map when editing locations

### LINKED ISSUE #3353: Edit (or Add) Place Showing Wrong Zoom and Non-Centered Pin
The pin and the edit view are not centered on the same coordinates.

For example, in my fork of 2.0.3 I could edit Italy or add a place to Italy and it would show me a map of Italy with a pin in the middle.

In 2.0.6 release it shows me a zoomed-in part of the middle of Italy with a pin in the top third of the map.

Places that have no children are displayed as a pin on the map of the parent place.  Again, this is not what I saved and less helpful than my old fork.  I can understand why we want to do this on the Places tab with multiple pins, but on the editing screens it feels more like the system is showing me the wrong place/wrong information.

I suspect this is all related to the new bounding rectangle logic, which is just less intuitive for editing.  I'd much rather edit what I've saved in my database than see my saved pin stuck on a different view of the map.

miqrogroove:
Fixes #3353

### DIFF
--- a/app/Http/Controllers/Admin/LocationController.php
+++ b/app/Http/Controllers/Admin/LocationController.php
@@ -319,6 +319,7 @@ class LocationController extends AbstractAdminController
             $latitude        = '';
             $longitude       = '';
             $map_bounds      = $parent->boundingRectangle();
+            $map_zoom        = $parent->zoom();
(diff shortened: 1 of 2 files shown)


miqrogroove:
Closing request due to lack of traction.

PR #3367, opened 2020-06-20
Fix use of LocalizationService

ric2016:
Construct instance of LocalizationService via the app container, so that custom modules have a chance to replace this service (in this context, elsewhere it's already ok).


fisharebest:
FYI, this entire class has already been removed in the `develop` (2.1) branch

PR #3417, opened 2020-07-18
Allow translation of custom module metadata even when disabled

### LINKED ISSUE #3413: Modules titles and description not translated when disabled
If the modules provides the translations for their own title and description (which is very likely), those are not loaded in the module list (`/admin/modules`) when the module is disabled, so they appear not-translated.

Should be just add the `$include_disabled = true` parameter in the following line, or are you afraid if could bring some unstability? 
https://github.com/fisharebest/webtrees/blob/7b284074109540a9b0565c5e7ba90553c940acef/app/I18N.php#L315

image

jon48:
Proposal to fix #3413

### DIFF
--- a/app/I18N.php
+++ b/app/I18N.php
@@ -312,9 +312,12 @@ class I18N
         // Add translations from custom modules (but not during setup, as we have no database/modules)
         if (!$setup) {
             $translations = app(ModuleService::class)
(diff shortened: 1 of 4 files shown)


jon48:
Fixed in 7b30610bdbe446efd1a5af081b1960eed3847c58

PR #3436, opened 2020-07-26
Historic events - British prime ministers

ddrury:
Qualify title to differentiate from PM of other countries.


fisharebest:
I'm not 100% certain this is correct...

The United Kingdom (of Great Britain and Ireland) was created in the 1801 act of union.

So prior to this, would these have been prime ministers of "Great Britain", rather than "The United Kingdom"?

ddrury:
See https://www.gov.uk/government/history/past-prime-ministers/robert-walpole

ddrury:
Updated to take account of act of Union. Even if you don't accept this, please ammend dates for William Petty & William Cavendish Bentinck

PR #3437, opened 2020-07-27
Qualify title to differentiate from PM of other countries

ddrury:
Take account of act of union & partition of Ireland. Corrected dates in office for William Petty & William Cavendish Bentinck (please correct these even if you don't accept the rest of this PR)


ddrury:
Just a though  ...
How about structuring this like
```
1 EVEN
2 TYPE Prime Minister
2 DATE ...
2 PLAC (or AGNC) United Kingdom ...
```

fisharebest:
> How about structuring this like

I thought of that initially, but (a) I didn't think it displayed quite so well, and (b) it wasn't strictly the location of the fact.

I also thought of adding links (notes or inline source structures), but thought that it detracted from the "secondary" status of these facts.  They shouldn't dominate the page.

BTW - I was only suggesting that the pre-1801 titles be changed to "of Great Britain".  The subsequent ones can continue to use the short form "United Kingdom".  I'll merge this, then do a search/replace.

PR #3460, opened 2020-08-02
improve edit control for SOUR/DATA/EVEN

ric2016:
This enhances the changes made a while back (#1377 etc):
The EVEN field is text such as "2 EVEN CHR, MARR, BURI", but we should provide a proper edit control for this. Events that may be selected here are configurable in the tree preferences.
The implementation uses a custom select2 control in order to preserve the order of elements.

NicerEditControl


fisharebest:
I will keep this on hold.

It would be too difficult to merge into both the `master` and the develop` (2.1.0) branches.

Once `develop` is merged into master, I will look at it again.

ric2016:
What if I provide a separate PR for the delopment branch with the same functionality?
Assuming that part will have to re-implemented anyway.

I also assume that 2.1.0 is several months away at least? I'd really like to conclude this earlier than that.

PR #3461, opened 2020-08-02
Fix: Link spouse to individual

ric2016:
fix link-spouse-to-individual.phtml

1. add ajax-modal in order to enable helptexts
2. add PLAC subtags (latitude/Longitude etc.) (without this the respective icon was without function, also better for consistency in general)


ric2016:
It may not be clear what this PR accomplishes.

There are two bugs in fix link-spouse-to-individual.phtml ("Add a wife using an existing individual"):
The Help Text icon for "Date of marriage" doesn't work.
The Latitude /Longitude icon for "Place of marriage" doesn't work.
This PR fixes both bugs.

PR #3468, opened 2020-08-07
Fix Geographic Data links

ric2016:
Fix error occurring when opening 'Geographic Data' with custom PlaceHierarchyListModule enabled
(discussed in https://www.webtrees.net/index.php/en/forum/help-for-2-0/34847-error-when-clicking-on-geographic-data)

### DIFF
--- a/app/Http/RequestHandlers/MapDataList.php
+++ b/app/Http/RequestHandlers/MapDataList.php
@@ -24,13 +24,13 @@ use Fisharebest\Webtrees\I18N;
 use Fisharebest\Webtrees\Module\PlaceHierarchyListModule;
 use Fisharebest\Webtrees\Services\MapDataService;
 use Fisharebest\Webtrees\Services\ModuleService;
+use Fisharebest\Webtrees\Services\TreeService;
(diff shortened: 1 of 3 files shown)


fisharebest:
As you have discovered, the `listUrl()` function requires a tree object - which is not immediately available.

Instead of `new Tree()`, I think it is better to use `$tree_service->all()->get($tree_name)`.

Also, the logic for generating the URL probably does not belong in the view.

The `Place` object contains the code for generating the URL.

Perhaps the line `$this->map_data_service->activePlaces($parent)` should be modifed to generate `Place` objects?

ric2016:
I have adjusted the code to use TreeService, as suggested.

Regarding: "Also, the logic for generating the URL probably does not belong in the view."

I agree, but this is a more complex change, which I would consider to be independent of the bugfix (the view currently generates the url as well, via `route`, so this isn't worse than before in that regard).

So this should be changed separately if necessary (following 'one change per PR').

fisharebest:
Can you fix the merge conflicts?

ric2016:
Updated version: #3487.

PR #3537, opened 2020-10-10
Fixed bug in the statistics distribution diagram, transferring string…

magicsunday:
… values instead of integers


magicsunday:
Hi @fisharebest, could you please take a look at this small fix?

fisharebest:
Do you think it would be better to modify `function statsPlaces()` to return an `int`?

magicsunday:
I changed the `statsPlaces` method to return integer values.

fisharebest:
Looks good.  Thanks!

Issue #3536, opened 2020-10-08
v2.0.8 Place Hierarchy map problem


ddrury:
If when drilling down a place hierarchy, a child place has the co-ordinates (0,0) then the map rather that displaying the place reverts to the global view.

Example, I have the hierarchy England, Wiltshire, Fovant (Fovant has co-ords 0,0 - don't know how that happened - maybe an import problem?) Click England gives map of England, Click Wiltshire, map reverts to world view. Add proper co-ordinates for Fovant and all is OK

ddrury:
Co-ords 0,0 solved. Webtrees 1.7 sees E0, N0 as valid and exports it as such.

ric2016:
E0, N0 **are** valid coordinates. [There is actually something at that location ;-) ](https://www.ndbc.noaa.gov/station_page.php?station=13010)

I think it is not optimal to use these valid coordinates as a marker to indicate 'coordinates unknown', null would be more appropriate.

ddrury:
Agreed

ddrury:
The problem is in Placelocation::boundingRectangle() line 249. If any child place has co-ords [0,0] then they are returned as the min values. I can see a couple of ways to fix it.

- Amend the filter when getting latitudes & longitudes to remove 'E0', 'W0' etc.
- Filter the results before line 249

Either of these options could potentially unbalance the lat/long arrays (ie if you have a place with co-ords something like [0, 1] . I don't think this would be a problem but leave it up to you to resolve

ddrury:
See new issue https://github.com/fisharebest/webtrees/issues/3638

PR #3551, opened 2020-10-16
Placelocation::boundingRectangle() bug fix

ddrury:
Fixes https://github.com/fisharebest/webtrees/issues/3536

### DIFF
--- a/app/PlaceLocation.php
+++ b/app/PlaceLocation.php
@@ -224,23 +224,21 @@ class PlaceLocation
             ->orWhere('pl_id', '=', $this->id())
             ->groupBy(['pl_lati'])
             ->pluck('pl_lati')
-            ->filter()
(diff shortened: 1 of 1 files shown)


ddrury:
In addition Placelocation::boundingRectangle() line 241 should I think read
`if ($latitudes->isEmpty() || $longitudes->isEmpty()) {` ie **or** instead of **and** as if either is empty then you can't have a valid co-ordinate.

fisharebest:
Not sure I understand this change.

Missing data have empty entries in the wt_placelocation table.  e.g.

Screenshot 2020-12-23 at 18 14 30

We currently filter these.

This change would also filter `N0` and `E0` - which are valid co-ordinates...

ddrury:
See new issue https://github.com/fisharebest/webtrees/issues/3638

PR #3592, opened 2020-11-14
make ClippingsCart more extensible

ric2016:
Make ClippingsCart more extensible by removing references to the static module name. Note that there was already one location in `download.phtml` [(here)](https://github.com/fisharebest/webtrees/blob/510d5393784510a2c189d8fd7cdda66f2530e5a5/resources/views/modules/clippings/download.phtml#L9) where this was the case, i.e. it was handled inconsistently.


fisharebest:
Thanks

PR #3594, opened 2020-11-14
Fix #3484 Display the age of an associated people

### LINKED ISSUE #3484: Display the age of an "associated people"
When associating an external people to an event, Webtrees should also display the age of this people.

Example of situation
The civil register of a death is pointing out the witness, who for example noticed the dead.
The civil register highlighting also the age of this witness.

Displaying the age of the "associate people" could help to see if this age matches with the information indicated in the Register.

Example: Currently for a Dead event we have
**Associate : Paul Klugesherz - Son**

Target 
**Associate : Paul Klugesherz (Aged 27 years) - Son**

ckl67:
Here the simple fix to this feature.
The case has been fixed in only a few lines.. in FunctionsPrinFacts.php

Would be fine to take into account

Thx


ckl67:
Only to take into account fix for 3484. Only modification in 1 file. For the rest I'm still pending your answer/ translation 
So not to take into account

PR #3598, opened 2020-11-18
UK 1939 Register date format change

ddrury:
Update birth date format to mirror transcription and record images on the FindMyPast website

### DIFF
--- a/app/Census/CensusColumnBirthDayMonthYear.php
+++ b/app/Census/CensusColumnBirthDayMonthYear.php
@@ -24,7 +24,7 @@ use Fisharebest\Webtrees\Individual;
 /**
  * The individual's date of birth.
  */
-class CensusColumnBirthDayMonthSlashYear extends AbstractCensusColumn implements CensusColumnInterface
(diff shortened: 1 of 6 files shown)


marku5 (commented):
The date Format should be '%d %M %Y'.

ddrury:
@marku5 

Thanks - fixed

fisharebest:
Here's an example of the 1939 register.

Screenshot 2020-12-14 at 14 20 02

Day numbers do not have leading zeros, so `%j` would appear to be correct.

The slash was to indicate the two columns in one.  Happy to remove the slash (or use two columns).

ddrury:
Actually the more you look the more you realise there is no standard, there are days with & without leading zeros, 2 and 4 digit years, numeric & alpha months etc. You're lucky if its legible! - did you know that the thing was adopted for use in the new NHS and wasn't abandoned until the early '50s. That's why there are scribbled anotations all over the place.

Anyway back to the plot. I agree about the two fields but eventually decided that it would be best to follow the transcription dates which (more or less) follow the Gedcom standard so That's what I ended up with.

PR #3603, opened 2020-11-22
Census assistant occupation

ddrury:
Get occupation based on census date


fisharebest:
Suppose I have this occupation record (from the 1891 census)

```
1 OCCU Coal miner
2 DATE 1891
```

I then add the 1881 census.  This fact would not be considered.  Yet it is the only OCCU record that I have, and is quite probably the correct occupation.

I can understand some sort of logic to prioritise records by date, but not excluding those dated after the census.

PR #3609, opened 2020-11-27
Correction of the census name

mpwt:
Since there have been different censuses in Germany and its predecessor states, those listed here should be clearly assigned. I hope that I have changed all relevant files.

### DIFF
--- a/app/Census/Census.php
+++ b/app/Census/Census.php
@@ -37,7 +37,7 @@ class Census
                     new CensusOfCzechRepublic(),
                     new CensusOfSlovakia(),
                     new CensusOfDenmark(),
-                    new CensusOfDeutschland(),
+                    new CensusOfMecklenburg(),
                     new CensusOfEngland(),
                     new CensusOfFrance(),
                     new CensusOfScotland(),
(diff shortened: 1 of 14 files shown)


ddrury:
Click on details for `Unit tests / Testing PHP 7.3 on ubuntu-latest (pull_request)` and you'll see you've missed a change in CensusOfMecklenburgTest::testPlace (what is returned is not what is expected)

mpwt:
Corrected. Thank you.

fisharebest:
The place name gets stored in the `PLAC` field.

The majority of users store the modern name - not the historic name.

Other censues do the same.  e.g.  we have censuses for Slovakia an Czech Republic.

So, I am not sure this change is correct.

mpwt:
Today Mecklenburg is part of a federal state within the Federal Republic of Germany. In former times it was an independent Grand Duchy within the German Empire.
With the renaming I want to clear up possible misunderstandings among users and on the other hand I want to give the possibility to create further censuses in other federal states. I can't say for sure if there are usable census documents for the whole of Germany.

I consider the place name in the `PLAC` field to be rather informative, since the census was always done at the place of residence of the person or family and in my opinion always has to be adapted by hand.

ddrury:
How does the original documentation classify it

mpwt:
I have now taken a closer look at the selection in the census wizard and the censuses for Mecklenburg-Schwerin are displayed correctly there. 

F9F5C358-0985-46B4-AB98-1A4E4D3CB73C

Sorry, I wanted to make it more clear by the file name. As long as no further censuses in federal states of Germany with the same date / year are to be recorded, the file names can remain the same.

Thanks for your patience.

PR #3627, opened 2020-12-13
Roots filter not working in family datatable

jon48:
The logic for the "roots" filter is not working properly, as it is checking a null on `->childFamilies()` which always returns a `Collection`. We want to check if the collection is empty.


fisharebest:
Thanks!

Issue #3638, opened 2020-12-24
Location [0,0] (Null Island) conundrum


ddrury:
_For Null Island see https://www.geographyrealm.com/zero-degrees-latitude-and-zero-degrees-longitude/_

This addresses what I think is the root cause of issue https://github.com/fisharebest/webtrees/issues/3536 and which I attempted to fix in https://github.com/fisharebest/webtrees/pull/3551 (both now closed)

Map co-ordinates [0,0] are valid but in a webtrees context can _practically_ only be valid at the global level, it is illogical for there to be subordinates to that place or that it can be subordinate to another place.

The fact that webtrees can currently store locations below the global level to [0,0] gives rise to the problem raised in the previous issue. The easiest place to see this effect is in the place hierarchy but may exist in other forms for other maps.

Imagine for eaxample the county of Wiltshire with towns Chippenham, Swindon & Devizes. Wiltshire, Chippenham and Swindon have valid co-ordinates, Devizes has [0,0]. When displaying this hierarchy Placelocation::boundingRectangle() incorrectly takes the [0,0] co-ordinates into account thus displaying a global view.

I think to fix this there are two areas to address:

- What to do with invalid entries already in the placelocation table
- How to stop them getting there in the first place. the geographic import routine sees [0,0] as valid and consequently imports them regardless of hierarchy

As an aside I've discovered a very simple way to inadvertently set co-ordinates to [0,0] in webtrees 1.7 (Googlemaps module - geographic data - edit).  After searching for a place one or a number of yellow markers appear, it is then necessary to click one of these markers and select "use this place" before the co-ordinates are updated. If this isn't done then [0,0] are used when the save button is selected.

**NOTE**
`PedigreeMapModule::getMapData() `line 277 && `PlacesModule::getMapData() `line 175 both ignore co-ordinates [0,0]

PR #3652, opened 2021-01-02
Map coordinate fixes

ddrury:
Fixes https://github.com/fisharebest/webtrees/issues/3638 and standardises co-ordinate formats.

- Use null to indicate an invalid coordinate rather than 0 (which is a valid coordinate)
- Use null rather than empty string for latitude & longitude when importing places from genealogy data
- Add field validation to Control Panel|Geographic Data|Add/Edit page so that location [0,0] is only valid at the global level
- Tighten up import/export routines to ensure invalid data is excluded and data written to the database has a consistent format.
- Fixed bug in `GedcomService::writeDegrees();` - Incorrect format definition
- Bump existing (but not applied) Migration44.php to Migration45.php to enable creation of new Migration44.php which removes invalid co-ordinates from database. **IMPORTANT** SCHEMA_VERSION in app/webtrees.php has not been changed so this will need to be applied separately.

NB I think application of this PR should make changes to allow Migration45 to be applied relatively simple.

### DIFF
--- a/app/Fact.php
+++ b/app/Fact.php
@@ -273,9 +273,9 @@ class Fact
     /**
      * Get the PLAC:MAP:LATI for the fact.
      *
-     * @return float
+     * @return float|null
      */
-    public function latitude(): float
+    public function latitude()
     {
         if (preg_match('/\n4 LATI (.+)/', $this->gedcom, $match)) {
             $gedcom_service = new GedcomService();
@@ -283,15 +283,15 @@ class Fact
             return $gedcom_service->readLatitude($match[1]);
         }
 
-        return 0.0;
+        return null;
     }
 
     /**
      * Get the PLAC:MAP:LONG for the fact.
      *
-     * @return float
+     * @return float|null
      */
-    public function longitude(): float
+    public function longitude()
     {
         if (preg_match('/\n4 LONG (.+)/', $this->gedcom, $match)) {
             $gedcom_service = new GedcomService();
@@ -299,7 +299,7 @@ class Fact
             return $gedcom_service->readLongitude($match[1]);
         }
 
-        return 0.0;
+        return null;
     }
 
     /**
--- a/app/Functions/FunctionsImport.php
+++ b/app/Functions/FunctionsImport.php
@@ -491,7 +491,7 @@ class FunctionsImport
 
         $location = new PlaceLocation($place_name);
 
-        if ($location->latitude() === 0.0 && $location->longitude() === 0.0) {
+        if ($location->latitude() === null || $location->longitude() === null) {
             DB::table('placelocation')
                 ->where('pl_id', '=', $location->id())
                 ->update([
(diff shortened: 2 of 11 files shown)


fisharebest:
Is there a reason to use `999.999` instead of `null`?

ddrury:
Not necessarily 999.999 but it's being compared to a float. If null
doesn't  throw a type mismatch then let's use that

ddrury:
I suppose it would be cleaner to use null and change return types etc. from float to float|null. Would you prefer that?

ddrury:
Changed from 999.999 to null

ddrury:
The github diff of LocationController makes things look a lot more complex than they really are. For example function exportLocations had a large chunk of code only used when exporting a csv so moving that chunk to function exportCSV simplified functions exportLocations and exportGeoJSON but the diff around this area looks really messy

fisharebest:
> PR is 3652 👎 and I've finished making changes now 👍

When you posted this on 16 Jan, I took a local copy of the PR, and have been merging your changes into my own local changes.

You have force-pushed a few times since then.   I can't tell if you simply rebased your PR over the latest code - or whether you made additional changes and merged your commits.

If the former, then no problem.
If the latter, then I don't have an easy way to see what changes you made.

ddrury:
Ah,

>  I can't tell if you simply rebased your PR over the latest code - or whether you made additional changes and merged your commits.

I've been overwriting the commit each time and force pushing. Can you let me have a copy of LocationController.php from the 16th and I'll work out the differences and create a new pull request

ddrury:
Don't know whether you can work from this:
Changes since 16th to LocationController
Added new constant 
```
//Used when exporting csv file
private const FIELD_DELIMITER = ';';
```
Re-organised functions `exportLocations, buildExport, exportCSV and exportGeoJSON`

changes line in  `importLocationsAction `to` while (($row = fgetcsv($fp, 0, self::FIELD_DELIMITER)) !== false) {` ie change semi-colon to constant

fisharebest:
I've merged many of your changes.

Some I didn't agree with (see my review comments).

ddrury:
Morning Greg,

Some I didn't agree with (see my review comments).

Can you point me to your comments - I can't find them

Regards
David

fisharebest:
There was the restriction that [0,0] could not have child places.

If [0,0] is "At Sea", and someone died at "RMS Titanic, At Sea", then this place *would* have co-ordinates.

fisharebest:
...actually - just scroll up.  Comments are shown on this page.

ddrury:
Afraid I disagree here. In your example we know where the Titanic sank so we know which ocean she was in therefore the entry should be: RMS Titanic, Atalantic Ocean with Atlantic Ocean and Titanic having appropriate and known coordinates.

I cannot see where mapping a place that is in effect unknown to [0,0] brings any benefit to the table.

Still you're the boss and I accept it's a once in a blue moon occurrence

PR #3661, opened 2021-01-07
Update ModuleCustomTrait.php

magicsunday:
Adding an additional method call to allow the processing of the response from the get latest version request. With this it would be possible for me to query the version information directly from githubs release json without the need for an exta latest-release.txt file. Using the method I could easily parse the JSON response and extract the version information.

### DIFF
--- a/app/Module/ModuleCustomTrait.php
+++ b/app/Module/ModuleCustomTrait.php
@@ -83,6 +83,18 @@ trait ModuleCustomTrait
     }
 
     /**
+     * Fetch the version information from the given content string.
+     * 
+     * @param string $content
+     *
+     * @return string
+     */
+    protected function extractVersion(string $content): string
+    {
+        return $content;
+    }
+
+    /**
      * Fetch the latest version of this module.
      *
      * @return string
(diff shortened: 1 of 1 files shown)


magicsunday:
Do you got some time to look at this?

ric2016:
I don't think this is necessary - why don't you just adjust `customModuleLatestVersion()` in your custom module, according to your specific needs? (Adjustments of that method are required anyway in case of a different versioning scheme, and/or a different cache expiration period, or should we have separate methods for these as well? I don't think so.)

magicsunday:
Because of duplicate code. You should avoid it where ever it is possible. As "customModuleLatestVersion" is a method from the core there is always a possibility of change. Maybe it contains a bug, so it would require me to fix this in my module too but my module shouldn't know about the core. So intercepting the core code using a callback or hook would be the better solution instead of duplicating it. Currently the method only accepts a single solution of querying module information without greatly modifying it.

06Games (approved):

arbor95:
Seems: will not be implemented->close

ekdahl:
Adding this would be great.

PR #3665, opened 2021-01-11
Update Mime.php

magicsunday:
Fixed mime type of SVG images => see https://www.w3.org/TR/SVG11/intro.html#MIMEType

With this fix SVG images could be used within asset URLs.


fisharebest:
Thanks

PR #3702, opened 2021-02-01
Datatables pagination

ddrury:
I know you don't like this but...

By changing the first/last/next/previous paging text to symbols and rearranging the controls slightly, the space saved allows the provision of the pagination & display controls on both header & footer of a list without increasing the overall table size. This is useful when viewing a long table (for example the media list on the control panel)

demo

### DIFF
--- a/resources/js/vendor.js
+++ b/resources/js/vendor.js
@@ -28,12 +28,12 @@ import {
 } from '@fortawesome/free-regular-svg-icons';
 import {
   // For resources/views/icons/*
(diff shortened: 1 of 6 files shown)


reteP-riS:
I support this idea, but it seems - at least from above screenshot - that the "Showing x to y of z" records information is missing. In the original lists this information is in the lower left corner of the table.

ddrury:
@reteP-riS 

> "Showing x to y of z" 

Bottom centre

reteP-riS:
I withdraw my comment and state the opposite.... Thanks!

PR #3706, opened 2021-02-03
Use Nominatim and Openrouteservice for place autocomplete

ddrury:
This replaces https://github.com/fisharebest/webtrees/pull/3120. OpenCage has been abandoned in favour of Openrouteservice as the free (trial) account of OpenCage is silently deleted after 6 months of inactivity and thereafter returns a 401 Unauthorized response. Openrouteservice is completely free (but to use it a key must be acquired). It allows 1000 geocode requests per day and uses pretty much the same opensource databases as OpenCage.

Openrouteservice (https://openrouteservice.org/) is built upon the Open Source Pelias search engine (https://github.com/pelias/pelias)

If no key is provided for Openrouteservice then Nominatim is used instead.

Both place autocomplete and the Geographic data add/edit search facility have been modified to use the above mentioned tools.

NB I've provided a migration to remove the old geonames key but haven't updated webtrees.php

### DIFF
--- a/app/Http/RequestHandlers/AutoCompletePlace.php
+++ b/app/Http/RequestHandlers/AutoCompletePlace.php
@@ -19,8 +19,11 @@ declare(strict_types=1);
 
 namespace Fisharebest\Webtrees\Http\RequestHandlers;
 
+use Exception;
+use Fisharebest\Webtrees\Gedcom;
 use Fisharebest\Webtrees\I18N;
 use Fisharebest\Webtrees\Place;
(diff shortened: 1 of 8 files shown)


fisharebest:
The terms and conditions for nominatim say that it should not be used for autocomplete.

https://operations.osmfoundation.org/policies/nominatim/

I have added OpenRouteService support to my dev code.  I'll push this to `main` in the next few days.

fisharebest:
I have merged this, and turned it into a module.

Some furthing polish may be needed.

PR #3710, opened 2021-02-05
Remove dead code

ddrury:
As a result of commits https://github.com/fisharebest/webtrees/commit/90949315619051b772509f417394730bd8b1bda0 & https://github.com/fisharebest/webtrees/commit/a45ad69d9fe14e475a431f9eb0e7184dab266144

### DIFF
--- a/resources/views/admin/location-edit.phtml
+++ b/resources/views/admin/location-edit.phtml
@@ -93,13 +93,11 @@ use Fisharebest\Webtrees\View;
     'use strict';
 
     window.WT_OSM_ADMIN = (function () {
-        const minZoom = 2;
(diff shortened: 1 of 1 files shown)


fisharebest:
I don't see the need for `$('.editable').trigger('change');`.

I can't see anything not working if I remove it...

ddrury:
Although we've removed the ability to store the zoom level, we can still 
zoom the map. (I sometimes find it useful to fine-tune the location) and 
sometimes the marker disappears of the map after zooming. The 
$('.editable').trigger('change');. re-centre's the marker on the map at 
zoom end.

Regards
David

fisharebest:
OK - thanks!

PR #3738, opened 2021-02-20
Fix wrong columns used in families query of ModuleDataFixTrait

jon48:
Add basic unit testing for Data fix handler and FixDuplicateLinks


fisharebest:
Thanks!

PR #3741, opened 2021-02-23
Add fact - census - modify census selector

ddrury:
use option groups


fisharebest:
`$census_place->allCensusDates()[0]->censusPlace()` would be better as `$census_place->censusPlace()`

Some of the censuses are just for a part of a country.  e.g.

Screenshot 2021-02-23 at 13 22 21

This info is now missing.  Either add it back in for every record, or just those where it differs from the parent country?  e.g. 

```
                  <?php if ($census_place->censusPlace() !== $census->censusPlace()): ?>
                    — <?= $census->censusPlace() ?>
                  <?php endif ?>
```

Screenshot 2021-02-23 at 13 40 55

ddrury:
Done

fisharebest:
Thanks!

Issue #3563, opened 2020-10-19
Reports - Family - PDF: missing INDI information (e.g. wifes name)


ungeahnt:
**HTML view is ok:**
![grafik](https://user-images.githubusercontent.com/28572767/96488256-fb19a900-123d-11eb-8b9e-663a4fcf82db.png)

 
 
**PDF view/download with missing data:** 
![grafik](https://user-images.githubusercontent.com/28572767/96488109-cd346480-123d-11eb-895a-7b1a1968d472.png)

I also see such behavior in my data with different families. Sometimes both (husband and wife) are missing.
https://dev.webtrees.net/demo-stable/index.php?route=%2Fdemo-stable%2Ftree%2Fdemo%2Freport%2Ffamily_group_report&xref=f1

ekiczek:
To pinpoint the issue, I only see this happening when photo display within the report is enabled. If I disable the photos, the disappearing blocks are present.

PR #3757, opened 2021-03-09
Fixing #3563.

ekiczek:
This fixes the family report PDF issue #3563.


fisharebest:
Thanks for the fix.

I looked, but couldn't find the problem...

PR #3787, opened 2021-03-17
Fix #3784. Suppresses GEDCOM NOTE facts from Family Report for all fa…

### LINKED ISSUE #3784: 2.0.13 - Family Report prints raw shared note number even with disabled notes
In some families I have shared notes.

When I print a **Family Report** and disable the option **Show notes**, it still prints the raw shared note number (like @N20474@) even if the Show Notes option is disabled.

image

ekiczek:
…mily members if "Show notes" is unchecked.

With "Show notes" checked:
Screen Shot 2021-03-16 at 8 33 22 PM

With "Show notes" unchecked:
Screen Shot 2021-03-16 at 8 33 41 PM

### DIFF
--- a/resources/xml/reports/family_group_report.xml
+++ b/resources/xml/reports/family_group_report.xml
@@ -36,7 +36,7 @@
 	<!-- Date -->
 	<Style name="date" font="$fonts" size="8"/>
 
-	<SetVar name="ignore" value="CHAN,NAME,SEX,SOUR,OBJE,RESN,FAMC,FAMS,TITL,CHIL,HUSB,WIFE,MARR,BIRT,CHR,BAPM,DEAT,CREM,BURI,_UID,_WT_OBJE_SORT" />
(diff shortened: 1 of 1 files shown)


ekiczek:
P.S. - I was hoping there would be a more elegant way to do this. If you know of a better way, I'm happy to modify the PR, just let me know.

fisharebest:
> I was hoping there would be a more elegant way to do this. If you know of a better way, I'm happy to modify the PR, just let me know.

I think that all we need to do is add `NOTE` to the definition for the variable `ignore`.

ekiczek:
> I think that all we need to do is add `NOTE` to the definition for the variable `ignore`.
> 
> https://github.com/fisharebest/webtrees/blob/2.0.13/resources/xml/reports/family_group_report.xml#L39

Thanks for the tip! Should be all set now.

fisharebest:
Thanks!

PR #3803, opened 2021-03-24
Add ability to select other map providers

ddrury:
Provides the ability to select alternative map services on the fly. By default only OpenStreetMap is provided, however files for Google, Bing, Mapbox, Here, Esri/Arcgis and Ordnance survey historic map of Great Britain are available. They are small (1-2k) json files that can be imported via the admin page.
**NOTE** the provider files are not included in this PR - should they be put somewhere in webtrees.net addons? - they can be imported from server/data/map-providers so could go there. Will be attached to this conversation in a zip file

To accomodate this PR a few things had to be moved.

- The option to enable/disable the use of a map in the place hierarchy module has been moved to a config option within the module itself.
- The old map provider class has been renamed to 'Gazetteer' (and re-titled "Place autocomplete") - new item on the control panel in the Maps section - Maps isn't a particularly good name now
- The search control on the add/edit map in Control panel - geographic data, has moved from top right to bottom left to accomodate the new provider control

Migration45 creates two tables, populates them with data for OpenStreetMap and sets the place hierarchy maps option based on the old setting
- map_names - names of providers & styles (eg OpenStreetMap with Deutsch, Francais & Mapnik)
- map_parameters - parameters associated with each provider/style

**Note** all values in the map_parameters table are serialized on insertion and unserialized when read, so it's unadvisable to play directly with them. (it was the easiest way to handle definitions that have parameters that are arrays - I looked at JSON fields but decided serialization was simpler)

Service MapProviderService & requesthandlers MapProviderPage, MapProviderAction, MapProviderImportPage, MapProviderImportAction, MapProviderDelete and MapProviderUpdateStyles along with associated views have been created.

Administration is managed from the control panel option 'Map providers'.

maps have a control (top right) for selection of providers/styles.

npm install and npm run production are necessary to load new control etc.


ddrury:
Additional map provider files
[providers.zip](https://github.com/fisharebest/webtrees/files/6198236/providers.zip)

tronsmit:
Really nice to be able to switch back to other providers. Google Map access tested. Looking forward to get this into the main branch at some time.

fisharebest:
> Looking forward to get this into the main branch at some time.

I have already started to merge this code.

jon48:
Very nice idea to be able to switch provider.

Having had a similar concern in one module, I agree that serialising unknown structure in one column is a practical way to store configuration, however I would strongly advise against the use of `serialize`\`unserialize` for that matter: `unserialize` comes with serious security concerns if not handled with care (even [the documentation](https://www.php.net/manual/en/function.unserialize.php) comes with a warning). When storing in the database, we should at least ensure that the content had not been tempered with, with a signature. Given that the parameters seem to have simple data type (string, int, array...), they should be easily JSON serialisable, and `json_decode` is much safer. You could even simplify the logic a bit serialising the whole parameters object without having to split the parameter_name and parameter_values.

More globally, may I ask why you did not go with a module approach? I.e. the PR would be adding the central management of map providers, but then each map provider would be a module, with a specific interface (`MapProviderInterface` for instance - I assume that what you have implemented as styles would be just different modules potentially extending each other), and a method would provide the provider & default parameters.
That would alleviate partly the need to store the provider parameters in the database. That would allow as well specific map providers to deal with their own site or user-specific setting (I am thinking the GoogleMaps API key), as I am not sure how that would be handled with the import mechanism.

fisharebest:
> may I ask why you did not go with a module approach

FYI, I am currently re-writing this PR to use modules.

I am adding two interfaces - one for map provider and one for geo-lookup.

jon48:
> FYI, I am currently re-writing this PR to use modules.

:+1:

ddrury:
Ha, I originally wrote this using modules but thought it was overkill - ho hum

ddrury:
> I am adding two interfaces - one for map provider and one for geo-lookup

Re geo-lookup you may find my other PR https://github.com/fisharebest/webtrees/pull/3706 useful

fisharebest:
I have merged this, and turned it into modules.

Some furthing polish may be needed.  In particular, I want to turn the geo-lookup function into  modules.

PR #3820, opened 2021-04-02
Initialised by

tronsmit:
1. importing dependencies
2. allowing mod rewrites


fisharebest:
Site owners may include a `.htaccess` file which contains more than just URL rewriting.
For example, HTTP authentication.

So, we cannot include a `.htaccess` file in our releases, as it would overwrite the existing one.

Including these files also goes against the best-practices recommended in [The twelve factor app](https://12factor.net/).

tronsmit:
Of course, never intended this to be here.

Why are .htaccess and package-lock.json not listed in .gitignore?

I had spent the most part of Friday setting up an ubuntu VM and finally (after about 10 hours) got the local webtrees running. Checking git, these were the only two files with changes (.htaccess new and package-lock.json modified). Since my plan is to rebase upstream main onto my fork, and keep all local changes in a sandbox (keeping it on top of main), I needed to either stash the local changes or commit them (help me out if there is a different workflow here).

(Once pushed to my fork on github - and being late at night - and being unfamiliar with this - I clicked on that tantalising "Create pull request" button thinking it somehow was a necessary step on my fork. We learn... slowly)

It would be nice if none of the shipped files (out of the box), and the minimum required to run a working site from the cloned directory, did not show up as changes in git. (For me, this was the two files mentioned above)

Issue #3814, opened 2021-03-30
2.0.14 Source citation link


jpretired:
The issue https://github.com/fisharebest/webtrees/issues/3763 is fixed for 2.0.14 but now it works so that the whole value of Citation details is converted to a link.
In my tree the Citation details' value is like this:
3 PAGE snímek http://digi.archives.cz/da/permalink?xid=be96ed78-f13c-102f-8255-0050568c0263&scan=251
where "snímek" means "scan".
The link href="snímek http://digi.archives.cz/da/permalink?xid=be96ed78-f13c-102f-8255-0050568c0263&scan=251" is of course invalid.
In earlier wt versions back to 2.0.11 and before only the "http://..." part became a link.
Please can this issue be reverted to the way it used to function always before?
In my tree I have several thousands of Citation details like that mentioned above.

tronsmit:
Yes, my PAGE are of the same format with the url often embedded in the middle of the text - I append the url's with an access date.

For example:

Death notice: Johanna Sophia BURGERS (1877), MOOC 6/9/162 #4845/1877, (https://familysearch.org/ark:/61903/3:1:3Q9M-CSQN-X4DZ?cc=2517051&wc=WV44-SZY%3A1598082381 : 12 September 2017), 007844948 > image 1945 of 2181; Pietermaritzburg Archives (Formerly Natal State Archives), South Africa : accessed 3 Nov 2018

I also see that following this link, leads to a page with URL that is not helpful (I presume also not for search engines)

https://genes.tronsmit.com/tree/Theron#4845/1877,%20(https://familysearch.org/ark:/61903/3:1:3Q9M-CSQN-X4DZ?cc=2517051&wc=WV44-SZY%3A1598082381%20:%2012%20September%202017),%20007844948%20%3E%20image%201945%20of%202181;%20Pietermaritzburg%20Archives%20(Formerly%20Natal%20State%20Archives),%20South%20Africa%20:%20accessed%203%20Nov%202018

jpretired:
A humble question - is there a plan to fix this issue for wt 2.1?
Thank you.

PR #3826, opened 2021-04-04
only url in SOUR:PAGE becomes a link

tronsmit:
interim solution for https://github.com/fisharebest/webtrees/issues/3814


tronsmit:
Not ready yet - e(..) breaks links with parameters passed by &, for example

PR #3855, opened 2021-04-20
Fix Add child to individual: FAMC XREF wrongly set to the parent individual XREF

jon48:
After having created the new one-parent family, and when creating the GEDCOM for the new child,  the FAMC record uses the parent XREF instead of the newly created XREF, thus creating an invalid gedcom


fisharebest:
Thanks!

PR #3858, opened 2021-04-22
Relationships : Bug  + French improvements + Finer pedigree linkage methods

jon48:
I know I should not really put several commits in a PR, but there are the result of the same work, just wanted to functionally separate them.

First, the French translation has been modified based on the first feedbacks on the forum.

A small bug was in `nameFromPath` when no relationships are found: `array_reduce` would then return the initial array, which did not have a 0 index.

Finally, I have added a couple of methods linked to the pedigree linkage types: now it is possible to define more precisely based on the different values of `\Fisharebest\Webtrees\Elements\PedigreeLinkageType` (with the exception of 'sealing' which is an obscure concept to me). This for instance allows to define more precisely half-brother where (in French at least) a blood-link is required. 'birth' is assumed though when no PEDI value exist.
I had some difficulties finding a good name for the `birth` linkage methods though, as the term I could think of (biological, natural,...) would be the same thing in both directions, so do not offer the dichotomy as adoptive/adopted, fostering/fostered. Feel free to modify.

### DIFF
--- a/app/Module/LanguageEnglishUnitedStates.php
+++ b/app/Module/LanguageEnglishUnitedStates.php
@@ -142,9 +142,15 @@ class LanguageEnglishUnitedStates extends AbstractModule implements ModuleLangua
             Relationship::fixed('foster-mother', 'foster-mother’s %s')->fostering()->mother(),
             Relationship::fixed('foster-father', 'foster-father’s %s')->fostering()->father(),
             Relationship::fixed('foster-parent', 'foster-parent’s %s')->fostering()->parent(),
+            Relationship::fixed('foster-sister', 'foster-sister’s %s')->fostering()->sister(),
+            Relationship::fixed('foster-brother', 'foster-brother’s %s')->fostering()->brother(),
+            Relationship::fixed('foster-sibling', 'foster-sibling’s %s')->fostering()->sibling(),
             Relationship::fixed('foster-daughter', 'foster-daughter’s %s')->fostered()->daughter(),
             Relationship::fixed('foster-son', 'foster-son’s %s')->fostered()->son(),
             Relationship::fixed('foster-child', 'foster-child’s %s')->fostered()->child(),
+            Relationship::fixed('foster-sister', 'foster-sister’s %s')->fostered()->sister(),
+            Relationship::fixed('foster-brother', 'foster-brother’s %s')->fostered()->brother(),
+            Relationship::fixed('foster-sibling', 'foster-sibling’s %s')->fostered()->sibling(),
             // Parents
             Relationship::fixed('mother', 'mother’s %s')->mother(),
             Relationship::fixed('father', 'father’s %s')->father(),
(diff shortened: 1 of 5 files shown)


fisharebest:
> A small bug was in nameFromPath when no relationships are found: array_reduce would then return the initial array, which did not have a 0 index.

I originally had a test for this - but the only time I could trigger it was when there was an error in the path (`$nodes` was invalid) or the relationship definitions contained an error.

----------

If multiple rules match, we take the first one.  We have rules for `->male()` and `->female()` but not for `->unknownSex()`.  This last one isn't needed.  For example:

```
Relationship::fixed('sister', 'sister’s %s')->sibling()->female(),
Relationship::fixed('brother', 'brother’s %s')->sibling()->male(),
Relationship::fixed('sibling', 'sibling’s %s')->sibling(), // can only match unknown sex.
```

If a language uses the same name for male and unknown, then we can simplify this to

```
Relationship::fixed('xxx', 'xxx’s %s')->sibling()->female(),
Relationship::fixed('yyy', 'yyy’s %s')->sibling(), // matches both male and unknown sex.
```

For this reason, I do not think we need the `->biologicallyBorn()` rules.
We just need to define the relationships in the correct order.

jon48:
> I originally had a test for this - but the only time I could trigger it was when there was an error in the path (`$nodes` was invalid) or the relationship definitions contained an error.

To be honest, I come up to it the same way as you did: with an error when writing my `$nodes` array. But looking a bit more at the code, you can trigger it with an empty array as input (which is a valid input to the public method). For instance: 
```php
self::assertRelationship('', [], $en);
```
Even within the `nameFromPath` method itself, there is an unsafe sequence:
```php
$relationships = $this->matchRelationships($nodes, $pattern, $relationships);
```
For which `$relationships = []` is a valid output (the default one for `matchRelationships` actually), and then, without any test:
```php
return array_reduce($relationships, static function (array $carry, array $item): array {
            return [sprintf($carry[1], $item[0]), sprintf($carry[1], $item[1])];
        }, [1 => '%s'])[0];
```
Which will raise the exception with `$relationships = []` 

In the UI, all code paths are protected by a check on the `$nodes` variable before calling  `nameFromPath`, so we cannot trigger it through the core code (for isolated persons for instance), however, as a publicly available method of `RelationshipService`, I feel it is its own responsibility to make sure it can cope with the empty input (and just adding the default value for the 0 index or testing it before `array_reduce` is  not a great cost). Additionnally, it seems to make phpstan happier 😄

jon48:
> If multiple rules match, we take the first one. We have rules for `->male()` and `->female()` but not for `->unknownSex()`. This last one isn't needed. For example:

> For this reason, I do not think we need the `->biologicallyBorn()` rules.
> We just need to define the relationships in the correct order.

I agree this is working fine with sex or marital status, which are nearly-binaries values, and I have made use of it several times in the French translation, but for pedigree linkages, webtrees support 5 possible values (birth|adopted|foster|sealing|rada), and from the experience of working on the French translation, it makes it a bit awkward to translate a specific case by exhaustion of all the other cases (as this is a dimension orthogonal to sex, it kind of multiply the cases). I had a lot of `half-sibling` coming up in situations where it is not appropriate at all, and that had me scratching my head.

My intention with the introduction of the `biological`/`biologicallyBorn` methods was to have a more positively defined relationship, where `birth` linkage is a **required** condition for the relationship, independently of relying on translations being available for all other pedigree linkage cases.

I do not expect it to be used in 95% of the cases, but this can be practical in specific cases, for a fairly limited cost (it is just a couple of very short methods reusing a common logic)
To come back to the `half-sibling` example: if you take the latest tree in the PR (not the rest of the code), and run `$service->nameFromPath([$i40mr, $f18, $i32f], $en)` (so a wet-nursed child vs the natural child of the mother), it will return `half-sister`, which I believe will not make much sense even in English (it does not in French, we apply the term of `half-sister`/`demi-sœur` **only** when there is a blood link between the two siblings - through one commun parent; it is not a fallback term when the pedigree linkage could not be translated), because the definition would make it the fallback one when talking about non-standard sibling, and the only way to ensure `half-sibling` is not used inappropriately is indeed to define all other possible PEDI values.
Again, it can be done, but I think it is much more clearer and safer when defining it explicitly in the actual definition.

fisharebest:
I have cherry-picked commit 2 (`array_reduce`)

fisharebest:
I would like to cherry-pick/merge commit "French Relationships : improvements based on forum feedback ", but it is conflicted.  Can you rebase/fix this?

fisharebest:
Commit "Bug + French improvements + Finer pedigree linkage methods":

* I don't want to add rada relationships at this stage.
* I don't fully understand the need for the "biological" relationship.

jon48:
I am closing this PR, following the cherry-picking of the small bug.

The "French Improvement" part was already split into #3916, and are now merged.

For the pedigree linkage methods, given that you have made clear in one of our previous discussion that we should not extend beyond the standard values (`birth`, `foster`, `adopted`), my remarks are less relevant, and as long as translations ensure that the foster and adopted cases are covered, then the risk of falling back to the "half-brother" description for non-birth linkage is remediated.

PR #3897, opened 2021-05-17
Adding 3 Canadian Census years

glarwill:
Uncertain about my ordering of the various censuses based upon locale.


fisharebest:
Thanks!

PR #3905, opened 2021-05-24
Fix imported boolean column for postgresql

AsamK:
When setting up webtrees with a PostgreSQL database (tested with 13.2), showing the manage trees page and importing gedcom files doesn't work.
It shows the following error:
```
SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: boolean = integer
LINE 1: ...dcom_chunk" where "gedcom_id" = $1 and "imported" = $2) as "...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts. (SQL: select exists(select * from "wt_gedcom_chunk" where "gedcom_id" = 1 and "imported" = 0) as "exists") …/vendor/illuminate/database/Connection.php:664
#0 …/vendor/illuminate/database/Connection.php(624): Illuminate\Database\Connection->runQueryCallback()
#1 …/vendor/illuminate/database/Connection.php(333): Illuminate\Database\Connection->run()
#2 …/vendor/illuminate/database/Query/Builder.php(2426): Illuminate\Database\Connection->select()
#3 …/resources/views/admin/trees.phtml(66): Illuminate\Database\Query\Builder->exists()
```

PostgreSQL seems to be more strict for boolean columns.
It only accepts string representations, but not a numeric representation: https://www.postgresql.org/docs/13/datatype-boolean.html
This PR replaces `0` with `'0'` for the `imported` column, as it is already done in the other places.

### DIFF
--- a/app/Http/RequestHandlers/GedcomLoad.php
+++ b/app/Http/RequestHandlers/GedcomLoad.php
@@ -129,7 +129,7 @@ class GedcomLoad implements RequestHandlerInterface
                 $n = DB::table('gedcom_chunk')
                     ->where('gedcom_chunk_id', '=', $data->gedcom_chunk_id)
                     ->where('imported', '=', '0')
-                    ->update(['imported' => 1]);
+                    ->update(['imported' => '1']);
 
                 // Another process has already imported this data?
                 if ($n === 0) {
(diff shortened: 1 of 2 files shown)


fisharebest:
SInce this column is defined as a boolean type, would it be better if we used the PHP values `TRUE` and `FALSE` instead of `0`, `'0'`, `1` and `'1'`?

I tried it with MySQL and SQLite - both work OK.

AsamK:
I agree using a PHP boolean should be the preferred way.

Unfortunately I can't confirm that it works, because of a bug in PHP 7.4.18 (which is currently installed on my system) ...
https://github.com/laravel/framework/issues/37215
https://github.com/laravel/framework/issues/37179

If I read those issues correctly, then using numbers should also work again with PHP 7.4.19 ... due to a rather strange PHP pgo_pgsql behavior: https://bugs.php.net/bug.php?id=80892

AsamK:
I've switch to mysql, so closing this PR.

PR #3911, opened 2021-05-28
Fix: invalid source XREF sent for SOUR:PAGE autocompletion

jon48:
I have not looked into the details, but it seems that following the new tag management implementation, the value in the XREF select2 dropdown contains the @ characters (whereas only the Xref value was present previously).

As a consequence, the full @XREF@ value is sent in the `extra` parameter of the autocompletion URL, whereas the autocomplete request handler expect a valid source XREF (already trimmed of the @ characters).

There are at least 2 approaches to fix: either trim before sending the request, or trim it in the autocomplete Request Handler. I have taken the former approach in the PR, but I have no preference if you prefer the latter.


fisharebest:
> the value in the XREF select2 dropdown contains the @ characters 

This was a delibarate change.  Previously, we had a hidden field to indicate whether an edit field was a link (and needed @ adding) or a value.  This caused problems with fields that could be both (NOTE, SOUR), and so it was made consistent that the edit field would contain the actual data (including @).

> I have taken the former approach in the PR

I think I would have done the same.

PR #3916, opened 2021-05-31
French Relationships : improvements based on forum feedback

jon48:
This is a split (and rebase) of PR #3858, containing only the improvements of the French relationship translations, without the changes to the Relationship class.
That will allow at least to integrate that non-controversial part of the initial PR in 2.1, and once merged, I will update #3858 so that it will contain only the proposed bugfix and changes related to pedigree-linkage, open for further discussion and improvements.

French improvement include:
+ Change the logic for "arrière-" prefix
+ 'gendre/bru' rather than 'beau-fils/belle-fille'
+ use simple cousin method
+ Use reflective relation tests
+ Adoptive/fostered siblings

### DIFF
--- a/app/Module/LanguageFrench.php
+++ b/app/Module/LanguageFrench.php
@@ -45,14 +45,14 @@ class LanguageFrench extends AbstractModule implements ModuleLanguageInterface
 
     protected const ASYMMETRIC_COUSINS = [
         1 => [
-            'F' => ['down', 'petite-', 'cousine', 'de la ', 'de la '],
-            'M' => ['down', 'petit-', 'cousin', 'du ', 'du '],
-            'U' => ['down', 'petit-', 'cousin', 'du ', 'du ']
+            'F' => ['down', 'petite-cousine', 'de la '],
+            'M' => ['down', 'petit-cousin', 'du '],
+            'U' => ['down', 'petit-cousin', 'du ']
         ],
         -1 => [
-            'F' => ['up', 'grand-', 'cousine', 'de la ', 'de la '],
-            'M' => ['up', 'grand-', 'cousin', 'du ', 'du '],
-            'U' => ['up', 'grand-', 'cousin', 'du ', 'du ']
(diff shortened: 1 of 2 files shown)


app/Module/LanguageFrench.php
```diff
@@ -183,60 +207,71 @@ static function (int $up, int $down, string $sex) use ($symmetricCousin, $compou
             Relationship::fixed('beau-père', '%s du beau-père')->parent()->husband(),
             Relationship::fixed('beau-parent', '%s du beau-parent')->parent()->married()->spouse(),
             Relationship::fixed('belle-fille', '%s de la belle-fille')->married()->spouse()->daughter(),
-            Relationship::fixed('beau-fils', '%s du beau-fils')->married()->spouse()->son(),
+            Relationship::fixed('gendre', '%s du beau-fils')->married()->spouse()->son(),
```
fisharebest, line 210: Is this change right?  It is inconsistent with the lines around it?

jon48: Thanks for spotting that, this is a mistake indeed. The line should not have been changed at all (there is a change further down from beau-fils to gendre, but that applies only to the in-laws context).
I have resubmitted a corrected file.

fisharebest:
Hi Jon - sorry for the delay in responding.

I am not 100% certain that we should include the definitions for "rada".  This is very much a non-standard pedigree/relationship.  It was added at the request of our arabic translator, and I do not know how it is used - (e.g. is it a one-parent family with the mother, or is the father also a part of the family).  Maybe an `ASSO` would be a better way to represent this relationship?

If we have it for one language, then we ought to have it for all.

Can you remove the rada part.  It needs more thought, and so I don't want to merge this while it is still there.

jon48:
No worries. I assume you meant from the feature test data, as I already removed it from the French language file. 

I pushed the corrected (and rebased) version.

fisharebest:
Thanks!

PR #3924, opened 2021-06-13
DRAFT: New and updated historical core-modules

Lars1963:
**WIP**

**NEW**
- module `HistoryChancellorsGerman`
- module `HistoryPresidentsFrench`
- module `HistoryMonarchsFrench`

**DELETED**
- module `FrenchHistory` in favor of new enhanced french modules

**added:**
-   * missing entries
-   * localized wikipedia sources

**fixed:**
-   * typos in dates
-   * typos in informatioan

**enhanced:**
-   more facts in most modules
-   use datetype JULIAN in dates prior to 15th october 1582
-   rename the modules for consistent naming and better overview in folder

**WIP as follow up to this pr:**
- [ ] new module `HistoryFactsFrench`
- [x] new module `HistoryPresidentsGerman`
- [ ] new module `HistoryMilitaryFactsWorld`
- [ ] new module `HistoryPandemicsWorld`
- [ ] new Module `HistoryMonarchsBelgium`

PR #3930, opened 2021-06-20
NEW module HistoryChancellorsGerman.php

Lars1963:
Also based on the work of and with thanks to Hermann Hartenthaler

- First effort to make a translatable history module:
- First Part of #3924

@fisharebest Greg, please review. Will the translation be working the way I did it? If no, please point out to me, what has to be changed. THX

If it works this way, I'll make the PR's for other almost ready Modules

- [ ] https://github.com/fisharebest/webtrees/blob/main/app/Services/ModuleService.php must also be adapted


Lars1963:
Fetch upstream

Lars1963:
closed for now. do some more cleanaup and testing

PR #3935, opened 2021-06-21
Can census

glarwill:
Added two more years of Canadian Census forms.


glarwill:
I guess this is ready for review? The failing unit tests fail on the main branch as well. The php7.4 one may need to be re-run as it terminated for some other reason after it noted the same issues as the php8.0 test.

fisharebest:
Thanks.

The failing test is caused by a caching issue.  The test does two things that would normally be separate requests (imports data and generates statistics), and the first one creates a cached value that is incorrect for the second.

PR #3936, opened 2021-06-24
Canada census 3

glarwill:
This is LARGE, sorry.

5 more years of Canada Census docs added.

Much is copied from other years' documents. Notable other changes include:

Age.php
6 new census columns for these years.
fixed a comment in CensusTest.php


glarwill:
This one is ready for review. The unit test that fails is caused by the current failure in main (in the php7.4) and perhaps the same thing in the 8.0 test.

fisharebest:
Can you use full names, instead of abbreviations.  Abbreviations can be confusing for people who speak English as a second language.

e.g. `BDay` => `Birthday`, `Can` => `Canadian`, `CanMW` => `Canadian????`

The various columns for status (e.g. `CensusColumnConditionCanMS`) unecessarily extend a class such as `CensusColumnConditionCan`.  Since they don't need anything from this class, they can simply extend `AbstractCensusColumnCondition`.

The `Age` class is soon to be removed.  Perhaps Move this function to your `AgeNextBirthday` class.

glarwill:
Any hints on resolving the phpstan errors?

fisharebest:
> Any hints on resolving the phpstan errors?

The PHPstan rules are continually updated, and so errors can appear/disappear without any corresponding changes to the code.

These issues are all in old code that hasn't changed for a while.

I've updated the PHPstan baseline file, so they should disappear shortly.

glarwill:
> The PHPstan rules are continually updated, and so errors can appear/disappear without any corresponding changes to the code.
> 
> These issues are all in old code that hasn't changed for a while.
> 
> I've updated the PHPstan baseline file, so they should disappear shortly.

OK, thanks!
I believe the unit test errors noted (one in each 7.4 and 8.0) are also from old changes.
This PR is ready to go, if you are satisfied with the changes per your previous suggestions.

PR #3951, opened 2021-07-09
Show the content title on the customize blocks page

ddrury:
Really only useful for HTML blocks


fisharebest:
Hard-coding block names into logic and/or ssuming that blocks have a config variable `title` that is suitable for this purpose is a bit [brittle](https://en.wikipedia.org/wiki/Software_brittleness). 

It was pointed out on the forum that many users prefer to store the title in the HTML, where they have better control over the formatting.

I think we need a different solution.

PR #3962, opened 2021-07-19
Update GedcomCodeTemp.php

z-yurets:
https://en.wikipedia.org/wiki/KyivNotKiev


fisharebest:
When you correct a translation string like this, you must also update the translation files (`.po`).  Otherwise all the existing translations would be lost.

If you update the `.po` files, then you must also synchronise the changes with the translation server at https://translate.webtrees.net

I have done this, and also made the changes in the `main` (version 2.1) branch of the code.

PR #4000, opened 2021-08-21
Bootstrap v5: reversed order of events on active state of toggle buttons

jon48:
I have not found a statement supporting this in the Bootstrap v5 migration documentation, but it seems that the `click` event is now triggered **after** the `active` class has been toggled on toggle buttons (it used to be before), so that the state persistence logic needs to be reversed.

Could you please confirm you experience the same behavior before merging?

_Steps to reproduce_
With the current code (before PR), go to the the individuals table (same issue with families table), for instance the [PHILLIPS individuals on demo-dev](https://dev.webtrees.net/demo-dev/tree/demo/individual-list?surname=PHILLIPS):
- Assuming that the buttons are initially not active (`state-of-show-statistics = F` in local storage), toggle the "Show statistics charts" button to activate it ("Show parents" could be used as well for thes testing). 
- Making sure that the action has been completed, note the status of the button (it should be `active`).
- At that stage, you can check the local storage value: it will probably be `state-of-show-statistics = F` - which is not correct
- Refresh the page. 
- The expected behaviour is that the button "Show statistics" should be active. Based on my testing, it is not active.

Another test if to toggle successively the button, and check the `state-of-show-statistics` value in local storage, it does toggle, but with the opposite value as the one expected


fisharebest:
> Could you please confirm you experience the same behavior before merging?

I experience the same behaviour.

It seems this code is duplicated elsewhere (`persistentToggle()`)

* I guess this code should be de-duplicated.
* It would probably be good to initialise these using `data-` attributes - similar to bootstrap.
* It would probably be clearer if (like bootstrap) we name-spaced our `data-` attributes.  e.g. `data-wt-xxx`

PS - for similar PRs, could you only include the changes to `*.js` and `*.css` - and not (`*.min.js` and `*.min.css`).

Firstly, it can make it difficult to merge changes.  Secondly, it is bad-practice to merge changes to minified JS code, as it is difficult to review the changes.  I'm sure you haven't added anything malicious, but it is difficult to check.  I'll always run `npm run production` after merging any changes to `.js` and `.css` files anyway.

jon48:
> PS - for similar PRs, could you only include the changes to `*.js` and `*.css` - and not (`*.min.js` and `*.min.css`).

Understood. :+1: 
I thought I was making your life easier by providing the minified versions, but I understand your points, and will only commit the code change in the future.

PR #4004, opened 2021-08-22
Bootstrap5: migration `form-select` and `card-deck`

jon48:
### Use of `form-select`
`select` elements used to be assigned the `form-control` class. This is aimed for `input` and `textarea`. It was not causing any issue in BS4, but in BS5, the following line has been added to `form-control` which is removing the dropdown caret on `select`. Use the appropriate `form-select` class instead.

```css
.form-control {
    [...]
    appearance: none; // Fix appearance for date inputs in Safari
```
image

### `card-deck` dropped

The `card-deck` class has been removed in favour of the grid system (see [documentation](https://getbootstrap.com/docs/5.0/migration/#card)).

This is not obvious everywhere, but for instance, the *Family* and *Other* tabs of the [statistics page](https://dev.webtrees.net/demo-dev/module/statistics_chart/Chart/demo) do not display correctly, as well as the [*Customise this page* page](https://dev.webtrees.net/demo-dev/tree/demo/my-page-edit).

### DIFF
--- a/resources/views/components/select-family.phtml
+++ b/resources/views/components/select-family.phtml
@@ -19,7 +19,7 @@ $at = $at ?? '';
 
 <select
     autocomplete="off"
-    class="form-control select2 <?= $class ?? '' ?>"
+    class="form-select select2 <?= $class ?? '' ?>"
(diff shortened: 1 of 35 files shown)


fisharebest:
Thanks.

Issue #4011, opened 2021-08-25
Problem with Egulias EmailValidator


hartenthaler:
Since several weeks I do not get any reminder message from the module "Reminder" of @UksusoFF. When seaching my site error log, I found the following messages (the cron job is started at 07:00 every day):
`25.08.2021 07:00:50 ahnen.hartenthaler.eu [client 2axx:xxx:xxx::] AH01215: PHP Fatal error:  Declaration of Egulias\\EmailValidator\\Validation\\MessageIDValidation::isValid(string $email, Egulias\\EmailValidator\\EmailLexer $emailLexer): bool must be compatible with Egulias\\EmailValidator\\Validation\\EmailValidation::isValid($email, Egulias\\EmailValidator\\EmailLexer $emailLexer) in /mnt/xxx/xxx/xxx/xxx/htdocs/hermann/hartenthaler/php/webtrees/vendor/egulias/email-validator/src/Validation/MessageIDValidation.php on line 23: /home/strato/http/premium/rid/xx/xx/xxx/htdocs/hermann/hartenthaler/php/webtrees/index.php`

Is there a bug in the Egulias package?

fisharebest:
Several people have reported this on the forum - but I have been unable to reproduce it.

They all reported that the issue was fixed by deleting and reinstalling the `/vendor/` folder.

fisharebest:
The file `Validation\MessageIdValidation.php' is removed between 2.0.14 and 2.0.15.

It is not used in 2.0.15, and should therefore be ignored.

But, I can reproduce this error by

* installing 2.0.14
* sending an email
* copying 2.0.15 over 2.0.14
* sending an email

It is caused by some code in Swiftmailer.
The version included in 2.0.15 uses `class_exists('MessageIdValidation')` - which loads the class and triggers the error.

fisharebest:
Since:

* we have always supported upgrades by simply overwriting existing files
* webtrees may not have write permission to delete this file

I think there is nothing we can do in code to automatically fix it.

The best option may be to add something like this to the control panel

```
if (file_exists('.../MessageIdValidation.php')) {
  try {
    delete('.../MessageIdValidation.php');
  catch {
    display_message('Please delete the file .../MessageIdValidation.php or your email won't work');
  }
```

fisharebest:
> I think there is nothing we can do in code to automatically fix it.

Except, perhaps patch the version of swiftmailer included in 2.0.17...

PR #4015, opened 2021-08-27
Replace EOL Swiftmailer library by Symfony Mailer

jon48:
Following my comment in #4011, I open this PR to prepare the replacement of the Swiftmailer library by Symfony Mailer.

As advertised on the website, and based on [the announcement of its author](https://symfony.com/blog/the-end-of-swiftmailer), Swiftmailer is reaching end-of-life, with its maintenance being stopped by the end of November 2021, and the Symfony Mailer component (written by the same author) is the recommended replacement.

The PR intends to implement the replacement, whilst discussing some minor points raised during the upgrade. I will keep it as a draft until all questions/comments have been addressed.

The upgrade is rather straightforward, with very similar classes and signatures. In my first commit, I have taken the approach to keep essentially the same code, with just the few namespace/class replacement and minor adjustments.

@fisharebest, there are a couple of points I would like to pass by you:

### Email validation

During my testing, it appears that the email handling is slightly less permissive than before, so an exception can be raised if the email is not valid, as early as the constructor of the new `Address` class, and so as soon as the first line of the `send` method (not covered by the `try/catch` block. This raised some questions to me:

- Shall we extend the scope of the `try/catch` to the whole `EmailService::send` method, so that exceptions are captured during the creation of the message or the transport?
- Should we add a validation of the email before sending? This is normally done when saving the email, so maybe not necessary
- I have slightly modified the `isValidEmail` method used when saving the email, so that it uses at least the same validation as the one used when sending. This is simply by calling the constructor, as Symfony Mailer does not provide any method or utility to check the validity of the email in a consistent manner (probably something they could improve a bit).
  We could go even further and replace the validation logic by using directly the egulias/email-validator library to perform the validation (as this is the underlying library in Symfony Mailer), using MessageIDValidation & DNSCheckValidation. This provides maybe a more robust check, but the current logic is probably enough (I have done some testing using the Egulias data set, and its validator and the `isValidEmail` do not agree only in a handful of edge cases).

### SSL / TLS / STARTTLS confusion

That may be the topic where I have noticed the most sensible change.
Swiftmailer had 3 modes for email security. I had never taken a deep look at them, but they were particularly badly named in the library:
- no security (plain non-secure SMTP was used, the 'none' option in webtrees)
- 'ssl' : this is the SMTP over TLS option, where the connection is secure from its initialisation, at the transport layer. This is the 'Implicit TLS' or 'Enforced TLS' option with some providers, or port 465 with Gmail. The name SSL is completely outdated now, as hopefully everybody have switched to TLS by now, but is probably useful not to confuse with the next option
- 'tls': probably the most confusing one. This is the option where the SMTP command STARTTLS is called after a non-secure connection has been established with the server, to upgrade the connection to a secure one (over TLS). This is the 'Explicit TLS' or 'Opportunistic TLS' option with some providers, or port 587 with Gmail.

Symfony Mailer has slightly modified the approach, and it now automatically tries to issue a STARTTLS command if the server allows it, without any configuration. Only the ex-'ssl' mode now needs to be explicitly set (even there, if no configuration is given, it will try to infer the 'ssl' mode if the port is 465).

It means that the options are now reduced to a simple binary choice: Use Implicit TLS ?
- If yes, then the TLS channel is created first when connecting to the SMTP server (ex-'ssl' option)
- If no, a non-secure channel is created, the STARTTLS command is issued, and depending on the handshake result, either the connection remains non-secure, or it will be upgraded to TLS (so, the ex-'none' and 'tls' options result in the same outcome).

In the first commit, I have not modified the options for the 'Secure connection' configuration, but two of them are redundant. So what's next?

- Keep the options as they are: I would however improve the wording, i.e. 'ssl' => '*SMTP over TLS (Implicit TLS)*', 'tls' => '*SMTP with STARTTLS (Explicit TLS)*'
- Keep a dropdown, with the same preference value `SMTP_SSL`, but reduced to two options : 'none' => '*Best effort security (use STARTTLS if available)*', 'ssl' => '*SMTP over TLS (Implicit TLS)*'. No migration is strictly required, as all tests can be done against 'ssl', so 'none' and the defunct 'tls' option with produce the same result.
- Refactor the preference `SMTP_SSL`, and transform it to a binary value. In the control panel, this would become a 'yes/no' choice to '*Use Implicit TLS*'. A small migration is to be planned to change 'ssl' to '1' and 'none'/'tls' to '0'.

### DKIM

How did you manage to test it? The email providers that I have tend to overwrite any DKIM value I set, so I find it difficult to test. The classes seem quite straightforward, so assuming the previous code was working, I guess the new one will work equally...

### DIFF
--- a/app/Services/EmailService.php
+++ b/app/Services/EmailService.php
@@ -25,13 +25,16 @@ use Fisharebest\Webtrees\I18N;
 use Fisharebest\Webtrees\Log;
 use Fisharebest\Webtrees\Site;
 use Psr\Http\Message\ServerRequestInterface;
-use Swift_Mailer;
-use Swift_Message;
-use Swift_NullTransport;
-use Swift_SendmailTransport;
-use Swift_Signers_DKIMSigner;
-use Swift_SmtpTransport;
-use Swift_Transport;
+use Symfony\Component\Mailer\Mailer;
+use Symfony\Component\Mailer\Transport\NullTransport;
+use Symfony\Component\Mailer\Transport\SendmailTransport;
+use Symfony\Component\Mailer\Transport\Smtp\EsmtpTransport;
+use Symfony\Component\Mailer\Transport\TransportInterface;
+use Symfony\Component\Mime\Address;
+use Symfony\Component\Mime\Crypto\DkimOptions;
+use Symfony\Component\Mime\Crypto\DkimSigner;
+use Symfony\Component\Mime\Email;
+use Symfony\Component\Mime\Message;
 
 use function assert;
 use function checkdnsrr;
@@ -64,11 +67,10 @@ class EmailService
      */
     public function send(UserInterface $from, UserInterface $to, UserInterface $reply_to, string $subject, string $message_text, string $message_html): bool
     {
-        $message   = $this->message($from, $to, $reply_to, $subject, $message_text, $message_html);
-        $transport = $this->transport();
-        $mailer    = new Swift_Mailer($transport);
-
         try {
+            $message   = $this->message($from, $to, $reply_to, $subject, $message_text, $message_html);
+            $transport = $this->transport();
+            $mailer    = new Mailer($transport);
             $mailer->send($message);
         } catch (Exception $ex) {
             Log::addErrorLog('MailService: ' . $ex->getMessage());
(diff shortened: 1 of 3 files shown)


fisharebest:
> Shall we extend the scope of the `try/catch` to the whole `EmailService::send` method

I guess that errors here could include things like invalid email addresses or server names.
We should probably handle these the same as failure-to-send errors - i.e. log them for an admin to find later.
So yes - add them to the `catch` block.

> Should we add a validation of the email before sending?

This would indicate that we've built the message incorrectly.
If invalid, we'll presumably get an error - which will be picked up by our existing error handling.
So probably not needed.

> by using directly the egulias/email-validator library to perform the validation (as this is the underlying library in Symfony Mailer)

Symfony could change its dependencies in the future.  Probably best to use any validation through the Symfony API only.

> I have not modified the options for the 'Secure connection' configuration, but two of them are redundant. So what's next?

There are two options now.  But other options may appear in the future (e.g. new encryption protocols).
So maybe best to keep it as an option list.

Is it possible that a future/different mail library may split STARTTLS back into two options?
If that's the case, then maybe keep the current options - but handle two of them the same?

But I agree that changing the names of the options to the "correct" terminology would be a good thing.
i.e. use the same text that users would find in their ISP's documentation.

> DKIM How did you manage to test it? 

Indirectly ;-)  The user who requested this feature tested it for me.

fisharebest:
PS - can you remove `composer.lock` from the patch?

It will probably conflict with recent updates - plus it makes it harder to back-port to the 2.0 branch.
I'll regenerate it after merging the PR.

jon48:
> Shall we extend the scope of the `try/catch` to the whole `EmailService::send` method
> So yes - add them to the `catch` block.

Done.

>  egulias/email-validator library: Symfony could change its dependencies in the future. Probably best to use any validation through the Symfony API only.

Indeed. I therefore keep the Symfony API call for the validation.

> Is it possible that a future/different mail library may split STARTTLS back into two options?

Yes, you are right. The fact that two options are the same are implementation details, and not inherent to the protocol, which could be back to 3 different implementations in the future. So better to stick to the protocol interface and keep the 3 options as they are.

> But I agree that changing the names of the options to the "correct" terminology would be a good thing.
> i.e. use the same text that users would find in their ISP's documentation.

I have tried to look around what mail or webhosting providers are using in their documentation. To say the least, they do not really help with clearing the ambiguity between the different protocols and terminology. But below is a table with the text I could find on their support page, so I have changed the text to : 'ssl' => '*SSL/TLS*' and 'tls' => '*STARTTLS*'

|Client/Service provider|No security|SMTP over TLS|SMTP with STARTTLS|
|-------------|-----------------|--------------|----------------|
|Outlook client	|None	|SSL/TLS	|STARTTLS|
|Thunderbird	|None	|SSL/TLS	|STARTTLS|
|Android		|	|SSL/TLS	|TLS or STARTTLS|
|Gmail 1		|	|Requires SSL: Yes	|Requires TLS: Yes (if available)|
|Gmail 2		|	|SSL		|TLS/STARTTLS|
|Yahoo		|	|Requires SSL - Yes	|Requires TLS - Yes (if available)|
|Outlook.com	|	|		|STARTTLS|
|Hostgator	|	|SSL/TLS	||
|A2 Hosting	|	|SSL/TLS	||
|Blue Host	|	|SSL/TLS	||
|Cpanel		|	|Secure SSL/TLS	||

> DKIM How did you manage to test it?
> Indirectly ;-) The user who requested this feature tested it for me.

Fair enough. Well, hopefully he will let us know if this stops working... :sweat_smile:

> PS - can you remove `composer.lock` from the patch?

Done, the PR checks will probably fail, because it will not load correctly the dependencies, but they are running fine on my PC.

fisharebest:
Thanks!

fisharebest:
> back-port to the 2.0 branch.

It seems I can't do this.  `symfony/mailer` requires PHP7.2 - and webtrees 2.0 needs to run on PHP 7.1.

jon48:
Hmm, indeed, when I integrated it, I was more with a view that this would be a 2.1 evolution, and 2.0 would stay with swiftmailer.

If we want to backport, we can downgrade symfony/mailer to 4.4 (LTS version, until November 2022), which is requiring php 7.1.3.
That would downgrade egulias/email-validator in the process to 2.1 (vs 3.1.1 currently),  which causes one of the `testIsValidEmail` test to fail (`example(comment)@example.com`), which is expected, as 3.1.1 adds new RFC constraints.

What is your view?

fisharebest:
Backporting would be done to simplify support - i.e. same code in both branches.

If we'd need different versions, then better to stick with the existing/tested code.

PR #4019, opened 2021-08-30
Bootstrap5: use new accordion native component

jon48:
Bootstrap 5 introduce a proper Accordion component, with specific classes (BS4 was using a combination of cards component and `data-toggle` attributes).

The PR migrate the few accordions structure to use the new component, as the old one shows some discrepancies (small gap between headers, rounded headers...)

The Accordion styling is a bit different, and have several states, but I have restored the card style, so the header does not look different.

However, the component comes as well with the dropdown arrow already added, so I have removed the collapse/expand icons in favour of the native styling. That can be reverted back if necessary.
The only aspect I am not entirely satisfied is that the arrow in the expanded state is blue. This is actually because the full text color is blue in the native BS5 styling. To change the arrow color, it is needed to add the full SVG image, and tweak the fill parameter (this is configurable in Bootstrap, but the themes are not generating their own Bootstrap stylesheet, and  only override the default stylesheet).

```css
.accordion-button:not(.collapsed):after {
    background-image: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' fill='#{$accordion-icon-active-color}'><path fill-rule='evenodd' d='M1.646 4.646a.5.5 0 0 1 .708 0L8 10.293l5.646-5.647a.5.5 0 0 1 .708.708l-6 6a.5.5 0 0 1-.708 0l-6-6a.5.5 0 0 1 0-.708z'/></svg>");
}
```

image

### DIFF
--- a/resources/css/_base.css
+++ b/resources/css/_base.css
@@ -83,6 +83,17 @@ a:hover {
     display: none;
 }
 
+/* Accordions */
+.accordion-button,
+.accordion-button:not(.collapsed) {
+    background-color: rgba(0,0,0,.03);
+    color: currentcolor;
+}
+
+.accordion-button:after {
+    display: none;
+}
+
 /* Link styling */
 a, .btn-link, .nav-link {
     color: var(--link-color);
--- a/resources/css/minimal.css
+++ b/resources/css/minimal.css
@@ -1042,3 +1042,13 @@ a > .wt-icon-arrow-up:hover {
 .col-form-label {
     font-weight: bold;
 }
+
+/* Accordions */
+.accordion .wt-icon-expand,
+.accordion .wt-icon-collapse {
+    display: none;
+}
+
+.accordion-button:after {
+    display: block;
+}
--- a/resources/views/admin/trees.phtml
+++ b/resources/views/admin/trees.phtml
@@ -49,21 +49,21 @@ use Illuminate\Database\Capsule\Manager as DB;
     </div>
 <?php endif ?>
 
-<div id="accordion" role="tablist" aria-multiselectable="true">
+<div id="accordion" aria-multiselectable="true">
     <?php foreach ($all_trees as $managed_tree) : ?>
         <?php if (Auth::isManager($managed_tree)) : ?>
-            <div class="card">
-                <div class="card-header" role="tab" id="card-tree-header-<?= $managed_tree->id() ?>">
-                    <h2 class="mb-0">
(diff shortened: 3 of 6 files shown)


fisharebest:
I've been looking at various UX/usability guidelines for accordions.

Most recommend having the icon to the left of the text.
So I am curious as to why bootstrap put them on the right.

With the current icons, we provide an opportunity for the themes to customise it.
With the rotating bootstrap icon, it is a little harder.

What do you think of:

a) minimal theme - hide the "webtrees" icons and use the default bootstrap icon (with its default colours).
b) other themes - hide the bootstrap icon, and keep the webtrees icons.

jon48:
> What do you think of:
> a) minimal theme - hide the "webtrees" icons and use the default bootstrap icon (with its default colours).
> b) other themes - hide the bootstrap icon, and keep the webtrees icons.

I am fine with that. I will work on resubmitting something along those lines.

>  (with its default colours).

When you say that, do you mean the default colour only for the icon, or for the complete accordion element (i.e. header background and text colour)?
My understanding is only the icon (which would be a pain to overwrite), and still customise the header background and text colour to stay consistent with the 2.0 styling, but just want to make sure.

fisharebest:
> When you say that, do you mean the default colour only for the icon, or for the complete accordion element (i.e. header background and text colour)?

I had originally envisaged the "minimal" theme as being a "bootstrap" theme.
(There were once many tools to create customized versions of bootstrap - which could then be added to this theme.)

But if we are going to do that, then the minimal theme might need a bit of an overhaul…

> My understanding is only the icon (which would be a pain to overwrite), and still customise the header background and text colour to stay consistent with the 2.0 styling, but just want to make sure.

…so for the time being, your understanding is correct.

jon48:
I have done the changes accordingly.

Let me know if you are happy with the way I implemented. 
Because it was so easy to hide the bootstrap icon, I have done it in `_base.css`, so all themes by default show the webtrees icon, and hide the bootstrap one, and any theme can invert that behaviour if they wish. That means that for minimal, I actually end up "overwriting the overwrite", which makes me wonder if this is even necessary to have a specific behaviour for that theme, and I would be happy to revert it.

I could have taken the other approach: by default, hide the webtrees icon, and leave the bootstrap one, but that did not seem the direction you wanted to go, as this is for the theme to explicitly invert this behaviour (so something to document for custom themes).

PR #4025, opened 2021-09-02
Tests: fix incorrect `TestCase::validateHtml` logic

jon48:
There are a few issues with the new `TestCase::validateHtml` method, some of them being hidden by the fact that the method is exited prematurely most of the time.

The main issue is the initial extraction of the html.
If my understanding is correct, the method progresses through the string by finding the next `<` or `>`
But by using `substr($html, 0, strcspn($html, '<>'))`, it was actually extracting whatever string was between the beginning of the string and the first `<` or `>`, which would be plain text.
 Most of the time, `<` would be the first character, so the result would be an empty string, and the method would exit, without having actually gone through further testing, The only case I could find was in the `AddressFaxTest`, where the input is a plain text, which was resulting in an infinite loop, because if would not enter in any of the two `str_starts_with` conditions, and would never been shorten up to an empty string.

Then, a couple of other issues appears understandingly because they could not be tested:

- the regex for extracting the tag is missing the definition of the attribute name (only `=` and the attribute value was present), so it was not matching correctly. The allowed characters for attribute name are fairly complex, so I have taken a simplified approach of allowing only `a-z`, `_`, `-` which is enough to pass the current test suite, but that can be tweaked. 
- There are cases where attributes does not have any value, so the value part needs to be optional. For instance, the input value for `EventsRecordedTest` contains `<select class="..." multiple>`
- The HTML input can contains JS code between <script> tags. The problem is that within that code, the `<` or `>` characters can be present, in a non-HTML context, and the next main `substr` will then extract a string that does the match the expected HTML format (resulting in either a `Closing tag matches nothing` or `Unrecognised tag` test failure). I have taken the approach of ignoring anything until the next `</script>` tag, and the HTML parsing can then continue. The `EventsRecordedTest` test is again one containing `<script>` in its input.

Just a final comment, the parser assumes the HTML tags are lowercase, which is not technically required by HTML5 (contrary to XHTML). This is what I expect from most of the code, the tested code does indeed, so I have not changed anything.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -54,8 +54,6 @@ use function strlen;
 use function strpos;
 use function substr;
 
-use function var_dump;
-
 use const UPLOAD_ERR_OK;
 
 /**
@@ -278,7 +276,6 @@ class TestCase extends \PHPUnit\Framework\TestCase
         $this->assertStringStartsWith('<DOCTYPE html>', $html);
 
         $this->validateHtml(substr($html, strlen('<DOCTYPE html>')));
-
     }
 
     /**
@@ -291,7 +288,7 @@ class TestCase extends \PHPUnit\Framework\TestCase
         $stack = [];
 
         do {
-            $html = substr($html, 0, strcspn($html, '<>'));
+            $html = substr($html, strcspn($html, '<>'));
 
             if (str_starts_with($html, '>')) {
                 $this->fail('Unescaped > found in HTML');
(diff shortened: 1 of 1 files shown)


jon48:
Just re-reading my comment, I realised that I was assuming the `script` tag was never self-closing.
I just force-pushed a version taking that case into account (there is no example in the test suite, but just in case).

fisharebest:
Hi @jon48 - I added this code to the wrong branch, and commited it by mistake.  I would much prefer to use a third-party library to do this - but cannot find anything suitable :-(

For consistency (and for the possibility to use XML tools in the future), I have been trying to generate "XHTML5".

Hence checking for `<input checked="checked" />` instead of just `<input checked>` is deliberate.

fisharebest:
> contains `<select class="..." multiple>`

I just changed this to `contains <select class="..." multiple="multiple">`

jon48:
> Hi @jon48 - I added this code to the wrong branch, and commited it by mistake. I would much prefer to use a third-party library to do this - but cannot find anything suitable :-(

Yes, I realised by looking something similar that there were very limited options.
I have seen Masterminds/html5-php, but whereas it provides a better handling of HTML5 than the native DOMDocument, it clearly states that it is not a proper validator.

> Hence checking for `<input checked="checked" />` instead of just `<input checked>` is deliberate.
It makes sense. I see that you have just committed the changes to remove the optional value, and fix the couple of other instances where attributes were used without value. 👍

PR #4029, opened 2021-09-06
Fix: OSK script toggle stopped working

jon48:
I am not sure if there was a reason behind, but the `name` attribute was removed from the `input` list of scripts in the on-screen-keyboard, in commit d4786c66.
However, this attribute is used by bootstrap to toggle between the buttons. Without it, you can switch the button on, but the others are not unchecked, and you cannot use it anymore after.

This PR reinstates the `name` attribute.


fisharebest:
> I am not sure if there was a reason behind, but the name attribute was removed

I couldn't see how it was used, and a (too) quick check showed it working without.

Thanks!

PR #4064, opened 2021-10-10
Fix validator signature for array method + unit tests

jon48:
While trying to use the new Validator class in my module, one of the code analysis tool (Psalm - phpstan was not detecting it) made me realised that `Validator::array` had an inconsistent signature with its implementation, as it does not cater for the `null` value being on of the valid output.

I added the nullability on the return type and the documentation, as well as the unit tests for that class.

### DIFF
--- a/app/Validator.php
+++ b/app/Validator.php
@@ -135,10 +135,6 @@ class Validator
                 return null;
             }
 
-            if ($value === null) {
-                return null;
-            }
-
             throw new LogicException(__METHOD__ . ' does not accept ' . gettype($value));
         };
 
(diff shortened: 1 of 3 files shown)


fisharebest:
> I am not sure I am understand what you mean. For me, there is a difference between an empty array and a missing one.

PHP / HTML forms specify array elements, not arrays.

Suppose we wish to include an array `$foo` in a form.  Each element is a separate input field.

```
<input name="foo[0]" value="a">
<input name="foo[1]" value="b">
```

But if the array is empty, then we don't include any elements at all.
Hence we cannot send an empty array in an HTML form - only no array elements.

PR #4079, opened 2021-10-18
search wt_place_locations table for autocomplete on places

tkempf:
allows autocompletion of Places by additionally searching in wt_place_locations table.


tkempf:
replaced by v4 module in 2.1.x

PR #4115, opened 2021-11-12
Census assistant - Use markdown unordered lists for plain text census elements

ddrury:
Rather than use markdown paragraph or line breaks, convert plain text paragraphs into an unordered list that can be styled by css


ddrury:
Rewriting so that all note formatting is handled on display, and tree preference 'FORMAT_TEXT' is always obeyed.

Issue #3927, opened 2021-06-18
Make history modules selectable by users


Lars1963:
_Originally posted by @hartenthaler in https://github.com/fisharebest/webtrees/issues/3924#issuecomment-863614426_

As discussed, it might be a good idea to make the history modules selectable by the user

`Maybe a special menu in the control panel would help the admin to manage more and more historic data files. But there is a deeper question behind: who decides about which historic data is of interest for the users. At the moment the admin selects the historic data files he thinks are of interest for his tree, but maybe a user from China knows Chinese history very well and likes to see Chinese history events even if all events in the tree are located in Europe because he likes to compare what happened in China at the time the events in the tree happened. Maybe the users should be able to select the history files, but the normal user will not find and use this possibility, so we should not make it too flexible.`

fisharebest:
I'm not sure this is necessarily a good idea.

Korean events are only likely to be of interest to Korean speakers, and only included on Korean sites.
Translating these into 60 other languages is a lot of effort.

For some events (perhaps in countries with multiple official languages), then it may make sense for the module to supply a version for each language - but not as translatable strings.

magicsunday:
I would be more in favor of making these special history modules available as separately installable extensions and removing them completely from the Webtrees core.

fisharebest:
Historic events are now shown only for the current language.

i.e. events in French are only shown when the current language is French.

Where applicable (countries with multiple official languages), the module can provide separate events for each language.

fisharebest:
> As discussed, it might be a good idea to make the history modules selectable by the user

I think the selection of *all* modules (including history modules) should be a matter for the site-admin, not for users.

PR #4116, opened 2021-11-14
ENHANCED: DutchMonarch.php now translatable 

Lars1963:
Please take this in account for actual dev. If all is working as supposed, I've more historical modules waiting.

#3927


fisharebest:
As mentioned in another issue, I don't think we should translate every event into every language.  Most are not relevant outside their own language/geographic area.

Instead, modules should provide (hard-coded) translations for all relevant languages.  The code structure has been updated to make this clearer.

Issue #4123, opened 2021-11-23
Markdown formatting


ddrury:
Here are a few comments based on this thread in the forum https://www.webtrees.net/index.php/en/forum/help-for-2-0/36162-v2-0-17-things-that-no-longer-work-after-update#88276, commit https://github.com/fisharebest/webtrees/commit/b6b785a1da4d8c051f96d8525d89e5f6e3d6e201 and pull request https://github.com/fisharebest/webtrees/pull/4115

- It is clear that formatting for markdown encoded text containing line/paragraph breaks goes further than just shared notes.
- The commit mentioned above only addressed the first line break in each census assistant part. My pull request fixes that insofar as each line break is considered a paragraph and is changed to a `<li>` element, nevertheless it is simple to adapt should the change to unordered lists be considered unnecessary.
- Only notes created after the above mentioned code changes are affected. A data fix is needed to apply the formatting changes to existing plain text content that contains line/paragraph breaks where that text needs to be interpreted as markdown.

ddrury:
Thinking further why not do all the processing on display and only do sufficient on creation (ie just add as many newlines as necessary to indicate whats required)

fisharebest:
> only addressed the first line break in each

Can you give any specific examples of what isn't currently working?

ddrury:
Specifically:
Add more than one paragraph to any of the census assistant fields, only the first paragraph gets the double space added to the `\n`. Practically this only really applies to the comments section, but for completeness should target each field.

Other things that don't work are formatting for notes attached to sources or media.

Blowing my own trumpet, all these "problems" are addressed in https://github.com/fisharebest/webtrees/pull/4128

fisharebest:
67721f6fe545e33878bd31b73a960fe472601f42 adds a new config option - "markdown with line breaks"

PR #4128, opened 2021-11-30
Format notes on display rather than by adding formatting on creation. 

ddrury:
Solves issue https://github.com/fisharebest/webtrees/issues/4123 (I think)

To achieve this:

- Recent changes to the census assistant have been reverted (this obviates the need for retrospective changes to existing notes)
- Add a new function getHtml() to app\Note.php which returns html formatted according to the tree FORMAT_TEXT preference. **Note** this is a potentially breaking change as the FORMAT_TEXT preference is strictly applied, currently any note containing markdown table formatting is always interpreted as markdown. However it is simple to revert to the current behaviour (see the comment in app\Note.php line 71)

Because the formatting code is in the Note class, it is necessary to convert in-line notes into temporary dummy note objects so the same code can be used for all notes.

- Non-markdown formatting just nl2br's the text.
- For markdown formatting, the new function works by splitting the text into paragraphs (split on "\n" char, tables are treated as a single paragraph) and applying formatting to each paragraph independently and then joining the bits back together. This means that each text paragraph is wrapped in `<p></p>` and tables are wrapped in `<div></div>` which allows styling if needed.

**Note**: some changes have involved classes FunctionsPrint & FunctionsPrintFact that have been slated for removal in 2.1

### DIFF
--- a/app/Functions/FunctionsPrint.php
+++ b/app/Functions/FunctionsPrint.php
@@ -35,7 +35,6 @@ use Fisharebest\Webtrees\Registry;
 use Fisharebest\Webtrees\Services\GedcomEditService;
 use Fisharebest\Webtrees\Services\ModuleService;
 use Fisharebest\Webtrees\Tree;
-use Illuminate\Support\Str;
 use Ramsey\Uuid\Uuid;
 
 use function app;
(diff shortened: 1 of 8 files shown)


ddrury:
BTW, I did think of shifting the view from Functions::printNoteRecord into app/Note::getHtml and by doing so it would make it simple to do away with Functions::printNoteRecord altogether, but it would make  app/Note::getHtml quite a lot bigger

PR #4165, opened 2022-01-19
On change of map layer, save new layer to local storage

ddrury:
- Added a name property to each layer object.
- Added leafletjs.baselayerchange() function to the map which when it fires saves the name property to local storage.
- On page load use the stored value to set the displayed layer.
- Tidy formatting in Bing maps module (unrelated to above)

### DIFF
--- a/resources/js/webtrees.js
+++ b/resources/js/webtrees.js
@@ -631,47 +631,36 @@
       zoomoutTitle: config.i18n.zoomOut,
     });
 
-    const resetControl = L.Control.extend({
-      options: {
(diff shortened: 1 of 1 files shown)


ddrury:
See also https://github.com/fisharebest/webtrees/pull/4727 and https://github.com/fisharebest/webtrees/pull/4703

As this PR provides a method to save a users preferred map provider and layer to local storage it mitigates the problem whereby the default layer may change depending on how the map providers are enabled/disabled in the control panel

PR #4166, opened 2022-01-19
Generate exception when no map providers enabled

ddrury:


fisharebest:
I see a page-within-a-page.

I guess something isn't setting the header X-Requested-With`, so the exception handler doesn't know it is responding to an ajax request.  Let me investigate...

Screenshot 2022-01-28 at 19 34 43

PR #4172, opened 2022-01-24
shift resetControl to webtrees.js

ddrury:
Avoid duplication of code.

resetControl existed in each map provider view, by moving to webtrees.buildLeafletJsMap() we remove that duplication although a small function (resetActions) that carries out the required reset actions for each individual view needs to be created and passed into webtrees.buildLeafletJsMap().


ddrury:
Withdrawn pending removal of jQuery from map views

PR #4176, opened 2022-01-26
Remove jQuery dependencies from Place tab

ddrury:
In addition to removing jQuery,

- As OpenStreetMap is now not the only provider, rename 'osm-' references to 'wt-'
- Reformatted to semistandard


ddrury:
Greg,
There is a bug here (got reverted somehow)
In the resetControl onAdd function line 59 needs to be changed from
`sidebar.scrollTo(sidebar.children(':first'));`
to
`sidebar.firstElementChild.scrollIntoView(scrollOptions);`

Sorry

PR #4178, opened 2022-01-26
Remove jQuery dependencies from location edit view

ddrury:
In addition to removing jQuery,

- As OpenStreetMap is now not the only provider, rename 'osm-' references to 'wt-'
- Reformatted to semistandard

### DIFF
--- a/resources/views/admin/location-edit.phtml
+++ b/resources/views/admin/location-edit.phtml
@@ -26,7 +26,7 @@ use Fisharebest\Webtrees\View;
 
 <div class="row form-group mb-3">
     <div class="col-sm-12">
-        <div id="osm-map" class="wt-ajax-load col-sm-12 osm-admin-map" dir="ltr"></div>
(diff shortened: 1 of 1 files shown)


resources/views/admin/location-edit.phtml
```diff
@@ -91,103 +91,105 @@ use Fisharebest\Webtrees\View;
...
-              if (!map.getBounds().contains(marker.getLatLng())) {
-                map.panTo(marker.getLatLng());
-              }
-            });
-
-        return 'Leaflet map interface for webtrees-2';
-    })();
+  'use strict';
+
+  window.WT_ADMIN_MAP = (function () {
```
fisharebest, line 96: Do we need this assignment (and the corresponding `return` at the bottom?

I can't see that it is used anywhere?

ddrury: Can't remember why it was originally done but it can be removed - I'll do it

ddrury: Done

PR #4193, opened 2022-01-30
correct Italian translation for age of individuals

gmarcon:


fisharebest:
FYI, translations are managed at https://translate.webtrees.net.

The registration system on the server is broken (it cannot send email), so if you would like an account, please email me directly.

PR #4195, opened 2022-01-31
Places module - disable marker clustering on Click of sidebar entry

ddrury:
~~**EXPERIMENTAL - DO NOT COMMIT TO PRODUCTION**~~

I'd appreciate your opinion on this.

**I now think this is robust enough for production - provided you're happy with the concept**

It is only applied to the Places module and is built on the current code base.

It overcomes the problem where a number of facts are located at precisely the same location and consequently get clustered together (therefore the cluster bounds are a point) and clicking on one of those facts in the sidebar causes the map to zoom to the bounds of the cluster (effectively the map maxZoom level) - `markercluster.zoomToShowLayer()`) leaving you looking at a popup on a pretty much blank map (you may see a road if you're lucky)

This change introduces another module (written by the Markercluster author leaflet.markercluster.freezable - https://github.com/ghybs/Leaflet.MarkerCluster.Freezable) which allows us to disable clustering. The outcome is that on click, the clustered markers are unclustered but zooming is handles by external code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "leaflet-control-geocoder": "2.3.0",
     "leaflet.control.layers.tree": "1.1.0",
     "leaflet.markercluster": "1.5.3",
+    "leaflet.markercluster.freezable": "1.0.0",
     "moment": "2.29.1",
     "pinch-zoom-element": "1.1.1",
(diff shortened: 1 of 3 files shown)


ddrury:
Withdrawn pending re-write

PR #4197, opened 2022-01-31
Shift place hierarchy css

ddrury:
Missed during shift of map css to central files

### DIFF
--- a/resources/css/_list-places.css
+++ b/resources/css/_list-places.css
@@ -18,3 +18,15 @@
  *
  * wt-route-place-list
  */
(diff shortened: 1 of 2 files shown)


fisharebest:
The place hierarchy isn't a chart - it's a list.

There is already an (empty) CSS file for it - `resources/css/_list-places.css`

The CSS should be moved there.

fisharebest:
The place hierarchy isn't a chart - it's a list.

There is already an (empty) CSS file for it - `resources/css/_list-places.css`

The CSS should be moved there.

PR #4203, opened 2022-02-02
Add generation identifier & thumbnail image to Pedigree map sidebar

ddrury:
Makes pedigree map sidebar look similar to that in webtrees 1.7.
- Generation identifier is a coloured dot with the same colour as the linking line
example


fisharebest:
After each commit, I see that you merge all your commits and force-push.

(FYI, when I merge the PR, github gives me the option to merge all the commits at that point)

A consequence of this is that I cannot see what has just been changed - and so need to start my review from the beginning.

Also, any comments I make get lost, as they apply to commits which no longer exist.

You added a "dot" class, with inline styles.

I would prefer to avoid inlne styles (and am removing them where possible).  This is because they are harder to customise in themes, and cannot be used with a content-security-policy..

I just pushed a commit which adds classes that can be used instead.

Also, I'm not sure the `dot` class is needed.  Could you use `&bull;` instead?

ddrury:
> A consequence of this is that I cannot see what has just been changed - and so need to start my review from the beginning.

You've said this before, I apologise as I completely forgot about it. New commit using colour variables pushed.

BTW any change of reverting to the old colour order, lumping similar colours together often makes the contrast between generations difficult to see.

PR #4206, opened 2022-02-05
Change method access modifiers for easier extensibility of module

ric2016:
Note: In my opinion there is little reason to have private methods in any of the module classes, so perhaps change this everywhere?

### DIFF
--- a/app/Module/IndividualFactsTabModule.php
+++ b/app/Module/IndividualFactsTabModule.php
@@ -186,7 +186,7 @@ class IndividualFactsTabModule extends AbstractModule implements ModuleTabInterf
      *
      * @return Collection<Fact>
      */
-    private function spouseFacts(Individual $individual, Individual $spouse, Date $min_date, Date $max_date): Collection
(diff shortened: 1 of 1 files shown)


fisharebest:
These functions are internal implementation details for the module.

IMHO, they should not be exposed, and they should not be relied on.

If my IDE tells me that a function isn't used, I may decide to remove/refactor it - which will break your code.

If the logic in these private functions is sufficiently useful that it will be used elsewhere, then perhaps it needs moving to its own `Service` (i.e. business logic) class?

ric2016:
So what if I want to slightly modify an existing module? Currently I have to re-implement all private methods, and align them with each change in the original module. How is that any better?

Having protected methods doesn't mean you're not allowed to change them - That happens all the time anyway.

ric2016:
Thank you - I appreciate it.

fisharebest:
> If the logic in these private functions is sufficiently useful that it will be used elsewhere, then perhaps it needs moving to its own Service (i.e. business logic) class?

I have just implemented this.

All these functions are now public methods in a new `IndividualFactsService`.

PR #4224, opened 2022-02-20
Bug fix - prevent page scrolling to top on map reset

ddrury:


ddrury:
Hm strange, it works for me on firefox, chrome & edge on a windows machine

fisharebest:
Ah.  In my test case, the lat/long fields were blank (a new location), which causes a JS error - which means that the `return false` is never reached...

ddrury:
> Ah. In my test case, the lat/long fields were blank (a new location), which causes a JS error

See https://github.com/fisharebest/webtrees/pull/4226

BTW the places tab is the best page to see the scroll effect when resetting the map.

PR #4226, opened 2022-02-21
Bug fix - Ensure marker gets valid lat/lng when using map reset control (version 2)

ddrury:


fisharebest:
We pass two sets of lat/long to the edit page.

`$latitude/$longitude` (may be null) - which initialise the edit fields.

`$location` (which contains not-null `latitude/longitude()` members) - which is used to initialise the marker location.

This allows the user to distinguish between "no coordinates set" and "coordinates set to 0,0".

If this fix is related to the "resetCallback" error, then perhaps the reset code should be using `$location` instead of `$latitude/$longitude`

ddrury:
Updated

PR #4236, opened 2022-02-26
Fix TomSelect virtual scroll

ric2016:
Do not display "Loading more results ..." if there aren't any (in particular in case of few results, or none at all). See below for current 'ugly' behavior (screenshot from demo server).
vs


fisharebest:
Thanks!

PR #4241, opened 2022-02-28
Rhode Island State Censuses

TheFirstCaptain:
New feature to add Rhode Island State Censuses from 1905, 1915, 1925. 

- Includes automated tests for all censuses
- The order in drop down places Rhode Island after US (others appear alphabetical, except first based on local)


fisharebest:
Thanks for your contribution!

PR #4246, opened 2022-03-05
Force footer to the bottom of the page regardless of content length

ddrury:


fisharebest:
We already have code to do this - https://github.com/fisharebest/webtrees/blob/2.1.0-alpha.2/resources/css/_base.css#L46-L55

But it appears to have stopped working a while back.  I guess the class `wt-main-container` must have moved from the `<main>` element to its child.

fisharebest:
We already have code to do this - https://github.com/fisharebest/webtrees/blob/2.1.0-alpha.2/resources/css/_base.css#L46-L55

But it appears to have stopped working a while back.  I guess the class `wt-main-container` must have moved from the `<main>` element to its child.

PR #4261, opened 2022-03-10
Change target url for FamilySearchFamilyTreeId

ric2016:
One final adjustment for FamilySearchFamilyTreeId: in my opinion, the actual detail page has more useful data than the shiny 'discovery page', which seems more targeted to casual users.

See e.g. [here](https://www.familysearch.org/tree/person/details/LK7V-Q4H) vs [here](https://ancestors.familysearch.org/LK7V-Q4H/).


fisharebest:
Thanks.  I don't use this service - I just used the URL that I was given.

PR #4298, opened 2022-04-09
Add more missing tags

ric2016:
[Subtag definition is already ok](https://github.com/fisharebest/webtrees/blob/5bbcfdd4dba43df94a937e2784d34c97de40a5f9/app/Gedcom.php#L1705).


fisharebest:
We already have definitions for `FAM:*:NOTE` and `FAM:*:SOUR:...` (lines 379 and 395-403)

So, I think we only need the `FAM:_ASSO` and `FAM:_ASSO:RELA`?

ric2016:
Yes, I missed that - I only saw the raw tag "FAM:_ASSO" in the dropdown on the family page and fixed that by adding "FAM:_ASSO".

Everything else was just copy & paste.

PR #4310, opened 2022-04-16
Fix replacement images

ric2016:
Replacement images aren't actually shown in browsers (tested with Firefox and Chrome) without this fix, because the intended content-type response header set via the headers array [is overwritten here](https://github.com/fisharebest/webtrees/blob/35e7ad0c309aa084344904fc9471036d7c215055/app/Factories/ResponseFactory.php#L109).

A more robust solution avoiding similar issues would be to check in ResponseFactory whether a content-type header is already set (case-insensitively!)

### DIFF
--- a/app/Factories/ImageFactory.php
+++ b/app/Factories/ImageFactory.php
@@ -324,9 +324,7 @@ class ImageFactory implements ImageFactoryInterface
         $svg = view('errors/image-svg', ['status' => $text]);
 
         // We can't send the actual status code, as browsers won't show images with 4xx/5xx.
-        return response($svg, StatusCodeInterface::STATUS_OK, [
-            'content-type' => 'image/svg+xml',
-        ]);
+        return response($svg, StatusCodeInterface::STATUS_OK)->withHeader('content-type', 'image/svg+xml');
     }
 
     /**


fisharebest:
> A more robust solution avoiding similar issues would be to check in ResponseFactory whether a content-type header is already set 

There are several other similar issues - so fixing ResponseFactory will be the correct action.

> (case-insensitively!)

We should only use lower case for headers

fisharebest:
> Replacement images aren't actually shown in browsers (tested with Firefox and Chrome)

I cannot reproduce this.

> the intended content-type response header set via the headers array [is overwritten here](https://github.com/fisharebest/webtrees/blob/35e7ad0c309aa084344904fc9471036d7c215055/app/Factories/ResponseFactory.php#L109).

Are you sure?

`$headers['content-type'] ??= 'text/html; charset=UTF-8';`

We are passing an array containing this key.  So the value is not overwritten?

ric2016:
I tested in beta 2. [This was actually fixed afterwards](https://github.com/fisharebest/webtrees/commit/6172e7f6cb6d9b42e9feef9b728f7561b0d70dc9).

Issue #4331, opened 2022-04-21
Error for Family Book - Future test system


melizaa:
I try to see the future test system Family book of Eva Grünblatt (I835).
I see: The parameter “spouses” is missing.

PR #4366, opened 2022-05-01
Validation of optional/boolean parameter on hourglass page

jon48:
Similarly to #4331 , the case when "Show spouses" is not ticked is not correctly handled, and the parameter validation fails.


fisharebest:
Thanks!

Issue #4363, opened 2022-05-01
Multi line census notes, link to expand the note incorrectly created


ddrury:
![demo](https://user-images.githubusercontent.com/104515/166140028-e4e4bce0-8470-49fe-9538-073dabce2604.jpg)

ddrury:
Function `Note::extractNames` needs to obtain the first line by using` "<br>"` as well as `"\n"` as the split characters.

miqrogroove:
I believe multi-line notes are always required to have two linefeeds after the title.  And I doubt this has changed in a long time.  Have you tried that?

ddrury:
Yeah, I've found the problem but there are a number of ways to skin this particular cat so I'll let Greg decide on what to do (see my second comment above)

fisharebest:
The common mark library uses `<br />` for "hard breaks", and a configurable value for soft breaks.
We've been using `<br>`.

Changing our config to `<br />` will simplify the processing.

PR #4367, opened 2022-05-01
Fix extraction of first line of multiline notes

ddrury:
Split on `<br>` as well as newline. Solves https://github.com/fisharebest/webtrees/issues/4363


fisharebest:
I've just pushed a slightly simpler fix for #4363 - which I think fixes this issue.

PR #4413, opened 2022-05-18
Update styling for v2.1 restriction lists

miqrogroove:
The class `form-control` should now be `form-select`


fisharebest:
Thanks!

I see there are many of these.  I'll do a search/replace for the rest.

Issue #333, opened 2014-12-22
Function getActiveModules sorts custom modules by its English title.


wooc:
It should sort all the modules by the translated title.

PR #4432, opened 2022-05-25
Reactivate logging of search queries

Bogiegit:
Could you please reactivate the search query logging feature with 2 additions;
- only for not logged in visitors
- with a configuration switch in the family tree settings that is set to off by default
That meets the requirement of quite a few users and does not change the current way it works.


Bogiegit:
html,body{background-color:#fff;color:#333;line-height:1.4;font-family:sans-serif,Arial,Verdana,Trebuchet MS;}Thank you, Greg! Looking forward to that feature.
Regards,
 
Bogie

https://ourtree.thespiegels.com

Greg Roach schrieb am 25.11.2022 12:38 (GMT +01:00):

Closed #4432 via  45ee34f.

PR #4459, opened 2022-06-04
Fix #4282.

### LINKED ISSUE #4282: 2.1 B1 delete confirmation popup of BIRT / CHR event is malformed
In the delete confirmation popup, the age of the parents is not displayed after the event date, but in separate lines

grafik

ric2016:
See issue #4282.

### DIFF
--- a/app/Fact.php
+++ b/app/Fact.php
@@ -589,7 +589,7 @@ class Fact
      *
      * @return string
      */
-    public function summary(): string
+    public function summary($neverIncludeParentsAge = false): string
     {
         $attributes = [];
         $target     = $this->target();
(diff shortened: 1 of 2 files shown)


ric2016:
I consider this to be a simple fix. Is there anything that keeps it from getting accepted? I'd appreciate some feedback.

fisharebest:
> I consider this to be a simple fix. Is there anything that keeps it from getting accepted? I'd appreciate some feedback.

Sorry - been busy.

I think the original `summary()` function out to be a view.  It creates formatted HTML, and doesn't belong in the `Fact` class.

For the "delete confirmation" text, we're creating formatted HTML, then stripping it off.   Since we just need the fact's type/value/date/place and since we only need it in one place, I think it should probably go inline here.

fisharebest:
We already have a function to create the required text: `Fact::name()`.
I had forgotten this function exists...

PR #4460, opened 2022-06-05
Make tables fixed width

ddrury:
For the individual Notes, Media & Sources tabs, make the containing table fixed width (like the facts tab) and used css rather than embedded styles.

Fixes problem with content overflowing the container


fisharebest:
> Fixes problem with content overflowing the container

Which problems?

> like the facts tab

I'm not sure why this is fixed.  

Taking it out and using the default table-layout seems to make better use of space.

ddrury:
> Which problems?

Anything that exceeded the width of the containing element, for example wide census tables
The facts and events tab had already fixed this but used inline styling rather than css. I just changes that and copied the code to the other affected tabs. 

If you're saying that this problem is now fixed without this bit of code then fine but my gut says it isn't.

PR #4578, opened 2022-10-16
Fix `Invalid GEDCOM data passed to Fact` when no required new facts

jon48:
Whilst testing on one of my development environments, I stumbled across the elusive `Invalid GEDCOM data passed to Fact::_construct(1,)` error reported occasionally on the forum.

In my case, I managed to pin it down to the fact that I had no "Facts for new families" set in that environment.

image

From the code perspective, that means that in `Services\GedcomEditService::newFamilyFacts`, when performing `explode` on `$tree->getPreference('QUICK_REQUIRED_FAMFACTS')`, the result is an array with the empty string `[""]`. In the subsequent code, it therefore tries to create a new fact with an empty tag, which is not allowed in `Fact::construct`.

The issue can be reproduced for `Services\GedcomEditServicenewIndividualFacts` as well, by removing the "Facts for new individuals" entries.

In this PR, I am basically filtering out empty strings, but feel free to implement it any other way you prefer.

### DIFF
--- a/app/Services/GedcomEditService.php
+++ b/app/Services/GedcomEditService.php
@@ -30,6 +30,7 @@ use Fisharebest\Webtrees\Registry;
 use Fisharebest\Webtrees\Site;
 use Fisharebest\Webtrees\Tree;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Str;
 
 use function array_diff;
 use function array_filter;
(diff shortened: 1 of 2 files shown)


jon48:
Added some unit tests around the new facts methods

fisharebest:
Thanks for this.

I think `Str::length($tag) > 0` might be simplified to `$tag !== ''`.

I'll merge it and change it.

Issue #4397, opened 2022-05-12
It's no longer possible to add a Note or a Shared Note to a Research task.


Manja-S:
In webtrees 2.1.2 it isn't possible to add a note or a shared note in the edit-screen of a new or an existing Research task, there's no input field for that. Currently the only way to add a note or a shared note to a new or an existing Research task is by editing the raw GEDCOM-record of the person. Changes to the raw GEDCOM-record of the research task are not saved.

(first mentioned in #4391)

PR #4585, opened 2022-10-23
Suggestion: allow autoformatting on ResearchTask element

jon48:
This is a suggestion to allow autoformatting on _TODO elements, if there is no side effects I have not anticipated.

As its purpose is close to a note, I think it would be beneficial to allow for more advanced formatting than just plain text (to display lists, URL for instance).

### DIFF
--- a/app/Elements/ResearchTask.php
+++ b/app/Elements/ResearchTask.php
@@ -57,4 +57,17 @@ class ResearchTask extends AbstractElement
     {
         return $this->editTextArea($id, $name, $value);
     }
+
+    /**
+     * Display the value of this type of element.
+     *
+     * @param string $value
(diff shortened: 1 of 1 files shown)


fisharebest:
I see the value of the `_TODO` field as a title/summary.
We display this in a table column on the home page.

If we allow tables in this field, then I think it is likely to break the layout.

#4397 allows us to add notes to research tasks - which can contain markdown formatting.

jon48:
Indeed, that makes more sense, and is consistent with the way tasks were recorded pre-2.1, I should have been a bit more thorough with my suggestion. I am therefore closing the PR.
Arguably, maybe the task "title" should not be a text area, but a simple text input, to avoid any confusion.

PR #4602, opened 2022-11-16
Add fullscreen button to map

ddrury:
Use browser [Fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) but use fscreen library to provide Vendor agnostic access


ddrury:
Greg,

This pr has got messy, I'm going to close it and resubmit two or three separate PRs which hopefully will be a lot clearer

PR #4605, opened 2022-11-22
Update reset control on map

ddrury:
1. Use view for control icon
2. Exclusively use setAttribute as assigning aria attributes with simple = sign does not work
3. Remove invalid and redundant attributes

### DIFF
--- a/app/Services/LeafletJsService.php
+++ b/app/Services/LeafletJsService.php
@@ -81,6 +81,7 @@ class LeafletJsService
             'icons'        => [
                 'collapse' => view('icons/collapse'),
                 'expand'   => view('icons/expand'),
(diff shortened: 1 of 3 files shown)


fisharebest:
I merged this, and made a few changes.

> Exclusively use setAttribute as assigning aria attributes with simple = sign does not work

aria-label doesn't yet work with firefox - it's a known bug - https://caniuse.com/?search=arialabel

Using `setAttribute('aria-label')` is a workaround until firefox is fixed.

Font-awesome provides icons for "redo" (clockwise) and "undo" (anti-clockwise).
I think "undo" would be better here.

> Remove invalid and redundant attributes

If you remove `href="#"`, you no longer get the "pointer cursor" when you move the mouse over the element.

PR #4606, opened 2022-11-22
Create general purpose fullscreen function...

ddrury:
…  and add fullscreen control to maps, now includes sidebar as discussed on forum

NOTE: works best if https://github.com/fisharebest/webtrees/pull/4605 is applied first


fisharebest:
I have merged this PR, and also made a few small changes.  e.g. separate expand/collapse icons and text.

Issue #4635, opened 2022-12-04
After auto-upgrade: nothing works


someone-somenet-org:
Just got the email that a new version is available.

After "successfully" upgrading nothing works and all pages only show this stack trace

```
[...]

PR #4636, opened 2022-12-04
Fix for #4635

someone-somenet-org:
This seems to fix my issue. #4635


someone-somenet-org:
Maybe there is another issue waiting to happen on line?: https://github.com/fisharebest/webtrees/blob/9d15b78eafc58b7682cca0e418e929d600dc2fb8/app/Http/Middleware/BadBotBlocker.php#L325

another catch without a target-variable. shall i fix that too? (dunno now to trigger/test)

someone-somenet-org:
also maybe you want to stop distributing the broken update, which breaks installations. :)

fisharebest:
> Maybe there is another issue waiting to happen on line?: 

This is the `main` branch.  This requires PHP 8.0 - 8.2

The `2.1` branch runs on PHP 7.4 - 8.2

PR #4640, opened 2022-12-05
2.1.8.mod

FrankWarius:


resources/views/robots-txt.phtml
```diff
@@ -27,7 +27,27 @@ Disallow: <?= $base_path ?>/admin
 Disallow: <?= $base_path ?>/manager
 Disallow: <?= $base_path ?>/editor
 Disallow: <?= $base_path ?>/account
-Crawl-delay: 10
+Crawl-delay: 1
+
+Disallow: <?= $base_path ?>/login   
+Disallow: <?= $base_path ?>/module 
+
+Disallow: /*/*-list
```
fisharebest, line 35: You cannot use `*` here.   See http://www.robotstxt.org/robotstxt.html

> Note also that globbing and regular expression are not supported in either the User-agent or Disallow lines. The '*' in the User-agent field is a special value meaning "any robot". Specifically, you cannot have lines like "User-agent: *bot*", "Disallow: /tmp/*" or "Disallow: *.gif".

resources/views/robots-txt.phtml
```diff
@@ -27,7 +27,27 @@ Disallow: <?= $base_path ?>/admin
 Disallow: <?= $base_path ?>/manager
 Disallow: <?= $base_path ?>/editor
 Disallow: <?= $base_path ?>/account
-Crawl-delay: 10
+Crawl-delay: 1
+
+Disallow: <?= $base_path ?>/login   
+Disallow: <?= $base_path ?>/module
```
fisharebest, line 33: This will block access to the sitemap module.  We want robots to view it.

FrankWarius:
sorry this was not intended, I please reject.

I got a bit confused by the frequent release changes

resources/views/robots-txt.phtml
FrankWarius: this works for (only?) pretty urls since 3 month

resources/views/robots-txt.phtml
FrankWarius: I made this changes in robots.txt 3 month ago in my installation and it's works.

Google recommends using Disallow in robots.txt instead of Not Follow,

I'll open a new threat for Google index in open discussion later this week

fisharebest:
> sorry this was not intended, please reject.

No problem.  I guessed this was not intentional :-)

PR #4651, opened 2022-12-07
Fix: interactive tree unable to expand on drag when in fullscreen mode

ddrury:


fisharebest:
Reverse-engineering your change, I see that we need to target the chart's parent - not the chart itself.

`wt-tab-tree` belongs only on the tab - and this chart has a full-page version as well.

So, I created a parent element with appropriate classes, and used that at the full-screen target instead.

And FYI, bootstrap provide a handy `h-100` class, which simplifies the change further.

Thanks for investigating, and identifying the underlying cause of the issue.

photon-flip:
Fantastic work gentlemen. I look forward to trying it out. 
Ultimately I can see this being added to all the charts.
Thanks again.

PR #4660, opened 2022-12-13
GedcomRecord format: Replace HTML in class by view

jon48:
This is a proposal to replace the HTML hardcoded in the `GedcomRecord` class for the `formatList` by a view.
I think this is going in the global direction of managing HTML output by views (and overwriting in subclasses is still possible), and more selfishly, that would allow modules to overwrite the display of this element.

As usual, this is only a proposal, feel free to close it or amend it.

### DIFF
--- a/app/GedcomRecord.php
+++ b/app/GedcomRecord.php
@@ -501,12 +501,7 @@ class GedcomRecord
      */
     public function formatList(): string
     {
-        $html = '<a href="' . e($this->url()) . '">';
-        $html .= '<b>' . $this->fullName() . '</b>';
-        $html .= '</a>';
-        $html .= $this->formatListDetails();
-
-        return $html;
+        return view('record-list-info', ['record' => $this]);
     }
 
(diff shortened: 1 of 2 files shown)


fisharebest:
As far as I can tell, this function is only used by the statistics.

Since it doesn't use any private functions/data of the `GedcomRecord` object, I don't think it belongs here.

Are you using it in your own code/modules?

jon48:
I am not calling the method itself in my modules. The only interaction I have with it in in my fork, where I am modifying its body to inject some icons. To improve the compabitility of my modules with the mainstream webtrees, I am looking to move away from those core code modifications, and replace it with views overwrites, and this piece of code is the only one I cannot overwrite (that is why this PR is quite selfish…).

But I agree with you, the method probably does not belong to that class (likely a legacy to old times when a lot of formatting was done in those model classes), and feel free to refactor as you wish, I will adapt to the outcome.

PR #4664, opened 2022-12-15
Add sort by BURI:PLAC to Burial(cemetery_report)

d4n5:
Burial report (cemetery_report)
In addition to optional sorting by name or date, adding place of burial


fisharebest:
I don't think this actually works.

There isn't anything in the code that that will sort by `BURI:PLAC`.

Additional code would be required here:
https://github.com/fisharebest/webtrees/blob/736b8b72ed0af202b700901a7965eb77c180581f/app/Report/ReportParserGenerate.php#L2259-L2278

Issue #4667, opened 2022-12-16
2.2 Dev Sqlsrv SeedUserTable  SQLSTATE[23000]: insert into [wt_user] IDENTITY_INSERTis OFF


FrankWarius:
good job so far - almost done!
#0 D:\GitRepos\fisharebest\webtrees\vendor\illuminate\database\Connection.php(720): Illuminate\Database\Connection->runQueryCallback('insert into [wt...', Array, Object(Closure))
#1 D:\GitRepos\fisharebest\webtrees\vendor\illuminate\database\Connection.php(546): Illuminate\Database\Connection->run('insert into [wt...', Array, Object(Closure))
#2 D:\GitRepos\fisharebest\webtrees\vendor\illuminate\database\Connection.php(498): Illuminate\Database\Connection->statement('insert into [wt...', Array)
#3 D:\GitRepos\fisharebest\webtrees\vendor\illuminate\database\Query\Builder.php(3218): Illuminate\Database\Connection->insert('insert into [wt...', Array)
#4 D:\GitRepos\fisharebest\webtrees\vendor\illuminate\database\Query\Builder.php(3336): Illuminate\Database\Query\Builder->insert(Array)
#5 D:\GitRepos\fisharebest\webtrees\app\Schema\SeedUserTable.php(40): Illuminate\Database\Query\Builder->updateOrInsert(Array, Array)
#6 D:\GitRepos\fisharebest\webtrees\app\Services\MigrationService.php(117): Fisharebest\Webtrees\Schema\SeedUserTable->run()
#7 D:\GitRepos\fisharebest\webtrees\app\Http\RequestHandlers\SetupWizard.php(399): Fisharebest\Webtrees\Services\MigrationService->seedDatabase()

FrankWarius:
I found a fix and Config.ini,php is created, but now  get 
SQLSTATE[42000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Die implizite Konvertierung vom nvarchar-Datentyp in varbinary(max) ist nicht zulässig. Verwenden Sie die CONVERT-Funktion, um diese Abfrage auszuführen. (SQL: insert into [wt_session] ([session_id], [session_time], [user_id], [ip_address], [session_data]) values (r12cqpbo56d7cvssdj8t3mpbsd, 2022-12-16 22:47:04, 0, 127.0.0.1, initiated|b:1;))

[...]

PR #4668, opened 2022-12-17
#4667 Fix Sqlsrv Setup

FrankWarius:


fisharebest:
Hi Frank,

Can you run some tests for me.

1) You set `IDENTITY_INSERT OFF` after each insert.

We didn't do this in the previous code (But that code did not work...!).

Can you check that this statement is required.  e.g. will it work without it.

2) In SQL-Server, I think that you use `[` and `]` to quote table names.  This can be important if the prefix contains special characters.

Can you tell me if it still works when you replace
`'SET IDENTITY_INSERT ' . $prefix . 'user ON'`
with
`'SET IDENTITY_INSERT [' . $prefix . 'user] ON'`

FrankWarius:
Hi Greg

> 1. You set IDENTITY_INSERT OFF after each insert.

I followed https://stackoverflow.com/questions/56560556/insert-data-in-laravel-identity-insert-is-set-to-off.
it seems You have to switch it On and OFF for every table.
Your actual fixes work for all migration scripts

> 2. In SQL-Server, I think that you use [ and ] to quote table names. This can be important if the prefix contains special characters.

Done and pushed.

Migration finished with creation of Config.ini.php but has an major issue with session table see #4669

fisharebest:
Can you tell me if it is necessary to use `->unprepared()`.
Will it also work wtih `->statement()`?

fisharebest:
Your PR adds some entries to `.gitignore`

It is recommended that this type of file are added to your global .gitignore file - not the project file.

I merged the `SET IDENTITY_INSERT` changes in 736b8b72ed0af202b700901a7965eb77c180581f

FrankWarius:
> Can you tell me if it is necessary to use `->unprepared()`. Will it also work wtih `->statement()`?

I doesn't test statement(). I found https://stackoverflow.com/questions/56560556/insert-data-in-laravel-identity-insert-is-set-to-off. and there You found

> You can use DB::unprepared() instead of DB::statement().
> The difference is that statement() executes a prepared statement, while unprepared() uses PDO::exec() instead.
>
> Every prepared statement starts a new session (inside the same database connection) and because the SET IDENTITY_INSERT query is session-specific, it doesn't affect other prepared statements like the INSERT query.

PR #4691, opened 2022-12-28
Use class instead of hardcoded style in read-only name input

jon48:
Commit 62cb0f58 introduces hardcoded styling to indicate that the full name input is read-only.
This is however relying on the standard Bootstrap color scheme, which custom themes may not be using (I am using a different color scheme in the Rural theme), and creates an inconsistent background color.

image

I am suggesting to switch to the use of the `disabled` attribute, which is styled in Bootstrap, and allow for the themes to apply its color scheme, like below:

_Read-only_
image

_Editable_
image

I am pushing the PR to `main`, but that would need to be backported to the `2.1` branch as well.

### DIFF
--- a/app/Elements/NamePersonal.php
+++ b/app/Elements/NamePersonal.php
@@ -148,7 +148,8 @@ class NamePersonal extends AbstractElement
         return
             '<div class="input-group">' .
             view('edit/input-addon-edit-name', ['id' => $id]) .
-            '<input class="form-control" style="background-color:#e9ecef" type="text" id="' . e($id) . '" name="' . e($name) . '" value="' . e($value) . '" readonly="readonly" />' .
+            '<input class="form-control" type="text" id="' . e($id) . '-disabled" name="' . e($name) . '" value="' . e($value) . '" readonly="readonly" disabled="disabled" />' .
+            '<input class="form-control d-none" type="text" id="' . e($id) . '" name="' . e($name) . '" value="' . e($value) . '" />' .
             view('edit/input-addon-keyboard', ['id' => $id]) .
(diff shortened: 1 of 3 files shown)


jon48:
And by the way, the `disabled` attribute is already correctly styled with `#e9ecef` in the standard themes.

image

fisharebest:
I tried using `disabled` originally.

The problem is that disabled form fields are not sent when the form is submitted.

Each field sends three values `level[]`, `tag[]`, and `value[]`, and the server receives three arrays.

If we do not send one of the `value[]` fields (because it is disabled), then that array has one fewer values.

The result is that the values do not align with levels/tags, and data becomes corrupted.

A solution may be to remove the `disabled` attribute on save - but I am nervous that a javascript error on the page would result in corrupt data.

jon48:
Oh, I understand now, I did not go as far as trying to save.

I have just force-pushed a similar approach but using a new class instead of the disabled attribute.
It it set to `#e9ecef` for standard themes, and I can use my own background color in the rural theme. It is less Bootstrap-integrated, but I feel it is better than the hardcoded style.

fisharebest:
I think your `disabled="disabled"` approach is better, for usability and accessibility./

> A solution may be to remove the disabled attribute on save - but I am nervous that a javascript error on the page would result in corrupt data.

We could have two fields.  One disabled, one hidden.
When the user clicks "edit", we can toggle the hidden atribute on both fields

Now, we can be certain that there will always be exactly one disabled field and one enabled field.

jon48:
Could you review if my latest push is what you had in mind?
Another pair of eyes will clearly be helpful, so that I am not messing up anything with the IDs.

fisharebest:
Thanks Jon!

PR #4713, opened 2023-01-08
Improve code analysis PhpDoc for ModuleService & related

jon48:
Following the introduction of better PhpDoc in Laravel 9, there are a lot more PhpStan errors that have appeared in the `main` branch, and that is cascading down my custom modules.
This is a first attempt to improve some of the PhpDoc of the `ModuleService::findByInterface`, removing some of the baseline errors, and fixing some related issues.

I am not particularly happy with the way I had to add the return type to most of the return statement, but otherwise PhpStan seems to expect the exact class, and I do not see how this is incompatible with the more generic term. I will probably continue investigating that one.

```
Method Fisharebest\Webtrees\Services\ModuleService::findByInterface() should return Illuminate\Support\Collection<string, Fisharebest\Webtrees\Module\ModuleInterface&T> but returns Illuminate\Support\Collection<string, Fisharebest\Webtrees\Module\ModuleFooterInterface>.
```
Another point of attention is the slight change in `RelationshipsChartModule` / `RelationshipService`, as the following do not guarantee that a `ModuleLanguageInterface` will be found (it is probably OK, but the code does not guarantee it), so `$language` can be null. I have taken the approach of pushing back the problem to `RelationshipService`, and return an empty string, but other approaches could be taken, e.g. we could force the en-US language to be used...

```
$language = $this->module_service
            ->findByInterface(ModuleLanguageInterface::class, true)
            ->first(fn (ModuleLanguageInterface $language): bool => $language->locale()->languageTag() === I18N::languageTag());

### DIFF
--- a/app/Module/RelationshipsChartModule.php
+++ b/app/Module/RelationshipsChartModule.php
@@ -323,6 +323,7 @@ class RelationshipsChartModule extends AbstractModule implements ModuleChartInte
             $language = $this->module_service
                 ->findByInterface(ModuleLanguageInterface::class, true)
                 ->first(fn (ModuleLanguageInterface $language): bool => $language->locale()->languageTag() === I18N::languageTag());
+            assert($language instanceof ModuleLanguageInterface, 'The application guarantees this, but static code analysis cannot infer it');
 
             echo '<h3>', I18N::translate('Relationship: %s', $this->relationship_service->nameFromPath($nodes->all(), $language)), '</h3>';
             $num_paths++;
--- a/app/Module/RelationshipsChartModule.php
+++ b/app/Module/RelationshipsChartModule.php
@@ -582,7 +582,7 @@ class RelationshipsChartModule extends AbstractModule implements ModuleChartInte
     /**
      * Convert numeric values to strings
      *
-     * @return Closure
+     * @return Closure(int|string):string
      */
     private function stringMapper(): Closure
     {
--- a/app/Module/RelationshipsChartModule.php
+++ b/app/Module/RelationshipsChartModule.php
@@ -42,10 +42,23 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 
+use function array_map;
+use function asset;
 use function count;
+use function current;
+use function e;
+use function implode;
 use function in_array;
+use function max;
+use function min;
+use function next;
+use function ob_get_clean;
+use function ob_start;
+use function preg_match;
 use function redirect;
+use function response;
 use function route;
+use function sort;
 use function view;
 
 /**
@@ -71,20 +84,16 @@ class RelationshipsChartModule extends AbstractModule implements ModuleChartInte
         'recursion' => self::DEFAULT_RECURSION,
     ];
 
-    private ModuleService $module_service;
-
     private TreeService $tree_service;
 
     private RelationshipService $relationship_service;
 
     /**
-     * @param ModuleService       $module_service
      * @param RelationshipService $relationship_service
      * @param TreeService         $tree_service
      */
-    public function __construct(ModuleService $module_service, RelationshipService $relationship_service, TreeService $tree_service)
+    public function __construct(RelationshipService $relationship_service, TreeService $tree_service)
     {
-        $this->module_service       = $module_service;
         $this->relationship_service = $relationship_service;
         $this->tree_service         = $tree_service;
     }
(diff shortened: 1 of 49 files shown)

### CODE REVIEW

**Comment by fisharebest** [comment]
Thanks for this.  `string-class` is new to me.

----------

`@return Collection<int,ModuleSidebarInterface>` -> `@return Collection<string,ModuleSidebarInterface>`

Looking at this code, I see we do use the module name as the array key - so `string` might be correct...

But I don't think we *should*.  Module names are never important.  We always refer to them by class or interface.

In any case, PHP converts numeric strings to integers, and since the module name is user defined, this would need to be `int|string`.

I think we need to review the code to check that the array keys are never assumed to be strings, and then change them to ints here, by using `map` instead of `mapWithKeys`: https://github.com/fisharebest/webtrees/blob/486af7ec72f06ae7b16b65d9303ada6b8d6c5953/app/Services/ModuleService.php#L612-L614

----------

> as the following do not guarantee that a ModuleLanguageInterface will be found (it is probably OK, but the code does not guarantee it), so $language can be null

The code guarantees it - we are selecting a language from a list of languages - and the language came from that list.

If it doesn't this is a logic error, and I think we ought to trap it as soon as possible.

Maybe something like this???

`assert($language instance of ModuleLanguageInterface, 'The code guarantees this, but PHPSTAN cannot infer it');`

**Comment by jon48** [comment]
> I think we need to review the code to check that the array keys are never assumed to be strings, and then change them to ints here, by removing `mapWithKeys`:

I agree with you. we should have a consistent approach, and as you say the module_name as key is probably not necessary. I was just fixing the discrepancy that PhpStan was (correctly) highlighting, without giving much thoughts about the underlying issue.
I have removed that part of the PR, as I think this requires a dedicated PR.

> The code guarantees it - we are selecting a language from a list of languages - and the language came from that list.

Your statement is a bit misleading: on its own, the filter condition does not guarantee that any of the `ModuleLanguageInterface` matches the condition `$language->locale()->languageTag() === I18N::languageTag()`, so from the static code analysis `->first(...)` could *theoretically* return a `null` object.

https://github.com/fisharebest/webtrees/blob/8634febf6487fcb3ee7b59e88f828071a22cba22/app/Module/RelationshipsChartModule.php#L323-L325

But, having looked a bit further, I see where you are coming from, and indeed, the application as a whole would not have reached that point if no `ModuleLanguageInterface` had met that condition. I have therefore replaced my initial approach by an `assert` statement as you suggested.

Just to prove my point, there are actually the same lines in `I18N::init` L.306-308, and you can trigger an error with the case `I18N::init("mk")` on those specific lines. This is a bit of a theoretical exercice, as that would probably require a module to try and set that value, but this is just to show that PhpStan is not wrong in higlighting it. (And yes, I have mapped all the fisharebest\localization locales, all the `resources/lang` folders, and the `ModuleLanguageInterface` modules to find that case, which I believe is unique, as other missing locales would fail trying to load the translations file... 😅 )

**Comment by fisharebest** [comment]
I've found several places that use the string array keys.  I will need to update this code.

File: app/Module/RelationshipsChartModule.php [hunk 1]
```diff
@@ -323,6 +323,7 @@ public function chart(Individual $individual1, Individual $individual2, int $rec
             $language = $this->module_service
                 ->findByInterface(ModuleLanguageInterface::class, true)
                 ->first(fn (ModuleLanguageInterface $language): bool => $language->locale()->languageTag() === I18N::languageTag());
+            assert($language instanceof ModuleLanguageInterface, 'The application guarantees this, but static code analysis cannot infer it');
```
fisharebest, line 326: Looking into this, we can obtain `$language` much more easily: `I18N::language()`

fisharebest:
I think `T&ModuleInterface` can be simplified to just `T`, as `T` will always extend `ModuleInterface`...

jon48:
> I think `T&ModuleInterface` can be simplified to just `T`, as `T` will always extend `ModuleInterface`...

That is what I thought initially, and I had started documenting with `@template T of ModuleInterface`.

Unfortunately, in in the below, `MiddlewareInterface` does not implement `ModuleInterface`. The method `findByInterface` will still work fine, as it will find the modules that implement both (hence the return type `T&ModuleInterface`), but as an input -as it is -, we cannot assume much about the interface being given. 
 https://github.com/fisharebest/webtrees/blob/d11be7027e34e3121be11cc025421873364403f9/app/Http/Middleware/Router.php#L131

fisharebest:
> Unfortunately, in in the below, `MiddlewareInterface` does not implement `ModuleInterface`.

Good call.

I've just about finished with the int/string keys. I'll merge the other changes shortly.

jon48:
I will not be able to access my development environment until this weekend, but I will double check then.

With a quick glance at the commits, you seem to have covered even beyond my PR, so I am pretty sure that will be fine.

jon48:
I ran my library through the latest, looks good to me (I still have a lot of warnings I did not have in 2.1 / Laravel 8, but they are mainly due to limitations in Laravel documentation rather than webtrees).

As a note, on `Collections` you can simplify directly to `->first($callback)` instead of `filter($callback)->first()`

fisharebest:
Thanks for the tip on `->filter()->first()`.

The laravel libraries (Collection, Str) provide lots of clever functionality and we use only a very tiny part of it.
They are also contribute many MB to the releases.

I am thinking of replacing them with my own lightweight/simple versions, which contains just the functions we need.

This would also give use better control over the PHPdoc and static analysis.

PR #4743, opened 2023-01-29
Fix #4700 - "Show all" does not display all individuals in list

### LINKED ISSUE #4700: Show all individuals gives zero results
See: https://www.webtrees.net/index.php/en/forum/help-for-release-2-1-x/37602-list-of-individuals#96825

When I try to load all individuals, I get zero results. Tested this in webtrees 2.1.14 and 2.1.15. In webtrees 2.1.13 it works as expected.

Multiple users confirm this error.

jon48:
Most of my comments about it are already in the bug section.

In the PR, I am giving a slightly different version than in my last comment, but hopefully, it has the same behavior.

It is the first time I have dug into the `n_surn` / `n_surname` logic, and I hope I have not missing something fundamentals. Whilst testing, I noted some oddities around the "no surname" case, and case sensitivity, but I feel they go beyond that PR, and I will probably raise further issues to discuss those.

### DIFF
--- a/app/Module/IndividualListModule.php
+++ b/app/Module/IndividualListModule.php
@@ -456,14 +456,21 @@ class IndividualListModule extends AbstractModule implements ModuleListInterface
                         }
                     }
                     if ($show === 'indi') {
+                        /** @var array<string> $surnames */
+                        $surnames = collect($surns)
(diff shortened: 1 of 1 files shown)


arbor95:
there has recently (#4904) been a change to this module, so that your patch is no longer valid. Please try actual code and perhaps create a new pull request and close this one!

PR #4749, opened 2023-02-08
Fix for code missed when adding full screen functionality

ddrury:
Raised in forum message https://www.webtrees.net/index.php/en/forum/help-for-release-2-1-x/37790-geographic-data-full-screen

### DIFF
--- a/resources/css/administration.css
+++ b/resources/css/administration.css
@@ -61,6 +61,10 @@ h4 {
     height: 55vh;
 }
 
+:fullscreen .wt-location-edit-map {
+    height: 100% !important;
+}
+
 /* Raw GEDCOM data */
 .gedcom-data {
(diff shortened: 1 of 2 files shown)


resources/css/administration.css
```diff
@@ -61,6 +61,10 @@ h4 {
     height: 55vh;
 }
 
+:fullscreen .wt-location-edit-map {
+    height: 100% !important;
```
fisharebest, line 65: I don't think the `!important` is necessary (we are not trying to override another rule with a higher specificity).
It seems to work without it - unless I've missed something.

ddrury: copy & paste!!

fisharebest:
Merged.  Thanks!

PR #4783, opened 2023-03-11
Fix #3651 - Insensitive searches with PostgreSQL

### LINKED ISSUE #3651: Postgres - General search would be better if case insensitive
It seems that the General Search function is case sensitive, but I feel it would be more convenient if it wasn't. It would be nice to quickly type "laura" and have "Laura" records to show up. More of an issue was when I was searching for "Joann" and thought there was no record, when in fact the correct spelling of the name was "JoAnn" and the record did exist.

Thanks for Webtrees! Great to find a rebooted PHPGedView which I used years ago.

papjul:


papjul:
Let me know if you need clarifications about this pull request.

Currently, the search feature is almost unusable with PostgreSQL, because in France, we usually wrote surnames all in uppercase (so that we know the difference between John SMITH and Smith JOHN). However, people usually don’t write their search in uppercase (they will write "smith").

Regarding the implementation, as far as I know, this class is the only one where `LIKE` should be replaced by `ILIKE`, other classes should keep the `LIKE` operator, since insensitive searches are not required in the other cases. That’s why I implemented it as a member of the SearchService class.

I believe it should also be backported to version 2.1, but I want to make sure the current implementation is fine for you, before making an additional pull request to the 2.1 branch.

fisharebest:
> Let me know if you need clarifications about this pull request.

Please be patient.  I'm very busy right now.

> I believe it should also be backported to version 2.1

I will merge the changes into both branches.

papjul:
No worries, I didn't mean to rush you!
I just saw that you merged another pull request of mine within minutes so I thought something was wrong with this other pull request ;)

fisharebest:
There are a couple of issues with this (hence the delay in merging).

1) A more general solution is required that supports other RDBMS.  This will probably involve collations.  We could probably fix this for postgres with collations, but I don't have the opportunity to test other RDBMS right now.

2) The business logic functions `app/Services/...` are all pure.  i.e. they have no local variables/state.

The first can wait - I'm looking at a rewrite of the database handling.  For the second, I merged your code, and modified it to be pure.

Thanks for your contribution.

PR #4808, opened 2023-04-19
consolidate map size css

ddrury:
merge all into one file - renamed /resources/css/_leaflet.css to /resources/css/_maps.css

### DIFF
--- a/resources/views/admin/location-edit.phtml
+++ b/resources/views/admin/location-edit.phtml
@@ -29,7 +29,7 @@ use Fisharebest\Webtrees\View;
 <div class="row wt-location-edit-wrapper wt-fullscreen-container">
     <div id="wt-map" class="col-sm-9 wt-ajax-load wt-map" dir="ltr"></div>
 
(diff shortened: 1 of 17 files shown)


resources/views/admin/location-edit.phtml
```diff
@@ -27,9 +27,9 @@ use Fisharebest\Webtrees\View;
 <h1><?= $title ?></h1>
 
 <div class="row wt-location-edit-wrapper wt-fullscreen-container">
-    <div id="wt-map" class="col-sm-9 wt-ajax-load wt-location-edit-map" dir="ltr"></div>
+    <div id="wt-map" class="col-sm-9 wt-ajax-load wt-map" dir="ltr"></div>
 
-    <div class="col-sm-3">
+    <div class="col-sm-3 wt-sidebar wt-global">
```
fisharebest, line 32: We use `wt-global` on the `body` element (because it is generallly better to target classes than elemen).

I don't think we should use it elsewhere.

fisharebest:
Thanks!

Issue #4853, opened 2023-07-21
Bug: personal facts module


ddrury:
If the historic facts module has content and is enabled and an individual has no personal facts recorded then the message "There are no facts for this individual." should be displayed.

This doesn't happen because the historic events are included in the personal facts collection. 

See [this commit](https://github.com/fisharebest/webtrees/pull/4861/commits/d218cc45523e80f5c6421267a2e7594d7cbfa893)

fisharebest:
Presumably, the same is true of events of relatives/associates.

ddrury:
No, they are never merged with anything else, it's only the individual facts that get the others merged
```
        $individual_facts = $individual_facts
            ->merge($family_facts)
            ->merge($relative_facts)
            ->merge($associate_facts)
            ->merge($historic_facts);
```
I've made a simple fix in https://github.com/fisharebest/webtrees/pull/4854, just add this line before the merge 
` $has_individual_facts = $individual_facts->isNotEmpty()`;
then include it in the call to the view and test for it in the view

ddrury:
BTW maybe a good idea to merge into a more appropriate name ($all_facts?)

arbor95:
Has this been done or is it obsolete?

fisharebest:
The purpose of this text is simply to prevent a completely empty tab - which could be confusing.

> If the historic facts module has content and is enabled and an individual has no personal facts recorded

We only show historic events that occur between the first and last personal event with a date.

So there should never be historic events and no personal events.

If this is happening, then it should be a separate bug report.

PR #4854, opened 2023-07-21
Convert tab table to bootstrap grid

ddrury:
~~**Proof of concept - Do not use for production**~~

~~This only works on the webtrees theme for the Individual facts and events tab, (other things will break!!!)~~

I **think** this is suitable for production now

I've replaced the containing table with a bootstrap grid. This fixes all the problems with content overflowing the right margin and even looks reasonable when the browser width is reduced such that the columns fold.

It only requires two new webtrees classes, primarily for theming, everthing else is pure bootstrap.

I'd appreciate comments particularly on whether it's worth continuing down this path

BTW I've included a fix for https://github.com/fisharebest/webtrees/issues/4853


ddrury:
Closed - will resubmit more complete and simplified code

PR #4877, opened 2023-09-18
Add support for MySQL/MariaDB UNIX socket connection

jacrasmussen:
This changes enable the configuration of UNIX socket connections for MySQL/MariaDB.
It adds an additional configuration field for the socket path.

The change has been tested locally and on a running Webtrees instance using Webtrees 2.1.17.

### DIFF
--- a/app/Http/Middleware/UseDatabase.php
+++ b/app/Http/Middleware/UseDatabase.php
@@ -94,6 +94,7 @@ class UseDatabase implements MiddlewareInterface
             // For MySQL
             'charset'                 => $charset,
             'collation'               => $collation,
+            'unix_socket'             => Validator::attributes($request)->string('dbsocket'),
             'timezone'                => '+00:00',
             'engine'                  => 'InnoDB',
             'modes'                   => [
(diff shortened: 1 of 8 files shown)


fisharebest:
I do not think this change is necessary...

Laravel uses the PDO library, which will connect using a socket automatically when you specify `localhost` as the server.

I use this on my own server.  I disable the MySQL network protocol (`skip-networking`), and specify `localhost` as the server name.  It works!

As far as I can tell this PR would only be useful if your server is mis-configured, and the `php.ini` setting for ` pdo_mysql.default_socket` is incorrect.

jacrasmussen:
I appreciate the link and I confirm that it does work with only having localhost as dbname and the socket path in php.ini, my mistake was having the complete DSN as dbname at first as this is the way to go for some PHP applications.

I still think the change can be useful for users that prefers to configure the application through the setup wizard, but I'll be happy to drop the PR as well

miqrogroove:
I seriously doubt anyone is using socket path strings in their webtrees config.  Sounds like there is no traction from fisharebest.  Recommend closing the PR and, if needed, starting a forum thread.

fisharebest:
There has been confusion in the past between `localhost` and `127.0.01`.

Port is a required field, but not used with `localhost`.

I kept this open as a reminder to update the database-options page in the setup.

fisharebest:
I have re-worked the input form to make clear the difference between `localhost` and a network connection.

Screenshot 2023-11-04 at 18 04 45

Issue #4883, opened 2023-09-27
Privacy settings for GEDCOM exports are not correctly applied to certain facts in HEAD (e.g. SUBM)


Jefferson49:
Observed with webtrees 2.1.17

**Example A:** 
If a GEDCOM export is started with privacy settings less than manager (e.g. visitor), and the privacy for certain facts in HEAD (e.g. SUBM) is set to " Show to managers", the fact is still exported, e.g. SUBM is included in the GEDCOM export. 

Prerequesite to reproduce: 
Create a new tree

Privacy restrictions (for HEAD:SUBM, "Show to managers"):
![privacy_restrictions](https://github.com/fisharebest/webtrees/assets/81484983/b3fa0d23-1964-4e8a-a99c-8302ffe5908f)

Export privacy settings ("Visitor"):
![export preferences](https://github.com/fisharebest/webtrees/assets/81484983/dd0d79de-54d6-41eb-b226-f183ac3635ea)

**Example B:** 
If a GEDCOM export is started as "Manager" with the custom module [DownloadGedcomWithURL](https://github.com/Jefferson49/DownloadGedcomWithURL), and no "Manager'" is logged in, and the privacy settings for submittors is set to " Show to managers", the fact is **not** exported, e.g. SUBM is **not** included in the GEDCOM export. The reason is that - if no manager is logged in - the code in GedcomExportService.php does not apply the "Manager" setting to the export.

While debugging in the webtrees code, I recognized the following background:
+ The [createHeader method](https://github.com/fisharebest/webtrees/blob/main/app/Services/GedcomExportService.php#L291) in GedcomExportService.php does not consider the privacy access level of the export settings for certain facts in [line 330 of GedcomExportService.php](https://github.com/fisharebest/webtrees/blob/main/app/Services/GedcomExportService.php#L330)
+ The privatizeGedcom method in [line 239 GedcomExportService.php](https://github.com/fisharebest/webtrees/blob/main/app/Services/GedcomExportService.php#L239) is not called for HEAD, because HEAD is recognized as String rather than as GedcomRecord

Jefferson49:
Related pull request #4884 with a proposal for fix.

fisharebest:
`HEAD:SUBM` is a mandatory field.  Every valid GEDCOM has at least one `SUBM` record.

If we don't have one that we can safely include, then we will need to create an empty one.

Jefferson49:
[...]

PR #4884, opened 2023-09-27
Proposal for bugfix of issue #4883: Privacy settings for GEDCOM exports are not correctly applied to certain facts in HEAD (e.g. SUBM)

Jefferson49:


Jefferson49:
The proposal in the pull request did not consider that SUBM is mandatory and will be exported regardless of the privacy access level. Therefore, the proposed code does not work correctly.

Issue #4374, opened 2022-05-05
England and Wales Census for 1921


dzeidman:
As per the forum [thread](https://www.webtrees.net/index.php/en/forum/help-for-2-0/36412-census-assistant-for-1921-census-for-england-and-wales#91453), I would like to request that a census template be created for the England and Wales Census of 1921 which was released a few months ago. (At the time of writing the Scotland 1921 census has yet to be released).

I am attaching some sample documents. The census documents are two sided. However the one side is manly instructions. It also contains the name of the person entering the data and the postal address. I have only attached one example of this.
![Harris Hamburg and family on 49 De Burgh Street, Cardiff 2](https://user-images.githubusercontent.com/66416546/166883698-30b9a132-664b-4a7e-8cea-835355b14d8c.jpg)
![Harris Hamburg and family on 49 De Burgh Street, Cardiff](https://user-images.githubusercontent.com/66416546/166883710-b1cf9077-6037-41b1-8685-e4cf49a49733.jpg)

![Isaac and Bella (nee Zeidman) Guyster and family at 13 Amhurst Park, London](https://user-images.githubusercontent.com/66416546/166883715-74379969-317b-4340-ab7b-a24bca8a2d7d.jpg)

![Harriett Gabrielson nee Prager at 58 Oxford Road, Exeter](https://user-images.githubusercontent.com/66416546/166884332-7b5817e5-6769-456b-8d5b-ca76166fd09d.jpg)

![Morris and Clara (nee Zeidman) Lermon at 43 Plasturton Avenue, Cardiff](https://user-images.githubusercontent.com/66416546/166884399-6cdd0593-fe95-4a7b-ba90-f0b558f09eb0.jpg)

malcolmholt:
I would like to add an example of the 1921 Scotland census
![ScotlandsPeople_1921-census-example](https://user-images.githubusercontent.com/13886001/205514459-1231fce9-7f1b-4d11-ab57-c771d67e0f89.jpg)

ddrury:
I've attemted to create this, however the column entitled  **Number and ages of all living children and step children under 16 years of age** is defeating me, I tried creating a separate column for each age column but this results in a ridiculously wide table.

@fisharebest help please

RAMINCNZ:
@fisharebest and @ddrury 

Has this been further pushed along?? If I had better skills I would say I could help with the **Number and ages of all living children and step children under 16 years of age** issue..

Keep Smiling
Rick M

PR #4907, opened 2023-11-27
Add missing England 1921 Census

kiwi3685:
1921 Census file added, and CensusOfEngland changed to include it.

I noticed this census was missing for latest webtrees. Hope it's useful.

### DIFF
--- a/resources/lang/eu/messages.po
+++ b/resources/lang/eu/messages.po
@@ -3,7 +3,7 @@ msgstr ""
 "Project-Id-Version: webtrees\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2024-03-19 13:12+0000\n"
-"PO-Revision-Date: 2024-07-22 16:56+0000\n"
+"PO-Revision-Date: 2024-08-30 16:48+0000\n"
 "Last-Translator: Oskar-Txoperena <o.txoperena@gogotik.com>\n"
 "Language-Team: Basque https://translate.webtrees.net/projects/webtrees/"
 "webtrees/eu/>\n"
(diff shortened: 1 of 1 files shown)


ddrury:
There is an open issue for this - https://github.com/fisharebest/webtrees/issues/4374. I started producing a PR but the column(s) for "Number and ages of all living children ..." defeated me (seee attached image). It seems to me that this should be a number of columns not one as your PR proposes and I'm interested in how you'd visualise completing this column.
166883710-b1cf9077-6037-41b1-8685-e4cf49a49733

kiwi3685:
Adding 17 columns to make it exactly as per the original document is ridiculously impractical.

The single column in my pull request is for free-form text that allows uswrs to add the same informatuion in a simple way that suits their own preferences. The aim shoulld be to capture the content and intent of the original document, not slavishly follow every part of the form. Some typical entries for that column from my own data include:

'Total 3 - 1 3y; 2 4y"
'2<1y'

martins250:
Just hoping this can be included in the next webtrees version.

fisharebest:
> Just hoping this can be included in the next webtrees version.

Not until we can decide how to implement these columns.

In other censuses, we have tried to copy the exact structure of the original document.

But such a large number of columns may be impractical.

I guess I need to implement both and see how they work in practice.

kiwi3685:
Is it possible for a custom module to over-ride an existing census template, or even multiple ones?

ddrury:
> But such a large number of columns may be impractical.

The United states 1941 census has 31 columns which is more than the England 1921 census is likely to have even including the child ages columns and that doesn't seen to cause problems.

RAMINCNZ:
Hi There

Can we go ahead with this pull request any time soon?? And @kiwi3685  Maybe reopen this so we can get it done.

Keep Smiling

Rick M

PR #4924, opened 2023-12-23
Refactor Timestamp to avoid mistakes

### LINKED ISSUE #4923: "Changes Report" Defaults are Broken by Timestamp Typo
The default range for the Changes report is coded as `NOW-30` to `NOW`.

https://github.com/fisharebest/webtrees/blob/a609a185b4ae6c27a52180bb877790ca77672e68/resources/xml/reports/change_report.xml#L5-L6

If you click the Changes report in webtrees, you'll see this is broken.

Due a typo in the Timestamp class, the date argument is being converted to hours.

https://github.com/fisharebest/webtrees/blob/a609a185b4ae6c27a52180bb877790ca77672e68/app/Timestamp.php#L175-L179

It should also be considered that none of this logic is necessary.  Negative values are handled natively through the PHP `modify` method or the large variety of Carbon helper methods.

miqrogroove:
Fixes #4923

By implementing Carbon helper methods, we avoid some of the copypasta found in v2.1.0.

### DIFF
--- a/app/Timestamp.php
+++ b/app/Timestamp.php
@@ -20,7 +20,6 @@ declare(strict_types=1);
 namespace Fisharebest\Webtrees;
 
 use Carbon\Carbon;
-use DateInterval;
 use Fisharebest\Webtrees\Contracts\TimestampInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


fisharebest:
Thanks!

miqrogroove:
@fisharebest Thanks for the commit!  I will work on refreshing the other Timestamp patch for the bigger bugs.

PR #4927, opened 2023-12-27
Update checkbox-inline.phtml

magicsunday:
Currently the Checkbox component cannot be properly used to save empty (unchecked status) settings. A hidden element is missing here, which is assigned the unchecked value.

### DIFF
--- a/resources/views/components/checkbox-inline.phtml
+++ b/resources/views/components/checkbox-inline.phtml
@@ -9,11 +9,13 @@ declare(strict_types=1);
  * @var string      $label
  * @var string      $name
  * @var string|null $value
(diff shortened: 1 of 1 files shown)


fisharebest:
This is how checkboxes have always worked - although I remember being annoyed when I first discovered this.

I am concerned that this change might break existing code that simply checks for the presence of an input, rather than the actual value.

What is the motivation for this change.  What would be better/simpler with it?

The current code uses:

*  `Validator::boolean('name', false);` // returns true for checked and false for unchecked
*  `Validator::boolean('name');` // returns true for checked and throws an exception for unchecked

Issue #4682, opened 2022-12-24
2.2 Dev sqlsrv: date/time "2022-12-24 17:06:37.327" does not match pattern "Y-m-d H:i:s"


FrankWarius:
date/time "2022-12-24 17:06:37.327" does not match pattern "Y-m-d H:i:s" …\app\Factories\TimestampFactory.php:68
#0 …\app\Statistics\Repository\GedcomRepository.php(169): Fisharebest\Webtrees\Factories\TimestampFactory->fromString('2022-12-24 17:0...')
#1 …\app\Statistics.php(273): Fisharebest\Webtrees\Statistics\Repository\GedcomRepository->gedcomUpdated()
#2 …\app\Statistics.php(2845): Fisharebest\Webtrees\Statistics->gedcomUpdated()
#3 …\app\Statistics.php(214): Fisharebest\Webtrees\Statistics->getTags('\r\n    <p class=...')
#4 …\app\Module\FamilyTreeStatisticsModule.php(189): Fisharebest\Webtrees\Statistics->embedTags('\r\n    <p class=...')
#5 …\app\Http\RequestHandlers\TreePageBlock.php(66): Fisharebest\Webtrees\Module\FamilyTreeStatisticsModule->getBlock(Object(Fisharebest\Webtrees\Tree), 23, 'tree')

doesn't work either with `"Y-M-d H:m:s"`

FrankWarius:
public function fromString(?string $string, string $format = 'Y-m-d H:i:s.v', UserInterface $user = null): TimestampInterface
works for sqlsrv

fisharebest:
The database migrations create timestamp columns with a default precision of `0`.

So, I would expect the column `wt_change.change_time` to contain integer seconds.

But your table contains milliseconds.

Question - did you do anything to change this column, or was it created by the migration?

FrankWarius:
it is a blank out of the (2.2 dev) box installation. After that I exported my 2.1 Gedcom and imported it to 2.2 and did one source change (successful).

FrankWarius:
it is already logged by the SetupWizzard if the database is MS SQL.

log_id	log_time	log_type	log_message	ip_address	user_id	gedcom_id
1	2024-01-01 12:08:05.553	config	Site preference "WT_SCHEMA_VERSION" set to "1"	::1	NULL	NULL
2	2024-01-01 12:08:05.563	config	Site preference "WT_SCHEMA_VERSION" set to "2"	::1	NULL	NULL
[...]

PR #4931, opened 2023-12-31
Timestamp class needed local input and UTC output modes.

### LINKED ISSUE #4549: Changes Log Sometimes Doesn't Work
I've noticed sometimes none of my edits get logged.  I can't see any pattern to what shows up in the Changes log and what doesn't.

As an experiment, I just deleted the entire log, added a Census event, performed some raw GEDCOM edits, and then went back to look at the Changes log.  It's still blank.  So I'm definitely not imagining the randomness of all this.

miqrogroove:
Fixes #4549 

Clarifies the existing fromString() factory is only appropriate for UTC input.

Provides the fromLocalString() factory for user input, such as log filters.

Provides fromZoneString() factory for de-duplication and abstraction of the other two factories.

Provides Timestamp::toUTCDateTimeString() for SQL output, so timestamps won't be converted to local time and compared to stored UTC values.

Updates the PendingChangesService and SiteLogsService to use the correct times and timezones.

### DIFF
--- a/app/Contracts/TimestampFactoryInterface.php
+++ b/app/Contracts/TimestampFactoryInterface.php
@@ -33,6 +33,8 @@ interface TimestampFactoryInterface
     public function make(int $timestamp, UserInterface $user = null): TimestampInterface;
 
     /**
+     * Constructs a Timestamp using UTC input only.
+     *
      * @param string|null        $string YYYY-MM-DD HH:MM:SS (as provided by SQL).
      * @param string             $format
      * @param UserInterface|null $user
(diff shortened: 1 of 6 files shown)


fisharebest:
@miqrogroove - thanks for this.  I think you have clearly identified the missing functionality.

But there's a few changes that could simplify it a bit.

We have two ways to display timestamps.  `format/isoFormat()` and `toDateTime()`.  The first is for display, and should always be converted to the user's timezone.  The second is for creating database queries, and should always be in `UTC`.

Let me look at the code again, and I'll make some specific proposals.

miqrogroove:
> We have two ways to display timestamps. `format/isoFormat()` and `toDateTime()`. The first is for display, and should always be converted to the user's timezone. The second is for creating database queries, and should always be in `UTC`.

This would be an interesting design concept for toDateTime(), however that criteria is currently invalid.  The only reason toDateTime outputs UTC dates is because the factory treats all inputs as UTC.  If you don't want to use the Carbon timezone logic, then the inputs would need proper conversion prior to using the factory.  So we need to decide if that's happening inside the factory (as in my patch) or not.

FrankWarius:
when redesigning TImeStamp keep in mind the difference between SQL and TSQL see https://learn.microsoft.com/en-us/sql/t-sql/functions/date-and-time-data-types-and-functions-transact-sql?view=sql-server-ver16

TSQL DateTime has a (non changeable) precision of 3 digits / SQL 0
TSQL DateTime2 has a precision of 8 digits / SQL TimeStamp 6

see also https://github.com/fisharebest/webtrees/issues/4682

miqrogroove:
Sounds reasonable.  I would suggest avoiding usage of `DateInterval` because it's unnecessary here.  The native `modify()` method is much more concise.

miqrogroove:
Resolved by https://github.com/fisharebest/webtrees/commit/c9024634792322db5ca60d2b0a40ed045fe9aeb1

PR #4937, opened 2024-01-07
Fix error when non-admins save tree preferences

nomis:
Managers can edit most of the tree preferences but the "gedcom" parameter isn't submitted unless the user is an admin resulting in the error message:
```
The parameter “gedcom” is missing.
```
Add a hidden field with the missing "gedcom" parameter.

### DIFF
--- a/resources/views/admin/trees-preferences.phtml
+++ b/resources/views/admin/trees-preferences.phtml
@@ -87,6 +87,8 @@ use Illuminate\Support\Collection;
                 </div>
             </div>
         </div>
+    <?php else : ?>
+        <input type="hidden" name="gedcom" value="<?= e($tree->name()) ?>">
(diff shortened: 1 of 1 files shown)


fisharebest:
I'm not sure this is the correct solution.

We do not want managers to change this parameter (or the media folder parameter).

By adding a hidden field, users will be change the content (e.g. using the browsers developer tools to set the value).

Instead, we need the request-handler to ignore this parameter for non-admin users.

nomis:
> By adding a hidden field, users will be change the content (e.g. using the browsers developer tools to set the value).

Adding a hidden field doesn't allow users to submit different content, they can already do that.

fisharebest:
OK - I see now.  There is also the check on the server side.  (Sorry!)

Issue #4948, opened 2024-02-03
CKEditor strips some bootstrap classes


photon-flip:
For HTML blocks with CKEditor, when adding bootstrap classes to HTML in the source mode, most work fine, but for reasons I can't figure out, some are striped out.
Examples are `<h1 class="display-1">` and `<ul class="list-unstyled">`
If I use `<ul style="list-style-type:none">` that works ok.

Strangely they work at first save but if you return to the editor or just switch in and out of source mode the class is striped out.
I doubt that these are the only classes that are affected.
I can't see any reasons these would be a problem for HTMLPurifier.
Is it a CKEditor config problem?

photon-flip:
I found a really instant fix is to add` config.allowedContent = true` to the `resources\views\modules\ckeditor\ckeditor-js.phtml`
see: `https://ckeditor.com/docs/ckeditor4/latest/examples/acf.html#editor1`
but is this too liberal?

photon-flip:
Digging a little deeper, [CKEditor documentation](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_best_practices.html#configure-content-filtering) gives this advice:
Configure content filtering
>Advanced Content Filter is a cool feature which assures that the content inserted by the user blends in with your website in the best possible way. If you have any issues with it, do not turn it off, but instead configure it to match your requirements.

>For example, to enable <iframe> and <style> elements, allow two additional attributes (id and rel given in square brackets) for all (* is a wildcard) already enabled elements, and allow usage of any class names (see (*)) for them, use the following code:
`config.extraAllowedContent = 'iframe style;*[id,rel](*)';
`
I've tried that without iframe `config.extraAllowedContent = 'style;*[id,rel](*)';
`
seems bootstrap classes are sticking ok.
Would this be an acceptable fix?

photon-flip:
Looking more carefully at this, the webtrees `ckeditor-js.phtml` uses `CKEDITOR.config.extraAllowedContent =*[class,style]` which is intended to allow any element to use classes and styles but this is not doing that.
I believe that `*(*)` is required to achieve that.
>CKEditor documentation:
```
elements [attributes]{styles}(classes)
Where:
```
```
elements – a list of space-separated element names or an asterisk (*) character,
attributes – a comma-separated list of attribute names or an asterisk (*) character,
styles – a comma-separated list of style names or an asterisk (*) character,
classes – a comma-separated list of classes or an asterisk (*) character.
```
Should I set this up and make pull request?

PR #4949, opened 2024-02-04
Fixes: #4948 - Update ckeditor-js.phtml extraAllowedContent

photon-flip:
Allow CKEditor to accept all element classes.
The current webtrees `ckeditor-js.phtml CKEDITOR.config.extraAllowedContent ='*[class,style];' `appears intended to do this but is not working consistently. See issue #4948 CKEditor strips some bootstrap and other classes. 
This also facilitates using classes added in the CSS and JS  module.

### DIFF
--- a/resources/views/modules/ckeditor/ckeditor-js.phtml
+++ b/resources/views/modules/ckeditor/ckeditor-js.phtml
@@ -25,7 +25,7 @@ declare(strict_types=1);
         // Disable toolbars
         CKEDITOR.config.language = <?= json_encode(strtolower($language), JSON_THROW_ON_ERROR) ?>;
         CKEDITOR.config.removePlugins = 'forms,newpage,preview,print,save,templates,flash,iframe';
-        CKEDITOR.config.extraAllowedContent = 'area[shape,coords,href,target,alt,title];map[name];img[usemap];*[class,style]';
+        CKEDITOR.config.extraAllowedContent = 'area[shape,coords,href,target,alt,title];map[name];img[usemap];*[class,style];*(*);';
 
         // Do not convert é to &eacute; in the editor
         CKEDITOR.config.entities = false;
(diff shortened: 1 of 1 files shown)


fisharebest:
I am hesistant to allow wildcard attributes.

e.g. `<img src="no-such-url" onerror="do_evil()"`

photon-flip:
The wildcards in `*(*) `are for any `element`, allow any `class` name. This doesn't  change `attributes`  - any attribute would be in square brackets` [*]`. 
Your example `CKEditor documentation:

`elements [attributes]{styles}(classes)`
Where:
```
elements – a list of space-separated element names or an asterisk (*) character,
attributes – a comma-separated list of attribute names or an asterisk (*) character,
styles – a comma-separated list of style names or an asterisk (*) character,
classes – a comma-separated list of classes or an asterisk (*) character.
```

photon-flip:
Another quick note.
According to the [CKEditor documentation](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_best_practices.html#configure-content-filtering), the element wildcard in `config.extraAllowedContent` is for already enabled elements. It's not allowing all and any element so still quite restrictive.

photon-flip:
Sorry to pester but any chance of moving this forward? 
I have a project I've been working on for a month that is rendered pretty useless without this. And for everyone else it takes the unpredictability and frustration out of editing in CKEditor source mode.
I believe I've researched the documentation and tested this pretty well but if you're unsure it would be quite simple for you test yourself.
And it looks to me as though this was the intention of the original `*[class,style]; ` but that is not working s it should.

MurrayJ

photon-flip:
Tested that `*(*) `only allows classes for already allowed elements.
`dl dt dd` elements (which are not allowed elements) when added in source mode these are striped out even with `*(*) `in the `config.extraAllowedContent`

photon-flip:
Are there any other tests you would like me to do to check that attributes are not affected by `*(*) `it only allows classes.

I can't seem to find what the current `*[class,style] `is doing in practice. Removing it seems to have no effect. It looks like it was supposed to allow classes and styles on all allowed elements but that is not working. `*(*) `is the way to do that. 
without it, styling anything other than div elements involved torturous work arounds.

fisharebest:
> I can't seem to find what the current `*[class,style]` is doing in practice.

I'm guessing that it enables `class` and `style` attributes on any element.  This is old code, so perhaps it was needed on an older version of ckeditor?

But as you say, these attributes seem to be allowed by default, so no longer a need to specify them.

You have added `;*(*)` to allow classes.  Presumably you also want to add styles? `;*{*}(*)`. The server-side filtering allows classes and styles.

photon-flip:
Great, classes as well as inline styles working on all allowed elements.
Thanks.

PR #4960, opened 2024-03-11
Update of the http:// scheme

mvdbnl:
Replace the http:// scheme to https:// scheme. Except from the location where it is not needed or where no SSL Certificate is supported.


fisharebest:
Some of these changes are good (thanks!) - but not all of them.

* some sites (e.g. http://www.phpgedview.net) do not appear to have a site at https
* the minified files `*.min.css` do not need to be changed.  They will be updated automatically from the source `*.css` files whenever they are next rebuilt.
* as far as I can tell, the `xlmns` attribute in `.xml` files should not be changed. See https://github.com/w3c/svgwg/issues/738
* vendor files (ckeditor) should not be changed as they will be overwritten when we upgrade.  Perhaps create a PR for the upstream project?

PR #4962, opened 2024-03-13
Update messages.po

mdkaleel:
Added some Translation


fisharebest:
For future translations, can you use https://translate.webtrees.net.

You can login with github you can upload .PO files there.

Thanks!

PR #4972, opened 2024-03-27
Fix Disallow syntax in robots.txt

jon48:
There should be a colon after `Disallow` directives in  the Robots.txt.
Google accepts it without, but raises a warning in the Search Console.


fisharebest:
Thanks.  I had been cursing search engines for ignoring these :-)

PR #4987, opened 2024-05-20
Add wrapping behaviour to the fact note.

p-litwin:
Issue: When you add long url to the fact note it overflows the container.
I resolved it by adding overflow-wrap: break-word to the .wt-facts-table td .wt-fact-notes p

### DIFF
--- a/app/Elements/NoteStructure.php
+++ b/app/Elements/NoteStructure.php
@@ -165,7 +165,7 @@ class NoteStructure extends SubmitterText
             '</button> ' .
             I18N::translate('%1$s: %2$s', $label, $value) .
             '</div>' .
-            '<div class="ps-4 collapse ' . ($expanded ? 'show' : '') . ' ' . e($id) . '">' .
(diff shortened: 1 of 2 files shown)


fisharebest:
We have a CSS class `ut` (user-generated text) which provides this functionality.

Would it be better to add this class to the `p` element?  This is (possibly) in `NoteStructure.php`...

p-litwin:
> We have a CSS class `ut` (user-generated text) which provides this functionality.
> 
> Would it be better to add this class to the `p` element?  This is (possibly) in `NoteStructure.php`...

Good idea
For one line inline note it's already there, but multiline note is formatted elsewhere. I didn't locate it yet. I'll check this later.

fisharebest:
> I didn't locate it yet. I'll check this later.

I am away for a few days.  If you need help finding this, ask me next week.

p-litwin:
> > I didn't locate it yet. I'll check this later.
> 
> I am away for a few days.  If you need help finding this, ask me next week.

I found it already. Now it should work fine.

arbor95:
It seems, there is no change. Can you remove the pull request?

PR #4999, opened 2024-06-21
Plcaes tab - use different colour for the individual's own events

arbor95:
change the color of the icon on map for birth and christening of the selected individuum to red.

### DIFF
--- a/app/Module/PlacesModule.php
+++ b/app/Module/PlacesModule.php
@@ -151,7 +151,7 @@ class PlacesModule extends AbstractModule implements ModuleTabInterface
                 if ($fact->record() === $indi) {
                     $icon = static::OWNBIRTH_ICON;
                 }
-            } 
+            }
             elseif ($fact->tag() === 'INDI:CHR') {
                 if ($fact->record() === $indi) {
                     $icon = static::OWNBIRTH_ICON;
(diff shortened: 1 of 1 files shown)


arbor95:
hi Greg
perhaps you can help resolving the unsuccessful checks.
thanks Franz

fisharebest:
The code-check is complaining that you have `elseif` at the start of a line.  It should be on the previous line, after the `}`.

Would this be a simpler approach?  It makes it easy to modify all the icons for the individual.

```php
    protected const OWN_ICONS = [
        'INDI:BIRT' => ['color' => 'red', 'name' => 'baby-carriage fas'],
        'INDI:CHR'  => ['color' => 'red', 'name' => 'water fas'],
    ] + self::ICONS;

```


```php
                $icons = $fact->record() === $indi ? static::OWN_ICONS : static::ICONS;
```

```php
                        'icon'    => $icons[$fact->tag()] ?? static::DEFAULT_ICON,
```

arbor95:
what to do to handle the remaining unsuccessful check?

arbor95:
How to  handle:
Added line #L153 was not covered by tests
and the others "Added line"

fisharebest:
> How to handle:
> Added line #L153 was not covered by tests
> and the others "Added line"

You can ignore this.

It says that there are no tests that use these lines of the code.

But none of this module has any tests yet.

fisharebest:
Thanks!

Issue #5002, opened 2024-06-28
Family Lists / Indi Lists: records for Given Name starting with Umlaut not shown


arbor95:
**but there is still a little bug:**
The Indis of given names, beginning  with Umlauts, do not show under the corresponding letter. (Test by Select A or B or ALL given names)

you can see the test at [dfrese](https://dfrese.de/webtrees/tree/ahnen/individual-list?surname=MILLER&show=indi&alpha=M&show_marnm=no)  .

good/ok Ä no longer in selection, but
clicking A: result: no indis shown (HOver says 2), B-> one indi, **All: 3 indis**

![bug](https://github.com/fisharebest/webtrees/assets/6124739/e8de95ab-5425-48b2-817f-6e08ef97f362)

All clicked:

![bug](https://github.com/fisharebest/webtrees/assets/6124739/e7a163d7-573c-4e35-b26f-237432501f14)

_Originally posted by @arbor95 in https://github.com/fisharebest/webtrees/issues/5000#issuecomment-2196323256_

postscript: (do not wonder) I changed setup to show list of Given Names in case of more than 2.

arbor95:
created pull request #5009

PR #5013, opened 2024-07-26
Fixes: #5002 - individual list, given names starting with di-graphs not included

arbor95:
* setting correct title
* Solution for issue #5002

### DIFF
--- a/app/Module/AbstractIndividualListModule.php
+++ b/app/Module/AbstractIndividualListModule.php
@@ -716,7 +716,7 @@ abstract class AbstractIndividualListModule extends AbstractModule implements Mo
             assert($individual instanceof Individual);
 
             // The name from the database may be private - check the filtered list...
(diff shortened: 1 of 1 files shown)


app/Module/AbstractIndividualListModule.php
```diff
@@ -383,7 +383,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
 
                         if ($surname !== '' || $show_all === 'yes') {
                             if ($show_all !== 'yes') {
```
fisharebest, line 385: If we are changing this, then should we also remove the `if { .. }` around it?

When showing all surnames, we also want to have this new title?

arbor95: { } of foreach seems not to be possible,

fisharebest:
Thanks!

Issue #5011, opened 2024-07-25
DATE_PHRASES are  not displayed correctly


webtrees-pesz:
1 BIRT
2 DATE INT 1900 (1. January 1900)
is displayed but

1 BIRT
2 DATE (1. January 1900) 
is **not** displayed.

GEDCOM-5.5.1 says:
DATE_VALUE:= {Size=1:35}
[
<DATE> |
<DATE_PERIOD> |
<DATE_RANGE>|
<DATE_APPROXIMATED> |
INT <DATE> (<DATE_PHRASE>) |
**(<DATE_PHRASE>)**
]

arbor95:
should be fixed with pull request #5012

fisharebest:
`$match[1]` can contain user-generated text.  So it needs to be escaped using `e()`.

Would it be better to put this fix in `Date::display()` so that it works everywhere?

e.g. at https://github.com/fisharebest/webtrees/blob/main/app/Date.php#L181-L182 add something like

```
if ($this->text !== '') {
    $tmp .= ' (' . e($text) . ')';
}
```

(Note: I haven't tested this.  It's just a quick idea from looking at the code).

webtrees-pesz:
Thank you Franz and Greg for your quick support.

 

Best regards

Peter

 

Von: Franz Frese ***@***.***> 
Gesendet: Freitag, 26. Juli 2024 14:16
An: fisharebest/webtrees ***@***.***>
Cc: webtrees-pesz ***@***.***>; Author ***@***.***>
Betreff: Re: [fisharebest/webtrees] DATE_PHRASES are not displayed correctly (Issue #5011)

 

should be fixed with pull request #5012 <https://github.com/fisharebest/webtrees/pull/5012>

arbor95:
> `$match[1]` can contain user-generated text. So it needs to be escaped using `e()`.
> 
> Would it be better to put this fix in `Date::display()` so that it works everywhere?
> 
> e.g. at https://github.com/fisharebest/webtrees/blob/main/app/Date.php#L181-L182 add something like
> 
> ```
> if ($this->text !== '') {
>     $tmp .= ' (' . e($this->text) . ')';
> }
> ```
> 
> (Note: I haven't tested this. It's just a quick idea from looking at the code).

Seems as if the current lines 181 and 182 are used for many (other) purposes, that I don't overview.
So I would be content, when Dates in facts are displayed corresponding to gedcom definition.
[...]

PR #5014, opened 2024-07-26
bugfix for issue #5011

arbor95:
DATE_VALUE may be (<DATE_PHRASE>) .
value only checked for beginning ( bracket.


arbor95:
change to Date.php

Issue #5019, opened 2024-08-15
Redirected list branches from old url adds 'x' in front of surname


BertKoor:
Spotted an issue after my long overdue migration.
When an old url to a list of branches gets redirected, an 'x' is put at the front of the surname.

Analysis:
when getting the old url `/branches.php?ged=MyTree&surname=Doe`
it redirects to `/module/branches_list/Page/MyTree?soundex_dm=&soundex_std=&surname=Doe`
which in turn redirects to `/tree/MyTree/branches/xDoe?soundex_dm=0&soundex_std=0`

I found the issue is in function getPageAction of `/app/Module/BranchesListModule.php` :

````
'surname'     => 'x' . Validator::queryParams($request)->string('surname'), 
````

I'm preparing a PR to fix this myself

BertKoor:
Here's the PR with a fix: https://github.com/fisharebest/webtrees/pull/5020

fisharebest:
Closed by #5020

PR #5020, opened 2024-08-15
fix issue #5019: forward from old branch list url

BertKoor:
I've removed the part 
````
'x' .
````
which caused the issue. No idea how that creeped in.

### DIFF
--- a/app/Module/BranchesListModule.php
+++ b/app/Module/BranchesListModule.php
@@ -167,7 +167,7 @@ class BranchesListModule extends AbstractModule implements ModuleListInterface, 
         return redirect($this->listUrl($tree, [
             'soundex_dm'  => Validator::queryParams($request)->boolean('soundex_dm'),
             'soundex_std' => Validator::queryParams($request)->boolean('soundex_std'),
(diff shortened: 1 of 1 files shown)


fisharebest:
Hi Bert.  Thanks for this.

Yes - this fixes the problem.

But there is a slightly deeper problem here.

We have two levels of redirect.  The middle one is old/unused and can be deleted.

The first redirect is here:

https://github.com/fisharebest/webtrees/blob/01b92fd6b02917bce4b544b1295f7954fc92e3ce/app/Http/RequestHandlers/RedirectBranchesPhp.php#L61-L68

The second is here:

https://github.com/fisharebest/webtrees/blob/01b92fd6b02917bce4b544b1295f7954fc92e3ce/app/Module/BranchesListModule.php#L160-L172

What I *think* we need is this:

* The first redirect should be to the final URL (e.g. `$module->getListUrl()`).
* The function `getPageAction()` needs deleting.

Do you want to update the PR to do this?

BertKoor:
> Do you want to update the PR to do this?

Sure, I'll have a go at that somewhere next week.

fisharebest:
I'll make the other change.

PR #5025, opened 2024-09-01
More information about people connected to slide-show photo

kiwi3685:
PR related to forum discussion at : https://www.webtrees.net/index.php/forum/9-request-for-new-feature/39211-more-information-about-people-connected-to-photo#106897


fisharebest:
Is there a reason for the extra `<span class="wt-slide-show-link-name">`?

Would it be simpler to use `<li class="wt-slide-show-link">`?

Also, can we remove the `small`?  For accessibility, I'd prefer that text didn't go smaller than the default/baseline size.

kiwi3685:
Fine hy me.  Nigel Osborne (mobile)On 14 Sep 2024, at 9:57 PM, Greg Roach ***@***.***> wrote:﻿
Is there a reason for the extra <span class="wt-slide-show-link-name">?
Would it be simpler to use <li class="wt-slide-show-link">?
Also, can we remove the small?  For accessibility, I'd prefer that text didn't go smaller than the default/baseline size.

PR #5027, opened 2024-09-03
fix: undefined array key in setup wizard

jbarbey:
Hello

With the development version and a new installation. I have a message _undefined array key "dbkey"_ in the Setup screen of the DB.
The error is related to issue 3571

I added the missing values ​​in DEFAULT_DATA of the SetupWizard.php


fisharebest:
Thanks!  Can you fix the alignment?  Add extra spaces so that the `=>` align vertically.

jbarbey:
I fix alignemnt in file SetupWizard.php

fisharebest:
Thanks!

PR #5033, opened 2024-09-26
automatically call and display the header and footer set in XML

snsupersharp:
Remove the lines and content of the default header and footer in TCPDF, automatically call and display the header and footer set in XML, and also automatically call when paging the page content, but the header and footer cannot have too much content beyond the page range.


fisharebest:
Can you give me some background to this change?

What problems does it solve?
Why make the change?
etc.

Thanks,

snsupersharp:
I am not a programmer, but I think there is a bug here in 2.1.20. When you define HEADER and FOOTER in the XML of the REPORT, they will be displayed as defined in version 1.7.20, but in 2.1.20, only TCPDF's default lines and page numbers will be displayed。

Part of XML file:
	<Doc pageSize="$pageSize" orientation="landscape" >
		<Header>
			<Cell align="center" height="20" newline="1" style="header"><var var="I18N::translate('Births')" /></Cell>
			<Cell align="rightrtl" newline="1" style="pagenum"><var var="I18N::translate('Page')" /> <PageNum /> <var var="I18N::translate('of')" /> <TotalPages /></Cell>

			<SetVar name="fill" value="" />
			<if condition="$sortby=='NAME'"><SetVar name="fill" value="#AAAAAA" /></if>
			<Cell bgcolor="$fill" border="1" width="200" style="label"><var var="I18N::translate('Name')" /></Cell>

			<SetVar name="fill" value="" />
			<if condition="$sortby=='BIRT:DATE'"><SetVar name="fill" value="#AAAAAA" /></if>
			<Cell bgcolor="$fill" border="1" width="100" style="label"><var var="I18N::translate('Date')" /></Cell>

			<Cell border="1" width="150" style="label"><var var="I18N::translate('Place')" /></Cell>
			<Cell border="1" newline="1" style="label"><var var="I18N::translate('Source')" /></Cell>
		</Header>

		......
		
		
		<Footer>
			<Cell align="rightrtl" newline="1" style="date"><Now/></Cell>
		</Footer>
	</Doc>
And the pdf report of 1.7.20 
1 7 20 report
but ,there will have nothing in 2.1.20
2 1 20 report


Furthermore, I need to define a special HEADER as follows

my report
my report1
my report2

So I made some modifications to implement it, but because I am not a developer, the code is not very standardized and only serves as a suggestion for reference。

Thanks.

PR #5035, opened 2024-10-03
Wrap long urls inside Fact notes and Shared notes

p-litwin:
I added "ut" class to the divs containing the Fact Note and Shared.
This change will prevent overflow of the long urls out of the note divs in exp[anded state.
original beheviour
updated behaviour


fisharebest:
Thanks for the PR.  The screenshots make it much easier to review.

PR #5039, opened 2024-10-09
Adding 2FA for webtrees attempt #2 

kbwhizz:
Adds ability for user to switch on preference in account which makes a QR code appear. This can be scanned by a Google authenticator app and the secret is saved. At next log in, user must enter the code for the site from the authenticator when they log in. Can be switched off from account too. Incorporated into setup wizard and any add/edit user functions.


kiwi3685:
Sorry if this is already included, but I couldn't find it. 

Just a personal opinion, but I would prefer to have an administrator option to enable (or not) 2FA on my site, before users are offered the option.

kbwhizz:
> Sorry if this is already included, but I couldn't find it.
> 
> Just a personal opinion, but I would prefer to have an administrator option to enable (or not) 2FA on my site, before users are offered the option.

It isn't there at present. What screen would that option appear on. Couldn't see anything obvious in control panel.

kiwi3685:
Perhaps Control panel > Website > Sign-in and registration? Seems the most logical.

kbwhizz:
I'm afraid I'm completely stuck on the errors phpstan gives on app/User.php 

 126    Binary operation "." between non-falsy-string and mixed results in an  
         error.                                                                 
         🪪  binaryOp.invalid                                                   
  295    Parameter #1 $secret of method                                         
         PragmaRX\Google2FA\Google2FA::verifyKey() expects string, mixed        
         given.                                                                 
         🪪  argument.type    

For line 126, I tried making saving $_SERVER['SERVER_NAME'] to a variable and concatenating that, but it still doesn't like it.
For line 295, it looks like the Illuminate database returns a mixed pseudo-type ($secret) but Googl2FA verify wants a string. Any idea how I force it to become a string?		 

Any tips appreciated.

kbwhizz:
> I'm afraid I'm completely stuck on the errors phpstan gives on app/User.php
> 
> 126 Binary operation "." between non-falsy-string and mixed results in an error. 🪪 binaryOp.invalid 295 Parameter #1 $secret of method PragmaRX\Google2FA\Google2FA::verifyKey() expects string, mixed given. 🪪 argument.type
> 
> For line 126, I tried making saving $_SERVER['SERVER_NAME'] to a variable and concatenating that, but it still doesn't like it. For line 295, it looks like the Illuminate database returns a mixed pseudo-type ($secret) but Googl2FA verify wants a string. Any idea how I force it to become a string?
> 
> Any tips appreciated.

Used settype to fix this

fisharebest:
PHPstan is set on the "max" level, and it is not always possible to fix every issue.

> Any tips appreciated.

You've closed the PR, so I can't see your code any more.

Did you mean to do that?

kbwhizz:
Superceded

PR #5051, opened 2024-10-27
Use SVG pedigree icon in chart box

ekdahl:
After this, 
`.icon-pedigree {
    content: url(minimal/images/pedigree.png);
}`
can be removed from themes, if preferred, to default to SVG icon instead.
https://github.com/fisharebest/webtrees/blob/main/resources/css/minimal.css#L890


ekdahl:
If these types of changes are desired, there are other places where I would like to make the same adjustments.

fisharebest:
This change has been included in #5087

PR #5055, opened 2024-11-15
Improved layout of reports

sevtor:
The Report module is updated to allow the creation of reports with clickable references and better layout of pdf reports.


kiwi3685:
Wouldn't this be better produced as an add-on module? 
My understanding of older issues discussed here suggests that the current report system is to move in a completely different direction, so incremental changes like this are not required.

PR #5087, opened 2024-12-28
Use SVG icons for welcome module

ekdahl:
Adds default SVG icons that is currently still overridden by supplied themes.
There are probably better ways to do this. Feedback welcome.


fisharebest:
> Feedback welcome.

I like this idea.

* I have created a new SVG icon `account`.  Otherwise we have the same icon for "My individual" and "My account"
* In the array `$links[]`, I have renamed the `icon` element to `class` (to be more accurate).
* In the array `$links[]`, I have passed a complete SVG icon - not just the class names.
* I have made similar changes to the "tree" welcome block. 
* I have made similar changes to other icons, such as the chart-box.
* I have removed many of the PNG icons from the FAB and Minimal themes.

Apologies for rewriting your PR.  It was just easier to make all the above changes than provide lots of feedback.

Issue #5049, opened 2024-10-26
Default silhouette images as inline SVGs


ekdahl:
Could the default silhouette images be inline SVG images instead, using currentColor as fill color?
This would enable a simple way to change the color of the silhouette images.
Maybe it would be a breaking change for existing themes.

A search online gives some sample silhouette images:
https://openclipart.org/detail/157891/male-profile-silhouette
https://openclipart.org/detail/157927/female-profile-silhouette

Or maybe the current ones have vector original images?

Alternatively, how can this be done in a custom theme?

fisharebest:
AFAIK, the current silhouette icons came from a public clipart library and don't have SVG originals.

More than happy to use SVGs instead of PNGs.

At present, we use 3 icons - male/female/unknown.

There is a fourth option "other", which currently uses the same icon as unknown.

So, we either need three (or four) icons - or we ignore gender and just use the same icon for both.

We already have a "user" icon (control panel, sign-in/registration).  We could use that.

I'd be curious to know the opinion of other users.

LarsRabe:
For "unkown" there are (neutra) silhouettes with a question mark in the head. How about that?

LarsRabe:
For gender neutral silhouettes there are a number of heads without hair or other markings which would be typical for male or female.

For unknown there are heads with a question mark in the head.

How about using those 2 beside female and male?

ekdahl:
I think it adds value to have different generic images for male and female individuals in the tree. But seems to be a bit difficult to find suitable sets of images with a "neutral" image (for unknown and other) also in the same style. At least I don't yet have a good suggestion of a set of images that also has this.

LarsRabe:
@ekdahl 
This a example of what I mean:

JustCarmen:
I use SVG images in the Justlight theme, made with [Icomoon](https://icomoon.io/) from the original webtrees silhouettes. 

https://github.com/JustCarmen/webtrees-theme-justlight/blob/d3b19ae7cb837d544940e4224d838af572aad7a4/JustlightTheme.php#L196

https://github.com/JustCarmen/webtrees-theme-justlight/blob/d3b19ae7cb837d544940e4224d838af572aad7a4/resources/fonts/icomoon.svg?short_path=8be431d#L10

https://github.com/JustCarmen/webtrees-theme-justlight/blob/d3b19ae7cb837d544940e4224d838af572aad7a4/resources/css/justlight.min.css#L20130

[...]

PR #5088, opened 2024-12-28
Add default SVG silhouette images

ekdahl:
~~Should not affect supplied themes, until "content" CSS property is unset for parent element.~~
Suggestion for #5049.

* Reworked PR to use more "human" images.
Images are from @magicsunday here: https://github.com/magicsunday/webtrees-fan-chart/tree/f7029367c384237249d73ff1241ea0203dc1e24a/resources/images

* Not sure exactly how to incorporate SVG images into webtrees. I guess resources/views/icons is intended for smaller icon images, not really this kind of image.
Consider this PR as a POC for the time being.

* Minimal theme is adapted to use the new images.

* I have not yet touched the smaller silhouette images.

Sample images
image
image


fisharebest:
I like the idea of using the SVG icons for the silhouette image.

Do you think the "solid" icon is better for this?  https://fontawesome.com/icons/user?f=classic&s=solid

We should use the same image in the chart boxes.

Perhaps create a new view (e.g. resources/views/silhouette.phtml) that we can use for both.

I see that the code for thumbnail images / silhouettes is slightly different for chartboxes and the individual page.  I need to make this consistent.

ekdahl:
Updated this PR. Description is in the first comment.

ekdahl:
Hi @fisharebest,
Did you see the update to this PR?
I think the SVG profile images from @magicsunday looks really nice and it would be a nice addition to use them in webtrees.

PR #5112, opened 2025-02-16
Update BadBotBlocker.php Added AI Bots

HerzScheisse:
stumpled upon this repo lately and thought it might be usefull to also add the AI bots listed in that repo... https://github.com/ai-robots-txt

Please let me know if you want me to add the comments/decription too...

### DIFF
--- a/app/Http/Middleware/BadBotBlocker.php
+++ b/app/Http/Middleware/BadBotBlocker.php
@@ -61,8 +61,13 @@ class BadBotBlocker implements MiddlewareInterface
         'admantx',
         'Adsbot',
         'AhrefsBot',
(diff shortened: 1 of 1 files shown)


FrankWarius:
BadBotBlocker.php is usually updated once or twice a year with each release. That's not exactly fast.

What if the bad bots were checked daily with the Webtrees release check - or would that be too much traffic on the server?

fisharebest:
I have copied this list of AI robots into a separate list in the BadBotBlocker.  This should make it easier to update.

> What if the bad bots were checked daily with the Webtrees release check - or would that be too much traffic on the server?

I think I need to create new releases more frequently... :-)

Issue #5119, opened 2025-03-14
Add block icons


ekdahl:
I think it would look nice if the blocks on "My page" and "Home page" would have icons.
A suggestion is to move the settings wrench to the top right corner and add icons representing the block to the top right corner.

![Image](https://github.com/user-attachments/assets/a2c5a589-125b-41de-b0ac-912006b254d6)

PR #5120, opened 2025-03-16
Add block icons

ekdahl:
Suggestion for #5119. If solution is ok, icons can added for other blocks in this PR.

Results in block icon on the left and settings icon on the right:
image


fisharebest:
There are two changes here:

1. Moving the "configure" icon to the right
2. Adding icons to the title

I think the first looks good.  I would prefer ` class="d-flex justify-content-between"` to `float: end`.
I think it would look better if the title is long an wraps onto two lines.

I'm not sure about the second change.  Maybe if we had icons for more/all of the blocks?  It seems odd to have them for just a few blocks.

ekdahl:
1. Ok, will update
2. Yes, that's why it's only a draft PR, just selected a few blocks to get feedback for the concept before doing more changes.

ekdahl:
> I think the first looks good. I would prefer ` class="d-flex justify-content-between"`
A question regarding this. Wouldn't that make the title centered instead of directly after the icon at the left side?

ekdahl:
Made an attempt to use d-flex instead. Looks like this:
image
image
image
image

Not so good for colors unfortunately
image

PR #5127, opened 2025-03-22
Fix phpstan call in GitHub Actions

serl:
Looks like a copy-paste mishap.

EDIT: there were ~500 violations. I just updated the baseline file to ignore them, not sure that was the right call

### DIFF
--- a/.github/workflows/phpstan.yaml
+++ b/.github/workflows/phpstan.yaml
@@ -26,4 +26,4 @@ jobs:
 
       - run: composer install --no-progress
 
-      - run: composer webtrees:phpcs
+      - run: composer webtrees:phpstan


fisharebest:
> Looks like a copy-paste mishap.

Good spot!

> EDIT: there were ~500 violations. I just updated the baseline file to ignore them, not sure that was the right call

Can you leave the baseline as it was.  I'll update this as I merge the PR.  Thanks!

serl:
> > EDIT: there were ~500 violations. I just updated the baseline file to ignore them, not sure that was the right call
> 
> Can you leave the baseline as it was. I'll update this as I merge the PR. Thanks!

👍

PR #5138, opened 2025-04-13
Media download fix

CDS83:
A proposal to fix download issues (gateway timeout and DB timeout) of ZIP archives including media files on shared servers with low performance and no control on DB time out values.

The fix idea is to use PHP zip extension as a primary solution to add media files to the archive, and use Flysystem streams if the extension is not available or if the zip creation throws an error.

### DIFF
--- a/app/Services/GedcomExportService.php
+++ b/app/Services/GedcomExportService.php
@@ -21,6 +21,7 @@ namespace Fisharebest\Webtrees\Services;
 
 use Fisharebest\Webtrees\Auth;
 use Fisharebest\Webtrees\DB;
+use Fisharebest\Webtrees\Site;
 use Fisharebest\Webtrees\Encodings\UTF16BE;
 use Fisharebest\Webtrees\Encodings\UTF16LE;
 use Fisharebest\Webtrees\Encodings\UTF8;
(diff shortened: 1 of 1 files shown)


fisharebest:
The problem with `$zip->addFile(...)` is that it assumes the file is stored on a local disk.

The design of webtrees allows media files to be stored in any filesystem - local or remote.
The core code uses a local filesystem, but modules can easily change this.

So, this code would break the downloads for users of such a module/customisation.


Now, I agree it would be good to use the PHP `ZipArchive` class directly (the FlySystem class is based on this), since `addFile()` is more efficient - especially if combined with `ZipArchive::setCompressionIndex(..., ZipArchive::CM_STORE)`.

But to use this, we would need to be able to detect whether the supplied `$media_filesystem` is local or remote, and only bypass it when it is local.  The FlySystem library does not allow us to do this.

So. before we can make this change, we need a way to know if the current media filesystem is local or not.
I'll investigate the best way to do this.

fisharebest:
I have made the changes suggested above.

The code you suggested (adding files to ZipArchive by name) is now present, but commented out.

Still need the right logic for deciding whether to use it.

CDS83:
Thank you for the detailled answer.

Please tell me: if nothing else works, could adding an option to WT config file to control the zip behavior be a valid solution ?

fisharebest:
There have been lots of changes to this functionality since the PR was raised, which fix the issues.

We use the PHP Zip library directly, not via Flysystem.  We don't compress images.  This reduces CPU and memory limits, including the restriction that the ZIP must fit in memory.

PR #5145, opened 2025-04-21
UID References in notes and Text

elysch:
Added an option to create _UID|UID references with options for different text on links.

It puts the full name if not specified an option.
It lets put any text as the link text, if added after a `:`. (It permits the use of # symbols in the link text, if escaped like: `\#`)

It works for UID's coming from Individuals, Families, FamilyNames *(is it needed this search?)*, Repositories, Sources, Shared Notes, Locations and Media. It search in all of them and stop's the first time it get's results, and the results are not other references.

Examples showing the full name only:

> `Somebody named #FE7C45EE57AD4C4DA6675AFED8B59B8F966987# was already there`

> `Somebody named #FE7C45EE57AD4C4DA6675AFED8B59B8F966987:FULLNAME# was already there`

Both examples get rendered like this: *Somebody named [John Doe](https://example.com/webtrees/tree/fam/individual/X771/John-Doe) was already there*

Example showing an arbitrary text as the link text:

> `Somebody named #FE7C45EE57AD4C4DA6675AFED8B59B8F966987:Jonny Doe (born in 1765, \#1 in his family)# was already there`

This example get's rendered like this: *Somebody named [Jonny Doe (born in 1765, #1 in his family)](https://example.com/webtrees/tree/fam/individual/X771/John-Doe) was already there*

Hope you get the idea with this.

I recommend to always use arbitrary text that permits to identify what information is being referred to, just in case the UID or _UID gets removed/replaced in the destination and the link stops working. Like my last example.

Ely

### DIFF
--- a/app/CommonMark/UidExtension.php
+++ b/app/CommonMark/UidExtension.php
@@ -1,0 +1,50 @@
+<?php
+
+/**
+ * webtrees: online genealogy
+ * Copyright (C) 2025 webtrees development team
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+declare(strict_types=1);
+
+namespace Fisharebest\Webtrees\CommonMark;
+
+use Fisharebest\Webtrees\Tree;
+use League\CommonMark\Environment\EnvironmentBuilderInterface;
+use League\CommonMark\Extension\ExtensionInterface;
+
+/**
+ * Convert UIDs within markdown text to links
+ */
+class UidExtension implements ExtensionInterface
+{
+    private Tree $tree;
+
+    /**
+     * @param Tree $tree Match UIDs in this tree
+     */
+    public function __construct(Tree $tree)
+    {
+        $this->tree = $tree;
+    }
+
+    /**
+     * @param EnvironmentBuilderInterface $environment
+     */
+    public function register(EnvironmentBuilderInterface $environment): void
+    {
+        $environment
+            ->addInlineParser(new UidParser($this->tree))
+            ->addRenderer(UidNode::class, new UidRenderer());
+    }
+}
--- a/app/CommonMark/UidNode.php
+++ b/app/CommonMark/UidNode.php
@@ -1,0 +1,59 @@
+<?php
+
+/**
+ * webtrees: online genealogy
+ * Copyright (C) 2025 webtrees development team
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+declare(strict_types=1);
+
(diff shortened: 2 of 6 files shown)


elysch (commented):
Changed the Regex for the reference. Now it's much simpler

elysch:
If you agree and want me to help doing it, please let me know which file or commit I can inspect to find out how to implement the new xref-uid table.

I can imagine we would need to:
a) create the table with the index in UID column. Must be possible to add multiple UID's to the same xref, so the xref column must not be unique.
b) go through all the records (individuals, families, sources, notes, locations...) looking for a UID or _UID field. Maybe it would be better to export to a *.ged file and take all from there. Right?
c) insert all the UIDs found in the new table
d) decide what to do with the duplicated UIDs, but they wouldn't be in the new table. Maybe move them to a note with some way to find them easily.
e) force the uid additions, deletions or changes to use the new table... not forgetting to do the same when the raw gedcom is edited.
f) change the firstUid method to use the new table.

anything else?

Ely

elysch:
Hello again.

I'm thinking on doing something like the following. Since I'm very rookie in webtrees development, not sure if it's the right way and not sure how to warrantee there is no inserted row with all XREF fields empty.

Also, I'm not sure what to do next. I think I'll need a class that inherits from GedcomRecord to handle the uid-xref relations. and go, one by one, through all the individuals, families, media, other (?) and sources looking for UIDs to insert in the new created table.

Any suggestions?

```
class Migration0 implements MigrationInterface
{
    /**
     * Upgrade the database schema
     */
    public function upgrade(): void
    {
        // Create a table with UID - XREF relations
        if (!DB::schema()->hasTable('uid')) {
            DB::schema()->create('uid', function (Blueprint $table): void {
                $table->integer('u_id', true)->charset('utf8')->collate('utf8_unicode_ci');
                $table->integer('u_uid');
                $table->integer('u_indi_xref')->charset('utf8')->collate('utf8_unicode_ci')->nullable();
                $table->integer('u_fam_xref')->charset('utf8')->collate('utf8_unicode_ci')->nullable();
                $table->integer('u_media_xref')->charset('utf8')->collate('utf8_unicode_ci')->nullable();
                $table->integer('u_other_xref')->charset('utf8')->collate('utf8_unicode_ci')->nullable();
                $table->integer('u_source_xref')->charset('utf8')->collate('utf8_unicode_ci')->nullable();
                $table->integer('u_gedcom_id');

                $table->unique(['uid']);

                $table->foreign('gedcom_id')->references('gedcom_id')->on('gedcom');
            });

            // The database engine shouldn't permit all xref id's null in one row. I think the only way to achieve this is with a trigger.
            // AFAICT webtrees doesn't use triggers.

            // SQL-server cannot cascade-delete/update on self-relations.
            // Users will need to delete all child locations before deleting the parent.
            if (DB::driverName() === DB::SQL_SERVER) {
                // SQL-Server doesn't support 'RESTRICT'
                $action = 'NO ACTION';
            } else {
                $action = 'CASCADE';
            }

            DB::schema()->table('uid', static function (Blueprint $table) use ($action): void {
                $table->foreign(['u_indi_id'])
                    ->references(['i_id'])
                    ->on('individuals')
                    ->onDelete($action)
                    ->onUpdate($action);
            });

            DB::schema()->table('uid', static function (Blueprint $table) use ($action): void {
                $table->foreign(['u_fam_id'])
                    ->references(['f_id'])
                    ->on('families')
                    ->onDelete($action)
                    ->onUpdate($action);
            });

            DB::schema()->table('uid', static function (Blueprint $table) use ($action): void {
                $table->foreign(['u_media_id'])
                    ->references(['m_id'])
                    ->on('media')
                    ->onDelete($action)
                    ->onUpdate($action);
            });

            DB::schema()->table('uid', static function (Blueprint $table) use ($action): void {
                $table->foreign(['u_other_id'])
                    ->references(['o_id'])
                    ->on('other')
                    ->onDelete($action)
                    ->onUpdate($action);
            });

            DB::schema()->table('uid', static function (Blueprint $table) use ($action): void {
                $table->foreign(['u_source_id'])
                    ->references(['s_id'])
                    ->on('sources')
                    ->onDelete($action)
                    ->onUpdate($action);
            });
        }
    }
}
```

fisharebest:
Changing the database structure needs a lot of careful thought.

It also needs some sort of data migration so that existing records are stored.

This is already planned - I want to change webtrees from being "record based" (level 0 GEDCOM) to being "fact based" (level 1 GEDCOM).  This will have an index on the fact type (e.g. `UID`) and the fact value.  But a lot of work is needed, and so won't be ready for a long time.

elysch:
The change from record oriented to fact oriented sounds really great. 

I don't have particularly much free time, but would be glad to help in any way I can.

elysch:
I'm closing this, since it's not going to be accepted the way it works now.

PR #5146, opened 2025-04-23
fix #4467 support for 4k/5k monitors

### LINKED ISSUE #4467: Request: Add one more container size for even bigger screens
In /webtrees/public/css/vendor.min.css there are different container sizes:

```
@media (min-width:576px){.container,.container-sm{max-width:540px}}
@media (min-width:768px){.container,.container-md,.container-sm{max-width:720px}}
@media (min-width:992px){.container,.container-lg,.container-md,.container-sm{max-width:960px}}
@media (min-width:1200px){.container,.container-lg,.container-md,.container-sm,.container-xl{max-width:1140px}}
```

Would it be possible to add one more for even bigger screens, e. g.:

`@media (min-width:1600px){.container,.container-lg,.container-md,.container-sm,.container-xl{max-width:1440px}}`

BertKoor:
Bootstrap sets usable width to 52% for 5k monitors.

See forum discussion 
https://www.webtrees.net/index.php/forum/help-for-release-2-1-x/39936-diagrams-how-to-set-to-use-the-entire-screen-width 
and bootstrap code https://github.com/twbs/bootstrap/blob/main/site/data/breakpoints.yml

### DIFF
--- a/resources/css/_base.css
+++ b/resources/css/_base.css
@@ -63,6 +63,11 @@ a:hover {
     text-decoration: underline;
 }
 
+/* Extend screen width limitation of bootstrap for 4k/5k devices */
+@media (min-width: 1700px) { .container, .container-lg, .container-md, .container-sm, .container-xl, .container-xxl {max-width: 1600px} }
+@media (min-width: 2100px) { .container, .container-lg, .container-md, .container-sm, .container-xl, .container-xxl {max-width: 1980px} }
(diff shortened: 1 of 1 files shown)


hiSandog:
A useful visual regression case would be a 4K/5K viewport with browser zoom above 100% and the diagram sidebar or controls open. Breakpoints are based on CSS pixels rather than physical panel resolution, so a nominally 5K display can still enter a narrower layout at common scaling settings. Checking that combination could catch overflow or unexpectedly constrained diagrams.

PR #5149, opened 2025-05-02
Update LanguageGerman.php

arbor95:
add the relationships for german language.


fisharebest:
Would you be able to add some test cases to https://github.com/fisharebest/webtrees/blob/main/tests/feature/RelationshipNamesTest.php

PR #5150, opened 2025-05-13
update for german relationships

arbor95:


fisharebest:
Hi @arbor95 - thanks for this!

I have made some small changes, to fix the tests and code-style.

You have a function `$genitive($prefix, $suffix, $gender)`.

If `$gender` is male/female/unknown, then it would be clearer to use `'M'`/`'F'`/`'U'` instead of `0`/`1`/`2`
The logic for `1` and `2` is the same.  Is this correct?

arbor95:
0 = M
1 = F
2 = independent  (I did not distinguish. its only used once for ->descendant() )

fisharebest:
I think we are missing definitions for cousins.

fisharebest:
Those are just simple "first cousins" (same grandparents).

What about cousins where the common ancestors are more generations?

Most languages have a relationship "Nth cousin" or "cousin of the Nth degree".  But the rules for calculating "N" are different in each language.

arbor95:
AfaIk there is no special naming in german language.

fisharebest:
> AfaIk there is no special naming in german language.

At present, German uses a similar system to Engish

First cousin (same grandparents) => Cousin 1. Grades
Second cousin (same great-grandparents) => Cousin 2. Grades
Third cousin (same great-great-grandparents) => Cousin 3. Grades

Should we keep this, or use "Cousin/e" for every type of cousin?

When cousins are not "symmetrical", English uses "removed ascending" and "removed descending".

German currently uses the same system, translating it as "Cousin 3 grades 2. Grades in aufsteigender Folge"

I asked ChatGPT (!), and is says German uses only symmetrical cousins.  e.g. Grandparent's Cousin, or Cousin's child.

arbor95:
What is missing to get this pull request into the webtrees code?

PR #5160, opened 2025-05-28
when logging out the homepage is fetched twice

### LINKED ISSUE #5153: Logout action triggers a GET of the homepage twice
While investigating an [issue reported on the forum](https://www.webtrees.net/index.php/forum/help-for-release-2-2-x/40009-forbidden-403-error-on-logout) I discovered that after a logout the homepage is fetched twice. I suspect some servers may be configured with a rule engine such that the second request gets a response 403-Forbidden.

Image

I'm surprised to see the first request has type 'fetch' (an earlier test showed it with size zero) while the second has type 'document'.

I have tested other actions which do a POST request (login, quick search) but these get a single response.

I suspect there's something wrong with the promise (consumer twice?)

Screenshots of logging out at the demo server:

----

Image
Image

----

Image
Image

----

Same as performed on the server of jantony:

----

Image
Image

----

Image

BertKoor:
fix #5153

The logout request handler may respond with a simple 204-NoContent. The originating view already contains redirection. As reported on the [forum](https://www.webtrees.net/index.php/forum/help-for-release-2-2-x/40009-forbidden-403-error-on-logout?start=0), some servers don't cope well.

Affected user reported back this patch does the trick: logout goes smoothly.

### DIFF
--- a/app/Http/RequestHandlers/Logout.php
+++ b/app/Http/RequestHandlers/Logout.php
@@ -29,8 +29,7 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 
-use function redirect;
-use function route;
+use function response;
 
 /**
  * Perform a logout.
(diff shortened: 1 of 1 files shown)


fisharebest:
Thanks!

Issue #4834, opened 2023-05-30
Search on "names" with accents


jgrincho:
When searching using Search + General Search and typing a name without accents (e.g. Joao),  the is nothing in the result. 
When searching using Search + Advanced Search and following the same as above, all individual named João are listed.  
When using phpmyadmin on the DB and running the query "SELECT `n_full` FROM `wt_name` WHERE `n_givn` LIKE 'Joao'", all individuals with João are listed.

I suppose there is an issue with the General Search.

PR #5163, opened 2025-06-03
fix #3646: same fix in List of Branches as what's done in #4834

### LINKED ISSUE #3646: General Search - name with accent
in webtrees 2.0.10 demo, there are individuals whose names include accents, for example :
- Augustus the Younger, Duke of Brunswick-Lüneburg
- Ferdinand Albert II, Duke of Brunswick-Wolfenbüttel 

In the 'General search' page, when search for Luneburg (without '¨') in the Families records, we can find married individuals whose names include Lüneburg. But when whe search for Luneburg (without '¨') in the Individuals records we can't find anything.

image

image

This is due to :
https://github.com/fisharebest/webtrees/blob/708e66987f7d6eed5675fea31f4074192e368cac/app/Services/SearchService.php#L1105

Perhaps one way to fix the problem should be adding something like this before mb_stripos ?

$rule = 'NFD; [:Nonspacing Mark:] Remove; NFC';
$myTrans = Transliterator::create($rule);
$gedcom = $myTrans->transliterate($gedcom);
$search_term = $myTrans->transliterate($search_term);

BertKoor:
The list of branches suffers from the same accented letter mismatch as the general search did

### DIFF
--- a/app/Module/BranchesListModule.php
+++ b/app/Module/BranchesListModule.php
@@ -158,7 +158,7 @@ class BranchesListModule extends AbstractModule implements ModuleListInterface, 
             ]));
         }
 
-        $surname     = I18N::language()->normalize(Validator::attributes($request)->string('surname', ''));
+        $surname     = Validator::attributes($request)->string('surname', '');
         $soundex_std = Validator::queryParams($request)->boolean('soundex_std', false);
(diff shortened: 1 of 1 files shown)


BertKoor:
too bad, it's a bit less efficient this way, but it makes sense

PR #5176, opened 2025-06-26
Update BadBotBlocker.php

communitysteps:
Update ai.robots.txt list to version 1.37 (33 bots added https://github.com/ai-robots-txt/ai.robots.txt/releases)


fisharebest:
There were some issues with the AI-Robots list; they used the wrong UPPER/lower case for some of the robot names.

I raised this, and they said the robots.txt spec says that the names should be matched case-insensitively.

But webtrees uses the correct case for each robot.  It's slightly faster, plus the case-conversion is no necessary.

So I merged the AI_ROBOTS list with the BAD_BOTS list to keep it simple.

So, this PR is now out-of-date and can't be merged.  But thanks for the contribution.

Another list that we could consider using is https://darkvisitors.com/agents - but we'd need to scrape the page to get it.

Issue #3648, opened 2021-01-01
Changes not fully shown after saving


WGroleau:
After adding a wife to a family, and refreshing the page, the wife is still not shown in the page title
<img width="403" alt="image" src="https://user-images.githubusercontent.com/5910274/103431734-f7819700-4b89-11eb-95aa-f8d9f6d21592.png">

WGroleau:
Actually, it is because the only name record is TYPE married.
Is it possible to use the first name record even if it is not "birth" or default?

fisharebest:
> Is it possible to use the first name record even if it is not "birth" or default?

We use the first name record (of any type) for the individual.

But to construct the family name, we use the first non-married name.

But if the only name is a married name, then I guess we should use it.

WGroleau:
One could argue that the same name should be used as on the INDI.  But one could also argue that the married name should be the one used on the page showing the marriage.

PR #5178, opened 2025-07-03
alt fix issue #3648: use married name in Family when no other name exists

BertKoor:
This is an alternative to the other PR.


fisharebest:
I prefered your first fix.  It is consistent with the logic for individual names.

PR #5179, opened 2025-07-10
completed list of Heredis custom tags

BertKoor:
Known issues: 

* restriction on subtag will show in red, so not to give impression this works
* coordinates of a submap/subdivision will not put a location marker on the map
* collapsing Research Information does not yet work as intended
* will put the labels in English and add French & German translations


fisharebest:
Thanks for all this.  Looks good,

I just need to review the new translation texts before I merge this.

PR #5192, opened 2025-08-01
fix #5191: removed leftover code from old datatables

### LINKED ISSUE #5191: User Administration triggers javascript console log: Uncaught TypeError: $(... ).dataTable(... ).fnFilter is not a function
Reported in the [forum](https://www.webtrees.net/index.php/forum/2-open-discussion/40129-new-release-of-webtrees-2-2-2-and-2-1-23?start=60#112390), it was a bit distracting while we were looking for a server response.

Located the cause here: https://github.com/fisharebest/webtrees/blob/main/resources/views/admin/users.phtml#L68
Looks like a leftover from a previous migration of DataTables 1.10 to 2.0.
The code "as is" seems to function, but imho it's best to fix this at the core.

BertKoor:
Removed the .fnFilter() which made the error go away
And the filter does work!

### DIFF
--- a/resources/views/admin/users.phtml
+++ b/resources/views/admin/users.phtml
@@ -65,6 +65,6 @@ use Fisharebest\Webtrees\View;
       /* verified          */ null,
       /* approved          */ null
     ]
-  }).fnFilter("<?= e($filter) ?>"); // Pre-fill the search box
+  });
 </script>
 <?php View::endpush() ?>


fisharebest:
Removing fnFilter does indeed fix the error - but we need to replace it with the 2.x code.  See the commit in #5191

PR #5205, opened 2025-08-12
fixed bug: Setup wizard reported memory_limit as zero

BertKoor:
Also made the magic division number a bit less magic.
Please also review the advised memory sizes here:
https://github.com/fisharebest/webtrees/blob/main/resources/views/setup/step-2-server-checks.phtml#L70-L74
This guide dates from over a decade ago, webtrees needs a lot more memory these days.


fisharebest:
Thanks!

PR #5208, opened 2025-08-12
fixed #5207: memory_limit = -1 means unlimited

### LINKED ISSUE #5207: Installation memory_limit and CPU warning but install wizard still continues
Hello,

SERVER OS: CloudLinux
Web Platform: cPanel
PHP: Using CloudLinux PHP selector

During installation, I received the following warning about the memory_limit and the CPU time limit:

Image

The installation continues so not sure why this warning is shown.

PLEASE NOTE: When using CloudLinux, the memory and cpu are set as hard ceilings but we as shared hosting users don't have access to these configuration settings so I do suspect that maybe webtrees is not playing nice with CloudLinux.

As you can see below, I have checked the memory_limit and have it set to 2G and confirmed this through the PHP CLI and the phpinfo() function.

The CloudLinux PHP selectors don't allow for manual input to use -1 as the value for the memory variable. I would have to switch to using the MultiPHP Manager in order to do that but the limit would anyway be limited to whatever CloudLinux has as a hard set.

Image

BertKoor:
Prevent an error or warning is displayed in setup wizard when memory_limit = -1.
See https://www.php.net/manual/en/ini.core.php#ini.sect.resource-limits

Reused the word "unlimited" which already had a translation.
The text is not perfect then, but it's already better than showing "-1"

### DIFF
--- a/app/Http/RequestHandlers/SetupWizard.php
+++ b/app/Http/RequestHandlers/SetupWizard.php
@@ -136,7 +136,7 @@ class SetupWizard implements RequestHandlerInterface
 
         $data['cpu_limit']    = $this->php_service->maxExecutionTime();
         $data['locales']      = $locales;
-        $memory_limit         = $this->php_service->memoryLimit();
+        $data['memory_limit'] = $this->php_service->memoryLimit();
 
         // Only show database errors after the user has chosen a driver.
         if ($step >= 4) {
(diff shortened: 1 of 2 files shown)


fisharebest:
The documentation only says that `-1` means "unlimited".  But AFAICT `0` has the same effect.

So I modified this to include zero in the check.

I also moved the bytes-megabytes conversion to the view, as I think this is part of the formatting/display logic.

Issue #5181, opened 2025-07-11
Backlink for level 1 shared notes


bschwede:
A **shared note** is linked to an individual, family etc. as _level 1 element_.
Starting from individual (...) page, there is no possibility to get to that shared note directly in order to edit it.
The pencil icon only edits the link.

Compared to shared notes linked to an fact/event as level 2 element, the label/title is a hyperlink to the shared note itself.

<img width="509" height="790" alt="Image" src="https://github.com/user-attachments/assets/22475f13-b710-412d-a416-99885cfe7495" />

**Wish**: You can't click the title (first line of note) / make clickable (as if you were in facts/events).
It would be more convenient for round tripping in webtrees 😄 

See also forum threads:
- https://www.webtrees.net/index.php/forum/help-for-release-2-1-x/37728-editing-shared-notes
- https://www.webtrees.net/index.php/forum/help-for-release-2-1-x/37189-link-to-shared-notes-from-individual-page

fisharebest:
> Make the first line of the Shared Note the link to the edit of the text of the note. Which is what we already do when a Shared Note is part of an fact in the fact and event list.

I'm not sure this works here.

For level 2 notes, we have expand/collapse.  We can show *both* the first line (a link to the note) and the entire note at the same time.  Remember that a note can consist of a link (e.g. using markdown `[example](http://example.com)`).

For level 1 notes, we do not expand/collapse them.
So for level 1 notes, we can't do this.

The only option, I think, is for the label "Shared note" to be a link.

PR #5212, opened 2025-08-22
fact label as link to shared note record

bschwede:
related to Backlink for level 1 shared notes #5181


fisharebest:
We use `Fact::label()` in lots of places - including ones where we cannot have HTML markup.  For example, in the `<title></title>` of some pages.

So I don't think that this is the correct solution to the issue.

PR #5213, opened 2025-08-22
fix #4828: prevent multiple uids on edit

### LINKED ISSUE #4828: Record has multiple uid fields
When editing **sources**, an additional unique identifier field (1 _UID) is added to the record each time.
The same behavior is seen with **media objects** and **repository** records.

Version: WT 2.1.16, no additional modules activated
Settings: Control panel > Manage family trees > Preferences Tree > Add unique identifiers = yes

See also: https://www.webtrees.net/index.php/en/forum/help-for-release-2-1-x/38047-record-has-multiple-uid-fields

bschwede:
related to Record has multiple uid fields #4828
covers the second part of that issue

### DIFF
--- a/app/Services/GedcomEditService.php
+++ b/app/Services/GedcomEditService.php
@@ -263,6 +263,7 @@ class GedcomEditService
         $next_level = substr_count($tag, ':') + 1;
         $factory    = Registry::elementFactory();
         $subtags    = $factory->make($tag)->subtags();
+        $hasUid     = str_contains($gedcom, "\n1 _UID ");
 
         // The first part is level N.  The remainder are level N+1.
         $parts  = preg_split('/\n(?=' . $next_level . ')/', $gedcom);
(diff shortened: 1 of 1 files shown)


fisharebest:
It is valid to have multiple UID fields.  I think we should be able to create valid data using the GUI.

So, we create an empty field, where a new UID can be created.

Screenshot 2025-08-24 at 20 37 38

If you do not click the `create` button, then no UID is saved.

So, the current logic allows us to create multiple UIDs, but by default does not.

This PR would prevent the second UID field from appearing.  I am not 100% convinced this is a good change.

But if we really want to prevent multiple UIDs, then it would be simpler to change the defintion of `_UID` from `{0:M}` to `{0:1}`.

bschwede:
Thanks for your comment.
How silly of me, I didn't want to prevent multiple UIDs from being assigned in general. I probably had the behavior of v2.2.1 in mind and didn't notice the new create button at the end of the input field. So this PR is more than superfluous :smile:

PR #5222, opened 2025-09-04
fix end of file check for tree-import cli command

rudy-b-1:
When the tree-importer cli command is checking if it has finished reading in the gedcom file, it is adding the size of the entire buffer on every read, instead of just the new bytes read. This makes it think it has already read more of the file than it actually has.  Mostly, the command will work normally since the leftover bytes in the buffer aren't enough to affect the next buffer read.  But, if the gedcom file is very large and/or has large records, the leftover bytes will accumulate past the 8192 buffer size and the command won't read in the complete gedcom file.


fisharebest:
Thanks!

Issue #5223, opened 2025-09-05
v2.1 Control Panel -> Privacy Restrictions shows no labels for Fact/Event


BertKoor:
See forum discussion:
https://www.webtrees.net/index.php/forum/help-for-release-2-1-x/40243-wrong-display-of-privacy-restrictions-in-v2-1-25

Expected:

<img width="892" height="383" alt="Image" src="https://github.com/user-attachments/assets/64fde244-5b31-4934-967d-62e22501dfc8" />

But was:

<img width="893" height="375" alt="Image" src="https://github.com/user-attachments/assets/1811d348-795d-41e7-9d2a-d848e06d182a" />

fisharebest:
AFAICT, this is fixed.

PR #5224, opened 2025-09-05
fix issue #5223 wt2.1: privacy fact labels on control panel

BertKoor:
Fixed by copying the TreePrivacyPage.privacyRestrictions method body from the main branch.
Tested with php8.1/8.2, not tested with PHP7.4-8.0


fisharebest:
Thanks!

PR #5227, opened 2025-09-08
Add VendorModuleService for Composer-based Webtrees modules

magicsunday:
# Add `VendorModuleService` for Composer-based Webtrees modules

### Overview
This PR introduces a new service class, **`VendorModuleService`**, that allows Webtrees to seamlessly load modules installed via Composer.  

A key part of this feature is the introduction of a **new Composer package type: `webtrees-module`**. This package type signals that a Composer package is a Webtrees extension, making it automatically discoverable and loadable by the Webtrees core.  

This modernizes module management and brings Webtrees closer in line with practices used by frameworks such as Symfony, Laravel, and TYPO3.

### Why?
Currently, Webtrees modules are installed and managed manually in the `modules_v4` directory. While functional, this approach has limitations:
- ❌ No standardized distribution format for third-party modules  
- ❌ Manual installation and updates are error-prone  
- ❌ No integration with Composer’s dependency management  

By enabling **Composer-managed modules** (via `webtrees-module`), we provide:
- ✅ A standardized installation process (`composer require vendor/package`)  
- ✅ Automatic updates alongside Webtrees itself  
- ✅ A cleaner separation of first-party and third-party modules  
- ✅ A familiar workflow for PHP developers who already rely on Composer  

This makes Webtrees more attractive to external developers and lowers the barrier for creating and sharing extensions.

### How It Works
The `VendorModuleService` leverages Composer’s **InstalledVersions API** to:
1. Discover installed Composer packages of type **`webtrees-module`**  
2. Load and validate their `module.php` entrypoint  
3. Register the module within Webtrees’ existing `ModuleService`  
4. Provide meaningful error handling via `FlashMessages`  

Internally:
- Modules are identified by `composer.json` → `"type": "webtrees-module"`  
- Installed paths are retrieved via Composer’s runtime API  
- Modules are assigned unique names based on their package (`_mypackage_`)  

### Benefits
- **For Developers**: Publish Webtrees modules directly on Packagist, making them discoverable and installable with a single command.  
- **For Administrators**: No need to manually upload files; Composer ensures correct versions and resolves dependencies.  
- **For the Project**: Aligns Webtrees with modern PHP ecosystem practices, ensuring sustainability and extensibility.  

### 📦 Example `composer.json` for a Webtrees Module
```json
{
  "name": "my-vendor/my-webtrees-module",
  "description": "An example Webtrees module distributed via Composer.",
  "type": "webtrees-module",
  "license": "GPL-3.0-or-later",
  "authors": [
    {
      "name": "Your Name",
      "email": "you@example.com"
    }
  ],
  "require": {
    "php": ">=8.1",
    "fisharebest/webtrees": "^2.2"
  },
  "autoload": {
    "psr-4": {
      "MyVendor\\MyWebtreesModule\\": "src/"
    }
  }
}

### CODE REVIEW

**Comment by kiwi3685** [comment]
I develop modules purely for my own use. I have no need or interest in using composer. I know there are many others doing the same. For us modules_v4 is simple and works well.

So how does this proposal affect us? There is no mention of users like us in this.

**Comment by magicsunday** [comment]
Thanks for raising this — it’s an important point.  

To be clear: **this proposal does not remove or break the existing `modules_v4` system**. If you are developing modules purely for your own use, or you prefer to keep things simple by dropping files into `modules_v4`, you can continue to do so exactly as you always have. Nothing changes for you.  

What this PR does is **add an additional option** for distributing and installing modules via Composer, using the new `webtrees-module` package type. This is mainly targeted at:  
- Developers who want to publish their modules to Packagist for broader use  
- Administrators who prefer dependency management through Composer  
- The Webtrees ecosystem as a whole, to provide a standardized, modern way of sharing extensions  

So to summarize:  
- ✅ Your existing workflow with `modules_v4` continues to be fully supported  
- ✅ No existing modules will need to be rewritten or migrated  
- ✅ Composer-based modules are an **extra path**, not a replacement  

The goal here is to **expand possibilities without taking anything away**.

**Comment by ekdahl** [comment]
Nice work!
As a module developer, what steps are needed to publish a module, more than adding the JSON file like in your example?
Does this add more dependencies (Composer) for webtrees, or is Composer always installed when I have PHP in for example Debian?

**Comment by kiwi3685** [comment]
Hi @BertKoor 
I asked a similar question when this was first presented here (see earliest comment above). 
For my purposes this was answered very satisfactorily. 
But you should review the answer though, as you may be thinking of something I missed.

**Comment by magicsunday** [comment]
Hi @fisharebest, thank you so much for looking into this and creating the minimal installer – I really appreciate your effort!

I'd like to share my thoughts on why I believe the VendorModuleService approach offers significant technical advantages over the installer-based approach. To clarify, this isn't a core change, but rather a core extension. The existing code remains unchanged; only an additional path for module discovery is added.

A real-world use case: Development workflow. Developing and maintaining multiple interdependent modules using the modules_v4 approach is cumbersome in practice. My modules (fan-chart, pedigree-chart, descendants-chart) all depend on a common base package (webtrees-module-base). With modules_v4, any change to the base package requires manually copying or creating symbolic links – and synchronizing all modules is prone to errors.

With the VendorModuleService, Composer manages the entire dependency graph automatically. A `composer update` loads the correct version of each module and its dependencies into the `vendor/` directory, making it immediately ready for use. No manual file moving, no broken symbolic links, no version conflicts.

You can see this in action in my project https://github.com/magicsunday/webtrees-docker. There, the entire Webtrees stack (including custom modules) is managed through a single `composer.json` file. This works seamlessly only because modules and their dependencies reside in the `vendor/` directory, where Composer expects them to be.

Established pattern in the PHP ecosystem: This approach follows the pattern used by major frameworks. For example, TYPO3 uses the package type `typo3-cms-extension` – and in Composer mode, the core delegates the search for extensions to Composer's own mechanisms. Extensions remain in the `vendor/` directory and are discovered via Composer's runtime API. The VendorModuleService follows this exact principle, using `webtrees-module` as the package type.

1. Composer-native behavior: 
With the VendorModuleService, modules remain in the `vendor/` directory, where Composer manages them natively. The installation approach physically moves files to `modules_v4/`, which can lead to problems in various scenarios—for example, file permission issues, interrupted updates, symlink configurations, or in container environments.

2. PSR-4 autoloading works by default. 
Packages in `vendor/` are automatically loaded correctly via Composer's `autoload.php` using PSR-4 autoloading. Modules moved to `modules_v4/` are decoupled from the Composer autoloader and therefore require a workaround: registering their own class loader in `modules.php`:

```
$loader = new ClassLoader();
$loader->addPsr4('MagicSunday\\Webtrees\\FanChart\\', __DIR__ . '/src');
$loader->register();

```

With the VendorModuleService, this workaround is unnecessary, as the Composer autoloader already knows the module's namespace.

3. Dependency Resolution is Preserved
If a module in `vendor/` has its own Composer dependencies—such as my chart modules, which depend on `webtrees-module-base`—everything is resolved correctly. With the installer approach, the module itself is moved to `modules_v4/`, but its dependencies remain in `vendor/`. This separation can lead to problems with autoloading and paths, especially if the dependency tree is more than one level deep.

4. No mandatory installer dependency per module
The installer approach requires each individual module to declare a dependency on an installer package. With the `VendorModuleService`, a module only needs to include `"type": "webtrees-module"` in its `composer.json` – detection is handled by the kernel.

5. Clear separation of responsibilities
With the `VendorModuleService`, there is a clear boundary: `vendor/` is managed by Composer, while `modules_v4/` is managed manually. The installation approach mixes both in `modules_v4/`, which can lead to conflicts – for example, if someone manually deletes a module installed with Composer while it is still present in `composer.lock`.

I am happy to adapt the implementation if you have concerns about specific code sections. However, I am convinced that keeping the modules in `vendor/` and utilizing Composer's features is the more robust and future-proof solution.

magicsunday:
Hi @fisharebest, following up here — I think the discussion has surfaced both approaches clearly, so I'd rather not restate the technical arguments again.

Where we are, briefly:
- Your preference: keep core unchanged and let modules install into modules_v4/ via an installer (your minimal one, or a more comprehensive one).
- This PR: an additive core extension — modules_v4/ is untouched and keeps working exactly as today; it only adds a second discovery path so Composer-managed modules can stay in vendor/, where their dependency graph and PSR-4 autoloading already live. The two can coexist; this doesn't remove the installer route.

What would really help is a direction decision, so the PR doesn't sit open indefinitely:
- If you'd consider taking vendor/-based discovery into core — now, or with any changes you'd like — I'm glad to do the work.
- If you'd rather keep core out of it and standardise on the installer approach, just say the word and I'll close this, no hard feelings; I'll keep maintaining the service downstream for my own stack.

Either outcome is fine — I'd just like to know which direction to invest in. Thanks for your time on this.

PR #5231, opened 2025-09-17
Update to api v3

ddrury:
Following the HERE Limited Plan being discontinued, have updated the module to support the HERE Base Plan instead. Please note this requires the creation of a new api key


ddrury:
It turns out that if you have a key that was created prior to HERE maps v3 then whilst things seem to work they sometime go a bit crazy with the zoom levels, I've added a note on the config page

PR #5244, opened 2025-10-22
Only use solid icons and upgrade fontawesome package

ekdahl:
To get a more consistent experience, use only one icon style.
As the solid icons has almost 10x as many icons to choose from as the regular style, is seems wiser to use that.


fisharebest:
Note: font-awesome 7.0 makes all icons fixed width, so we can remove `fa-fw` after we upgrade.

PR #5249, opened 2025-11-01
SVG fact icons POC

ekdahl:
This is a proof of concept for adding fact icons as SVGs.
Is it a wanted feature?
How should it preferably be implemented?

Looks like this without theme adjustments:
image
image
image
image


RAMINCNZ:
IMHO, the choice of icon could be that of the Admin, with as Fredrik suggests simple EXAMPLE/BASIC icons being used as place holders. Debating Icons could go on forever given the number of WebTrees Users/Installs.

I'm not much for understanding the coding details for this, however, I would like to see that CSS ids, classes be used in the code where possible to make it easier to 'change' to suit.. 

A definite PLUS ONE from Me

Rick M
Nu Tirani - New Zealand

fisharebest:
I'm happy with this change in principle.  But first,

1. I want to review the list of facts/icons
2. I want to think more about where the facts are defined.  This could be in `class Fact` or in each element.

PR #5251, opened 2025-11-03
Add fontawesome children icon for individuals and families table

ekdahl:
Use the SVG icon in minimal theme


fisharebest:
Thanks!

PR #5256, opened 2025-11-15
Add light/dark mode switching

### LINKED ISSUE #5255: Set bootstrap color mode in ModuleThemeInterface
I'm doing some guessing here, since I don't have all the dots connected about how the whole webtrees app is "wired"...

Anyway, from Bootstrap 5.3 light or dark [color modes](https://getbootstrap.com/docs/5.3/customize/color-modes/) can be selected. I think it would be useful to be able to set this in the theme interface, to tell if the theme wants light or dark "base mode". This interface extension would then conditionally add `data-bs-theme="light"` or `data-bs-theme="dark"` to the html element.
This change would make it easier to implement dark themes for webtrees.

ekdahl:
Had a go at implementing #5255, not sure if it's the right way to do it.
Resolves #5255


ekdahl:
> I tried it, and IMHO it looks awful. I think the individual themes will need lots of work to be compatible with bootstrap dark mode.

I think you misunderstanding the purpose of this. The reason for adding this is for theme makers to be able to easily select the "base theme" from bootstrap to base their own theme on. By doing that, lots of variables will already have reasonable colors set for a dark theme and the theme CSS can be made much more compact.
Of course it will look bad with all the supplied themes in webtrees, as they are based on the "light" theme in Bootstrap, because that was the only one available until the color mode feature was introduced in Bootstrap 5.3.0.

ekdahl:
It is a bit tempting though to look at making a very plain "bootstrap" theme for webtrees with light and dark setting (or automatic switching) and as little CSS as possible.

Issue #5137, opened 2025-04-10
Enhancement: Re-order children - Sort by Date of Birth


sleebooth:
Inserting a child into a family always appends the child to the bottom of the list of children, even if the new child has a date of birth event. More often than not this necessitates reordering the children. Ideally the child would be inserted at the correct position chronologically. If this is not possible, improving work-flow as suggested below would be helpful:

1. At the moment, on the Family page, at the bottom of the children there is an option "Re-order children". 
2. Selecting this option takes the user to a list of children that can be sorted manually.
3. If there are a lot of children, it is necessary to scroll down to the bottom of the list of children to see the options.
4. There is an option "Sort by Date of Birth". which I use most often rather than the manual sort option.
5. Having chosen the latter option, the user can then select "Save".

6. On the Individual's Families tab, the option below the children is "Re-order Children". Steps 2-6 above result if it is selected.

I propose changing the text at stage 1 & 6 to "Re-order children manually" and adding an extra option below or alongside it: "Re-order children by Date of Birth". This would save one or two actions each time.

fisharebest:
Fixed in #5260, thanks to @BertKoor

PR #5260, opened 2025-11-25
fix #4511 & #5137: when adding a child to a family, put it between older and younger children

### LINKED ISSUE #4511: Auto order child
When adding a child informing birth date, using 2.1.6, the ordering of other children is not working. Then it is necessary use the reorder children button, that works fine.

To reproduce, include a child in a family, informing some birth date between two other children.

BertKoor:
Inspired on the algorithm of the 1.7 branch: put a new child after older children, just before the first child which is younger.

This saves the user from having to reorder or sort children after a child is added.

I did try to make some actual tests, avoid mocking everything, but alas got stuck on tree authorisations when creating a family or child.

### DIFF
--- a/app/Http/RequestHandlers/AddChildToFamilyAction.php
+++ b/app/Http/RequestHandlers/AddChildToFamilyAction.php
@@ -20,6 +20,8 @@ declare(strict_types=1);
 namespace Fisharebest\Webtrees\Http\RequestHandlers;
 
 use Fisharebest\Webtrees\Auth;
+use Fisharebest\Webtrees\Date;
+use Fisharebest\Webtrees\Fact;
 use Fisharebest\Webtrees\Family;
 use Fisharebest\Webtrees\Individual;
 use Fisharebest\Webtrees\Registry;
@@ -59,7 +61,7 @@ class AddChildToFamilyAction implements RequestHandlerInterface
         $child  = $tree->createIndividual("0 @@ INDI\n1 FAMC @" . $xref . '@' . $gedcom);
 
         // Link the child to the family
-        $before_id = $this->factIdOfYoungerSibling($family, $child);
+        $before_id = $this->childFactOfYoungerSibling($family, $child)?->id() ?? '';
         $family->createFact('1 CHIL @' . $child->xref() . '@', true, $before_id);
 
         $url = Validator::parsedBody($request)->isLocalUrl()->string('url', $child->url());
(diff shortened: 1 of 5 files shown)


app/Services/FamilyService.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public function addChildToFamily(Individual $child, Family $family): void
+    {
+        $child_birth_day = $child->getBirthDate()->julianDay();
+        $child_gedcom = '1 CHIL @' . $child->xref() . '@';
+        $family_facts = ['0 @' . $family->xref() . '@ FAM'];
+
+        // Insert new child at the right place
+        $done = false;
+        foreach ($family->facts() as $fact) {
```
fisharebest: You need to pass some extra parameters to this function.

By default, it will exclude private facts (we need to include them), and include deleted facts (we need to exclude them).

See https://github.com/fisharebest/webtrees/blob/c71f8c66ed90cbc9fbc0cde9b40f6719531d5341/app/GedcomRecord.php#L808

fisharebest:
I wonder if we need a new service for this?
Would it make sense to add this function to `GedcomEditService`?

app/Services/FamilyService.php
BertKoor: OK, good to know. I can fix that.
But in defense, my code was inspired on ReorderChildrenAction.php which uses plain ->facts()
There's another 36 occurrences of `->facts()` so perhaps these should be scrutinised.

BertKoor:
> I wonder if we need a new service for this? Would it make sense to add this function to `GedcomEditService`?

That was my initial thought as well: would simply adding a function there suffice?
Then I decided against that: all functions in `GedcomEditService` are very generic and tailored to abstract GEDCOM structures.
But perhaps you're right. There's already `newFamilyFacts` and `newIndividualFacts`, and a service with one public function is not doing much service. So you won me over.

Thanks for the review!

fisharebest:
`GedcomRecord::createFact()` now has a parameter `$before_id`.

https://github.com/fisharebest/webtrees/blob/85ca24a2537fd226aad3b96fc7dae008b2bfd51b/app/GedcomRecord.php#L747

This should let you simplify your code.  e.g. something like this...?

```php
$before_id='';
foreach ($facts(['CHIL']) as $fact) {
  if (this birth date is after the new child's birth date) {
    $before_id = $fact->id();
  }
}
createFact(..., $before_id);
```

fisharebest:
> It seems this recent change has broken current functionality on an individual to add to existing family. The child becomes member of the family, but the reverse link from indi to parents is not set. I've also somehow managed during testing to get the whole family purged.

I'll take a look...

fisharebest:
I fixed this issue.  Any similar ones will now trigger a deprecation (dev code) or just work (stable releases)

fisharebest:
A couple more comments.

Firstly, `$fact()->target()` might return any type of object, or it might return `null`. (This would mean the user has broken links in their data, which happens...).  If the returned object isn't an individual, we can't call `->getBirthDate()`.

So, we should change

```php
if ($child_birth_day < $fact->target()->getBirthDate()->julianDay()) {
    return $fact->id();
}
```

to

```php
$target = $fact->target();
if ($target instanceof Individual && $child_birth_day < $target->getBirthDate()->julianDay()) {
    return $fact->id();
}
```

Secondly (this is just a style thing, feel free to ignore if you disagree...).  We could use `Collection::first()` to find the first fact matching a condition.  e.g. something like this:

```php
        $filter = static function (Fact $fact) use ($child): bool {
            $target = $fact->target();
            return $target instanceof Individual && Date::compare($child->getBirthDate(), $target->getBirthDate()) < 0;
        };
        $before = $family
            ->facts(['CHIL'], false, Auth::PRIV_HIDE, true)
            ->first($filter);

        $family->createFact('1 CHIL @' . $child->xref() . '@', false, $before?->id() ?? '');
```

The reason I like this style is that there are no loops/conditionals, which are all sources of error, and cognitive complexity.

fisharebest:
Thanks!

FYI, I'm going to make a change the new parameter in `createFact()`
from `string $before_id = ''` to `Fact|null $before`.

This will let simplify the calling code a little.

PR #5274, opened 2025-12-27
Fix #5113: Add strikethrough extension to markdown

### LINKED ISSUE #5113: Add strikethrough extension to markdown

bschwede:
Fix #5113


bschwede:
Thanks for the insides. Your concerns are understandable. It makes sense not to lose sight of the big picture and don't get caught up in feature fever.
On the other hand IMHO some of those additional markdown features would support storytelling very well.
There is, of course, always a risk involved in adapting to other libraries. If a successor library does not support a feature,
the damage in terms of readability in Markdown is not so severe, but it is certainly not ideal. All in all, not easy!

Do you want to use the same editor for html and markdown notes or do you also consider a non-wysiwyg editor just with syntax highlighting?
If the second case is applicable: Do you know [tiny-markdown-editor](https://github.com/jefago/tiny-markdown-editor) (active since 09/2020)?

PR #5278, opened 2025-12-30
ignore case of media_file attributes in slideshow with postgres

### LINKED ISSUE #5277: [webtrees 2.2.4, postgres 16] Slide show on My page says no images to display despite media control showing media files
### Description

The "Slideshow" box on the user's homepage ("My Page") in Webtrees 2.2.4 always shows the message:

> "This family tree has no images to display."

Image

This happens **even though**:

- Image media objects (`.jpg`, `.jpeg`) exist
- Images are linked to individuals/events
- Images are visible on person pages
- User is logged in as administrator
- Media privacy settings are not restrictive
- File paths and types are correct
- Images are marked as type "image/jpeg"

The slideshow never displays any images, even after clearing cache and reloading.

- Fancy Image Bar also fails to display images (likely related), s. [[Webtrees 2.2.4, Postgres 16.8-1.pgdg120+1] Fancy Image Bar says no records to display despite media control showing media files](https://github.com/JustCarmen/webtrees-fancy-imagebar/issues/161)
- first mentioned in [combined ticket #5174 ](https://github.com/fisharebest/webtrees/issues/5174)

BertKoor:
fix #5277

### DIFF
--- a/app/Services/GedcomImportService.php
+++ b/app/Services/GedcomImportService.php
@@ -729,8 +729,8 @@ class GedcomImportService
         // Have we already created a media object with the same title/filename?
         $xref = DB::table('media_file')
             ->where('m_file', '=', $tree->id())
-            ->where('descriptive_title', '=', mb_substr($title, 0, 248))
-            ->where('multimedia_file_refn', '=', mb_substr($file, 0, 248))
+            ->where('descriptive_title', DB::iLike(), mb_substr($title, 0, 248))
+            ->where('multimedia_file_refn', DB::iLike(), mb_substr($file, 0, 248))
             ->value('m_id');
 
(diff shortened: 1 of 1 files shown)


BertKoor:
The issue is present as well with SQLite.
I've found more occurrences of problematic searches, and fixed some with case-insensitive `LIKE` operator.
If no `%` is present in the search string, it behaves as case-insensitive equals.

If you rather have two upper() (or lower) expressions for the database to evaluate, then that's an option as well. So to be clear it's an equals operator, not abuse of iLike.

fisharebest:
@BertKoor - see my comments in #5277

Thanks for the time you spent on this, but I would prefer a different solution.

Issue #5264, opened 2025-12-05
Feature request: upload indicator for files


HonkXL:
When uploading a file, for a time nothing happens. Sometimes users making a double click or clicking after a few seconds. It would be great to get any feedback after starting the upload. A text "upload in progress" or a nice animation would be great.

And no - files are not to large. This happens when uploading 2-4 MB.

Anke:
+1

As to the file size, wouldn't it be possible to automatically resize/compress picture files to preset maximum values in order to prevent that members upload photos with more than 8 MB?

kiwi3685:
No, webtrees can't change a file before it's been uploaded. 😀

fisharebest:
Actually, we can resize/resample a file before uploading it.  There are lots of ways to do this in the browser, using JavaScript.

But whether we should resize it is another question.  If an image is large because it is a high-resolution scan of a document with small handwriting, then it would be wrong to resample it.

However, this is unrelated to the upload-progress indicator, and should probably be raised as a separate issue.

Anke:
Sorry, I know, I may open a discussion for topics around file handling.

PR #5297, opened 2026-01-24
Use spinner as default loading icon

ekdahl:
This is an attempt to replace the loading icon that is used in the supplied themes today with a more modern spinner from bootstrap, that can be styled with the currentcolor CSS variable by the themes (or by replacing loading.phtml).
I'm abandoning the empty element approach that is used currently and instead adding icons/loading inside the element and let that be replaced when loading is done.

This is a POC that I welcome feedback on.
1. Is the change wanted?
2. Is the approach good or should something be adjusted?

Later I can add the same change for all usages of a loading icon.


fisharebest:
> How about places where we don't current have a spinner like this request while you're looking at these?
> 
> #5264

File uploads are completely different - so that issue will need another solution.  Probably a third-party library.

RAMINCNZ:
How about the initial Cookie Check as well??

Could there be, and a silly suggestion but, Sesame Street Cookie Bear munching a cookie instead of the text??

PR #5303, opened 2026-01-28
SlideShow module - configurable delay time as block setting

### LINKED ISSUE #5302: New Feature: Set delay on slide show
The only way to change the delay time on the slide show at the moment is to modify the code. This gets overwritten by an update of course. It would be good if it could be set in the settings tab.

BertKoor:
fix #5302: delay time of slides is made configurable as a block setting in a range from 1 second to 15 minutes.
Previous hard-coded delay time of 6 seconds is taken as default value.

### DIFF
--- a/app/Module/SlideShowModule.php
+++ b/app/Module/SlideShowModule.php
@@ -46,7 +46,7 @@ class SlideShowModule extends AbstractModule implements ModuleBlockInterface
     private const string LINK_INDIVIDUAL = 'indi';
 
     // How long to show each slide (seconds)
-    private const int DELAY = 6;
+    private const string DEFAULT_DELAY = '6';
 
     // New data is normalized.  Old data may contain jpg/jpeg, tif/tiff.
     private const array SUPPORTED_FORMATS = ['bmp', 'gif', 'jpeg', 'jpg', 'png', 'tif', 'tiff', 'webp'];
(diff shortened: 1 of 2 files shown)


app/Module/SlideShowModule.php
```diff
@@ -236,6 +238,7 @@ public function saveBlockConfiguration(ServerRequestInterface $request, int $blo
         $this->setBlockSetting($block_id, 'filter', Validator::parsedBody($request)->string('filter'));
         $this->setBlockSetting($block_id, 'controls', Validator::parsedBody($request)->string('controls'));
         $this->setBlockSetting($block_id, 'start', Validator::parsedBody($request)->string('start'));
+        $this->setBlockSetting($block_id, 'delay', (string)Validator::parsedBody($request)->integer('delay'));
```
fisharebest: Code style: add a space after type casts `(string)`

resources/views/modules/random_media/config.phtml
```diff
@@ -51,3 +52,12 @@ use Fisharebest\Webtrees\I18N;
         <?= view('components/radios-inline', ['name' => 'start', 'options' => [I18N::translate('no'), I18N::translate('yes')], 'selected' => (int) $start]) ?>
     </div>
 </div>
+
+<div class="row mb-3">
+    <label class="col-sm-3 col-form-label">
+        <?= I18N::translate('How long to show each slide (seconds)') ?>
+    </label>
+    <div class="col-sm-9">
+        <?= view('components/select-number', ['name' => 'delay', 'selected' => $delay, 'options' => [1, 2, 3, 4, 5, 6, 8, 10, 12, 15, 20, 30, 45, 60, 75, 90, 120, 150, 180, 240, 300, 360, 480, 600, 720, 900]]) ?>
```
fisharebest: Whatever options you have here, someone will want something different...

Perhaps just a simple `<input type="number" min="1" ...>`?

PR #5304, opened 2026-01-30
implement GEDCOM7 tag DATE:PHRASE

### LINKED ISSUE #5171: Support for GEDCOM7 tag DATE:PHRASE
Since GEDCOM v7.0 the datatype [DateValue](https://gedcom.io/terms/v7/type-Date) is supposed to have an optional subtag `PHRASE`. 

Examples:
````
1 BIRT
2 DATE 29 MAR 1964
3 PHRASE Easter Sunday of 1964

1 EVEN
2 TYPE custom
2 DATE
3 PHRASE 5 January (year unknown)
````

This subtag is not yet supported by webtrees. I'm working on the required code changes.

BertKoor:
fix #5171 

I thought I already had a PR for this, but could not find it anymore.


fisharebest:
Since the GEDCOM 5.5.1 spec says we can store date phrases in INDI:*:DATE:NOTE, I would probably implement this by adding support for inline notes here, with the label "Date phrase" to indicate that it is has a specific purpose and is not for general notes.

Then, if we add DATE:PHRASE, the user will have an edit form with two fields, both labelled as "Date phrase".

So, adding the g7 DATE:PHRASE would mean that we also need to remove the g5.5.1 DATE:NOTE.
We wouldn't want both at the same time.

BertKoor:
Why not... I have added tag `DATE:NOTE` labelled as `Note on date` (on par with `Note on place`) by default hidden but can be made unhidden via the GEDCOM Tags page of the Control Panel.
It's not labelled `Date phrase` because it is a different tag and I think semantically it serves a different purpose. A note can contain any free text while the date _phrase_ is meant for a payload _instead_ of a proper date, or giving a bit extra information. But that is open to interpretation.

I have borrowed the styling from what's currently done for explanatory text within the date value put at the end between brackets (not valid GEDCOM but who am I to judge that)
https://github.com/fisharebest/webtrees/blob/main/app/Date.php#L57-L60
The brackets are put in via CSS styling so it can overridden (e.g. made hidden)

PR #5307, opened 2026-02-03
CLI commands should have admin privilege

BertKoor:
Otherwise access to private trees and their private data is prohibited.

 Manual test scenario:
* Control Panel - tree
* Privacy - change "Show the family tree" to "Members only"
* make a GEDCOM export
* run `php ./index.php tree-list` should include `tree1`
* run `php ./index.php tree-export tree1`
* compare both GEDCOM files, should be identical (except timestamp)

### DIFF
--- a/app/Cli/Console.php
+++ b/app/Cli/Console.php
@@ -26,6 +26,9 @@ use Fisharebest\Webtrees\Registry;
 use Fisharebest\Webtrees\Session;
 use Fisharebest\Webtrees\Webtrees;
 use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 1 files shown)


fisharebest:
The security layer of users/accounts/roles is for web access.

To be able to run a CLI command, you must already have full access to the server and all files/data.

The CLI is designed for server admin tasks, not genealogy.

Effectively, they are just user-friendly shortcuts for things like:

```
mysql --database=webtrees -e "select * from wt_gedcom"
mysql --database=webtrees -e "select * from wt_user"
mysql --database=webtrees -e "update wt_user set email = 'whatever' where user_id = 7"
```

PR #5311, opened 2026-02-06
Cli commands should run with admin privilege

BertKoor:
Please reconsider you decision on PR #5307.
Usually your verdicts are right, but in this case I have strong doubts.


BertKoor:
too many conflicts

PR #5320, opened 2026-02-20
Ignore Visual Studio cache folder

ekdahl:
For people using VS, this folder gets created anytime the folder is opened in the editor, would be convenient to ignore it in Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@
 /tests/clover.xml
 /tests/coverage
 /vendor/
+
+.vs/


BertKoor:
Would you then also like to have added:

* .vscode/ for VS Code
* .idea/ for Jetbrains PHP Storm & IntelliJ
* .metadata/ and .settings/ for Eclipse

May I suggest you create a .gitignore file in your `home` directory, and put this in there. Fixes it for all projects.

PR #5328, opened 2026-02-28
Adjust base CSSs and minimal theme to use Bootstrap variables

ekdahl:
* Adjust minimal theme (and other base CSS files) to use CSS variables from Bootstrap instead of hardcoded colors, radius, thickness etc.
* This makes the minimal theme usable with both light and dark bootstrap themes, as implemented in #5256
* Next step would be to implement light/dark switching in the minimal theme.
* To test dark mode, add this attribute to the html element: `data-bs-theme="dark"`

light-start
dark-start
light-indi
dark-indi

### DIFF
--- a/resources/css/_base.css
+++ b/resources/css/_base.css
@@ -159,7 +159,7 @@ a:hover, .btn-link:hover, .nav-link:hover {
 }
 
 .wt-icon-spinner {
-    color: lightgrey;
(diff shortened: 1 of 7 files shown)


fisharebest:
@ekdahl - thanks!

Let me review this.  I have some small comments.

fisharebest:
Thanks!

PR #5346, opened 2026-03-26
Use small text in chart box to make more info fit

ekdahl:


fisharebest:
My initial thoughts are no.

webtrees inherited lots of `<small></small>` text from PhpGedView, and I gradually removed all of it for accessibility reasons.

Users with poor eyesight will increase the page zoom until the text is legible.
At this point, any "small" text is still unreadable.

fisharebest:
I am going to reject this.

You have two solutions, within webtrees itself.

Using the custom-css-js module, you can either reduce the text size, or increase the box size.

When I removed all the `small` text from webtrees, it was a deliberate decision, based on accessibility.

PR #5351, opened 2026-03-27
CLI compile-po-files - added pattern argument for providing custom path to po files

bschwede:
With reference to thread [convert mo/po - file to php-file](https://www.webtrees.net/index.php/forum/webtrees-help-and-support/40583-convert-mo-po-file-to-php-file?start=10#115926) it would be nice if this feature could also be used to prepare custom modules depoyment without having to resort to symbolic links or copy operations :-)

### DIFF
--- a/app/Cli/Commands/CompilePoFiles.php
+++ b/app/Cli/Commands/CompilePoFiles.php
@@ -21,6 +21,7 @@ namespace Fisharebest\Webtrees\Cli\Commands;
 
 use Fisharebest\Localization\Translation;
 use Fisharebest\Webtrees\Webtrees;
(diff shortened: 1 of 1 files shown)


bschwede:
This [example repository](https://codeberg.org/bschwede/wt-translations) might help custom module developers. There you will find an alternative po2php gulp/javascript implementation.

fisharebest:
I added options to specify both source and destination folders.

bschwede:
@fisharebest THX for making this tool more flexible 😄 

Is there a reason for the fixed pattern? Some projects have a flater naming pattern like `resources/lang/[lang].po`

PR #5359, opened 2026-04-06
Fix sequential script loading in webtrees.load()

magicsunday:
## Summary

- Fix race condition in `webtrees.load()` where inline scripts execute before external scripts have finished loading
- Replace `forEach` with a `Promise` chain that waits for each external script's `onload` event before appending the next script

## Problem

The `webtrees.load()` function (which replaced jQuery's `.load()`) dynamically injects `<script>` elements via `appendChild`. External scripts (with `src`) load asynchronously by default, but the `forEach` loop immediately appends the next script without waiting. This causes inline scripts that depend on a preceding external library to fail with errors like:

```
Uncaught TypeError: WebtreesFanChart.FanChart is not a constructor
```

This affects any custom module that uses the common pattern of pushing an external script tag followed by an inline initialization script via `View::push('javascript')`.

## Fix

Scripts are now loaded sequentially using a `Promise` chain. External scripts wait for their `onload` event before the next script in the chain is appended to the DOM.

## Test plan

- [x] Custom modules using `View::push('javascript')` with separate external and inline script blocks load correctly
- [x] Tested with magicsunday/webtrees-fan-chart, webtrees-pedigree-chart, and webtrees-descendants-chart


magicsunday:
For reference, this regression was introduced in commit a3ca8e1fef ("Convert jQuery code to VanillaJS"). The previous jQuery `.load()` implementation executed scripts sequentially, while the vanilla JS replacement uses `forEach` + `appendChild` which starts external script downloads without waiting for `onload` before appending the next script.

This affects all custom modules that use the common two-block pattern (`View::push('javascript')` with an external script followed by inline initialization).

magicsunday:
Our modules no longer trigger this race condition — we switched to dynamic `import().then()` inside a single `<script type="module">`.

Keeping this PR open since the fix in `webtrees.load()` would still benefit other custom modules that use external script tags via `View::push('javascript')`. Happy to rebase if needed.

fisharebest:
Thanks for this.

Looking again at this code, I see other issues:

* the `<script>` detection is incomplete.
* we delete the node elements from the parser, not the dom-document that we are about to insert.

I think, perhaps, this function needs to be rewritten from scratch.

fisharebest:
> I think, perhaps, this function needs to be rewritten from scratch.

I've taken your code, and other versions of `load()` replacement, and merged/simplified them.

Thanks again.

PR #5364, opened 2026-04-20
Reports - sort by place

BertKoor:
Requested on the [forum](https://www.webtrees.net/index.php/forum/webtrees-help-and-support/40593-add-extra-sort-order-for-burials-report): add the option in Reports of Birth - Death - Burials & Marriage to sort on their place.
The sorting is performed on the reversed parts of hierarchical place names.

### DIFF
--- a/app/Comparators/FamilyComparator.php
+++ b/app/Comparators/FamilyComparator.php
@@ -28,4 +28,9 @@ final readonly class FamilyComparator
     {
         return Date::compare($first->getMarriageDate(), $second->getMarriageDate());
     }
+
+    public static function byMarriagePlace(Family $first, Family $second): int
+    {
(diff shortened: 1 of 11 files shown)


fisharebest:
Hi Bert.  These changes are mostly OK.

I was already looking at functions such as `Individual::getBirthDate()`, and the comparison functions.

These are growing, and don't strictly belong in the indi/fam classes; they don't need access to any internal/private methods, and are just syntactic shortcuts.

They are also inconsistent, as getDeathPlace() behaves subtely differently to your new getBurialPlace().

I want to look at these before merging your PR.

BertKoor:
Sure, I've just copy-pasted what was there for dates, expanding it to places, making the smallest possible change to make it work. Happy refactoring!

PR #5375, opened 2026-05-13
Divert to setup wizard's admin step when config.ini.php is pre-written

magicsunday:
## Problem

Headless and containerised installs (Docker images that inject DB credentials via environment variables, for example) typically ship a pre-written `config.ini.php`. With `config.ini.php` in place, the existing `ReadConfigIni` middleware bypasses the setup wizard entirely, so an install that has not yet created an administrator account responds `404` to every route — there is no admin to authenticate, no tree to show, and the wizard is unreachable because the config file already exists.

The operator has no way out short of deleting `config.ini.php` and retyping the DB credentials they already provided through the environment.

## Change

A new `BootstrapAdminWizard` middleware runs immediately after `UpdateDatabaseSchema`. When `config.ini.php` was loaded (a `dbhost` attribute is present on the request) and the `user` table contains no administrator, it diverts the request to the wizard's administrator step (step 5).

`SetupWizard::userData()` now also reads request attributes as fallback defaults, so the hidden DB fields in step 5 are seeded from `config.ini.php` and step 6 can finish the install without asking the operator to retype credentials they already provided through the environment.

`SetupWizard` is unsealed (`final` → open) so the new middleware test can substitute it with a stub. Routing pipelines keep working unchanged otherwise:

- Fresh install, no `config.ini.php` → routed to step 1 by `ReadConfigIni` (unchanged).
- Install with populated user table → no behaviour change.
- Install with `config.ini.php` but no admin → now routes to step 5 instead of a blanket `404`.

## Tests

- `BootstrapAdminWizardTest`: four behavioural cases (pass-through without `dbhost` attribute, pass-through when admin exists, divert to wizard when admin missing, divert when only non-admin users exist).
- `SetupWizardTest`: covers the new `userData()` fallback precedence (body → request attribute → hard-coded default).

Targeted `tests/app/Http/` run: 478 / 478 green.

## Compatibility

No schema changes, no config file changes, no removed APIs. The only API-shape change is `SetupWizard` losing its `final` modifier; downstream subclassing is not encouraged but the existing surface stays identical.

### DIFF
--- a/app/Http/Middleware/BootstrapAdminWizard.php
+++ b/app/Http/Middleware/BootstrapAdminWizard.php
@@ -1,0 +1,90 @@
+<?php
+
+/**
+ * webtrees: online genealogy
+ * Copyright (C) 2026 webtrees development team
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+declare(strict_types=1);
+
+namespace Fisharebest\Webtrees\Http\Middleware;
+
+use Fisharebest\Webtrees\Contracts\UserInterface;
+use Fisharebest\Webtrees\DB;
+use Fisharebest\Webtrees\Http\RequestHandlers\SetupWizard;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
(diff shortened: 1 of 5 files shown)


magicsunday:
Closing this PR — internal audit surfaced findings that make the approach unsuitable for upstream:

- The middleware runs **before** `UseSession` and `Router`, so neither the session layer nor route-scoped `CheckCsrf` protects the diverted POST. Combined with the wizard's `Auth::login()` in `step6Install`, this opens a pre-auth admin-takeover window on every deployment that ships `config.ini.php` without an admin yet — wider than the existing setup-wizard window.
- The reusable shape webtrees already exposes (the docker entrypoint already writes `config.ini.php` from env vars and gates a headless admin bootstrap) makes this fixable entirely downstream: at install time, if no admin exists, the entrypoint can create one (random password, surfaced in the install banner). No new web-reachable bootstrap path, no `final` removal, no extra middleware.
- Several smaller concerns (sentinel coupling to `ReadConfigIni`'s attribute namespace, `CONFIG_INI_KEYS` placement, request-attribute injection cascade in `SetupWizard::userData`, reinvention of `UserService::administrators()`) reinforce that the right home is the deployment layer, not core.

Apologies for the noise. Will solve in the downstream docker repo.

fisharebest:
There are now CLI scripts available to create trees/users.

These were designed to support scripted installation.

`php index.php list`

Do these provide the features you need?  e.g. creating initial admin user, initial tree, etc?

PR #5385, opened 2026-05-28
access to calendar menu restricted to members by default

BertKoor:
We know that the calendar page in webtrees can work as a tarpit for bots, and imho it will be a matter of time before bots will find a way to break through the defense in our BadBotChecker.

As a preparation the following is changed with this PR:

* The default access level of the `CalendarMenuModule` is changed from visitor to member, so visitors do not get a link to the Calendar in the main menu. Administrators can explicitly grant access to visitors via the Control Panel if they want.
* The `Date` class only renders links to the calendar page when the current user has access to the calendar module.
* The `CalendarPage` itself checks access for the current user to the calendar module, in order to not rely on availability of a link.

These changes should definitely close the calendar tarpit for unauthenticated users.

### DIFF
--- a/app/Date.php
+++ b/app/Date.php
@@ -21,6 +21,9 @@ namespace Fisharebest\Webtrees;
 
 use Fisharebest\ExtCalendar\GregorianCalendar;
 use Fisharebest\Webtrees\Date\AbstractCalendarDate;
+use Fisharebest\Webtrees\Module\CalendarMenuModule;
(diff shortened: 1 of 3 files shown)


kiwi3685:
Excellent proposal. 
It's very similar to what I've been doing on my own site manually, although I simplistically set links to admin only. No user has complained, and I have more than 120!
It has solved my ai attack problems very well.

fisharebest:
Thanks!

PR #5393, opened 2026-06-04
better handling of bad requests

BertKoor:
A truely bad request results in this error message:
```
I18N::$translator must not be accessed before initialization
```

While validating a http request the `I18N` class is not yet initialised and there's no connection yet with the database.
The error handler thus should not depend on either.

Changes in this PR make sure that at least a stack trace with the appropriate error message is shown to the user.

This can be tested with e.g. changing `client-ip` to another string e.g. `client-eyepea` on line 1530 of `BadBotBlocker`.

### DIFF
--- a/app/Http/Middleware/HandleExceptions.php
+++ b/app/Http/Middleware/HandleExceptions.php
@@ -19,6 +19,7 @@ declare(strict_types=1);
 
 namespace Fisharebest\Webtrees\Http\Middleware;
 
+use Error;
 use Fig\Http\Message\RequestMethodInterface;
 use Fig\Http\Message\StatusCodeInterface;
 use Fisharebest\Webtrees\Http\Exceptions\HttpException;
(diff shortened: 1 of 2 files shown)


fisharebest:
Hi Bert.  Thanks for identifying this issue.

I think that these validation error messages are in the same category as PHP errors.
They are coding errors (or possibly hacking attempts).  So, don't really need translating.

I notice that some messages are translated and some are not.

So, I think a better solution is to remove `I18N::translate()` from the `Validator` class.

Do you want to update the PR to do that instead?

BertKoor:
Sure I can, but next Tuesday is the earliest opportunity I have. Will check where the other validations are called from.

Did I mention that on the forum somebody reported this particular validation error went off on his home network. He used ipv6 with something extra added which logicallt should be stripped off.
https://www.webtrees.net/index.php/forum/4-customising/40680-small-code-change-to-handle-ipv6-lla#116601
No way I can reproduce that.

fisharebest:
> next Tuesday is the earliest opportunity I have

No problem.  I updated this.

BertKoor:
I've merged back your recent `main` branch into [mine](https://github.com/BertKoor/webtrees/tree/premature_I18N). Found (and fixed) two instances of `HttpBadRequestExceptions` with `I18N` messages.

FYI: the try-catch I added in method `httpExceptionResponse` of `HandleExceptions.php` is imho essential. When a `HttpBadRequestException` is thrown then there's no db connection yet, so `Site::getPreference` will fail. That error is anticipated, so simply displaying a stack trace will suffice.

Issue #5391, opened 2026-06-02
General search includes individuals and families if search for "Locations" is selected only


Jefferson49:
Observed with: webtrees 2.2.6

How to reproduce: Open search and select "Locations" while un-selecting "Individuals" and "Families".

<img width="1829" height="249" alt="Image" src="https://github.com/user-attachments/assets/b65b5f93-cc65-4b35-8c83-bdffd854528e" />  

.  
Seemingly, the following code in [SearchGeneralPage.php](https://github.com/fisharebest/webtrees/blob/main/app/Http/RequestHandlers/SearchGeneralPage.php#L95) does not evaluate $search_locations and activates the search for individuals and families.
```PHP
        // Default to families and individuals only
        if (!$search_individuals && !$search_families && !$search_repositories && !$search_sources && !$search_notes) {
            $search_families    = true;
            $search_individuals = true;
        }
```

BertKoor:
Is this with a specific 3rd party module enabled?

Locations are not "first class" GEDCOM objects. Yet. And I don't see such a checkbox on our demo site.

<img width="1080" height="588" alt="Image" src="https://github.com/user-attachments/assets/dc81803b-a856-4494-8d55-df4792f8db28" />

reteP-riS:
These locations are the _LOC records from GEDCOM-L which can be activated in the Control Panel.

Jefferson49:
> These locations are the _LOC records from GEDCOM-L which can be activated in the Control Panel.

Yes, it is about _LOC records.

> Is this with a specific 3rd party module enabled?

In my opinion not. While creating the screenshot above, I disabled "Vesta shared places" (for _LOC) and the checkbox was still shown

> Locations are not "first class" GEDCOM objects. Yet. And I don't see such a checkbox on our demo site.

As far as I can see, the checkbox is shown depending on whether _LOC records exist in the database:
+ The GeneralSearch request handler checks whether _LOC records exist in the database and stores this information in a variable [$exist_locations](https://github.com/fisharebest/webtrees/blob/main/app/Http/RequestHandlers/SearchGeneralPage.php#L80)
+ The view in [search-general-page.phtml](https://github.com/fisharebest/webtrees/blob/main/resources/views/search-general-page.phtml#L110) decides whether to show the checkbox based on the status of $exist_locations

BertKoor:
I think I understand the issue now. 

Could someone help me obtain an example .ged file with these GEDCOM-L structures?
I searched, but found nothing.
[...]

PR #5398, opened 2026-06-13
fix #4981 - let General Search look in Shared Notes by default as well

### LINKED ISSUE #4981: Search Should Include Shared Notes
https://github.com/fisharebest/webtrees/blob/32bac9818dbdbb42ec586b03059aea4dcbbadb27/app/Http/RequestHandlers/SearchGeneralPage.php#L112

I figured out that I need to hack this block of code to change the default search scope.  I need it to include shares notes.

Would there be any harm in changing this in the next version too?

BertKoor:
This PR checks the Shared Notes box by default.
This is not so much needed when using the _General Search_ function.
But when using the _Quick Search_ function, then Shared Note records will be included in the search.

this fixes #4981 and #5391


fisharebest:
This PR is doing a few things, and we have merge conflicts.  Let me extract them individually to simplify the merge.

fisharebest:
Thanks.  Changes merged.

PR #5402, opened 2026-06-20
Remove references to deleted 32x32 loading image

jon48:
Commit a30e37c5 removes the `loading-32x32.gif` image, but it is still referenced here and there in the CSS files.

In terms of impact:
- `.wt-ajax-load` disappeared as part of the previous commit, so no impact expected
- I could not find any reference to the class `.icon-loading-large` in `minimal.css`, so no impact expected
- Colorbox already comes with its own loading icon, so it actually fixes the loading image that stopped appearing.

I will let you build the CSS bundles.

### DIFF
--- a/resources/css/_colorbox.css
+++ b/resources/css/_colorbox.css
@@ -99,10 +99,6 @@
     text-align: center;
 }
 
-#cboxLoadingGraphic {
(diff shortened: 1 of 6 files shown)


fisharebest:
Thanks!

PR #5405, opened 2026-06-21
Update guzzlehttp/guzzle to 7.12.1 to resolve two security advisories

magicsunday:
The currently pinned `guzzlehttp/guzzle` 7.11.0 is affected by two security advisories (both medium), fixed in 7.12.1:

- *Dot-only cookie domains match all hosts* (`PKSA-93qv-9n9h-6k6p`, affected `< 7.12.1`)
- *Silent HTTPS proxy downgrade to cleartext* (`PKSA-k22t-f949-t9g6`, affected `< 7.12.1`)

This bumps the exact pin `7.11.0` → `7.12.1` and its transitive `guzzlehttp/psr7` `2.11.0` → `2.12.1`. After the bump `composer audit` reports no advisories.

Because guzzle is pinned to the exact affected version, Composer's advisory-blocking policy (`policy.advisories.block`, default in Composer 2.10+) currently refuses to resolve it, so a clean `composer install`/`update` fails until the pin moves to 7.12.1.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
         "fisharebest/algorithm": "1.6.0",
         "fisharebest/ext-calendar": "2.6.0",
         "fisharebest/localization": "1.17.0",
-        "guzzlehttp/guzzle": "7.11.0",
+        "guzzlehttp/guzzle": "7.12.1",
         "illuminate/database": "13.14.0",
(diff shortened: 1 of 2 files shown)


fisharebest:
Thanks!

Issue #5379, opened 2026-05-21
New column "imported at" in wt_gedcom


webtrees-pesz:
Please add a new column titled “imported at” to the wt_gedcom database table, as suggested. 

See https://www.webtrees.net/index.php/forum/webtrees-help-and-support/40654-how-to-find-out-in-the-database-when-a-tree-was-created-or-imported-last-time#116378

ddrury:
Not needed - just add an entry in the gedcom_setting table

BertKoor:
Peter_S confirmed on the forum that a record written to wt_log would work well.
https://www.webtrees.net/index.php/forum/webtrees-help-and-support/40696-webtrees-db-wt-gedcom-imported#116693

PR #5407, opened 2026-06-24
write a log record when tree import succeeded or failed

BertKoor:
this fixes issue #5379


ddrury:
Wouldn't the edit log be a better bet than the config log

BertKoor:
> Wouldn't the edit log be a better bet than the config log

Good question! My consideration to determine which to use, is that an editor cannot do an import. It can only be done by an administrator via the control panel. And thus the `config` section is imho a better fit.

I did change the log records written when an import fails to the `error` type.

Issue #5388, opened 2026-05-31
NotFound silently drops original URL when unauth user hits a tree-private route → login redirect loses deep-link target


jbh4x82:
Summary
When an unauthenticated user clicks a deep-link into a private tree (e.g. an email link to /tree/meran.ged/forum/topic/3594), they are redirected to the login page, but the original URL is silently lost. After successful login they land on the home page, not the page they originally clicked. This breaks any "click this link to see the post" workflow for sites with private trees.

Steps to reproduce
Site with a private (gedcom.private = 1) tree.
Log out (or open an incognito window with no session).
Visit any deep tree-URL, e.g. /index.php?route=/tree/<name>/individual/I1 or any custom-module tree-URL.
Observe redirect to /index.php?route=/login&url= (note: empty url query param).
Sign in.
Land on the home page instead of the original target.
Root cause
TreeService::all() (app/Services/TreeService.php) filters private trees out for users without access. In Router::process() (app/Http/Middleware/Router.php), when the matched route requires {tree} but tree_service->all()->get($value) returns null, Router calls return $handler->handle($request) to short-circuit the matched handler.

The next handler in the chain is eventually NotFound::handle() (app/Http/RequestHandlers/NotFound.php). For a GET request, NotFound does:

return redirect(url: route(route_name: HomePage::class));
The original request URI is not preserved. The user then hits HomePage, which redirects to LoginPage with ['url' => ''], and LoginAction's redirect($url) ends up redirecting to HomePage's URL (the default).

Suggested fix
In NotFound::handle(), before the catch-all redirect to HomePage, detect the unauth-hitting-tree-URL case and redirect to LoginPage with the original URI preserved as ?url=:

if (Auth::id() === null) {
    $uri  = $request->getUri();
    $path = $uri->getPath();
    $url_route = (string) ($request->getQueryParams()['route'] ?? '');
    if (str_contains($path, '/tree/') || str_contains($url_route, '/tree/')) {
        return redirect(route(LoginPage::class, ['url' => (string) $uri]));
    }
}
LoginAction already reads url from the POST body and redirect($url) on success, so this closes the round-trip without further changes.

Same pattern is used by AuthLoggedIn and AuthMember middlewares for their auth-required routes — extending it to the tree-private fall-through is consistent with the existing webtrees design.

Tested on
[...]

PR #5409, opened 2026-06-27
Preserve deep-link when a guest hits a private-tree URL (#5388)

jbh4x82:
Fixes the deep-link loss described in #5388.

## Problem
When an unauthenticated visitor follows a deep link into a **private** tree (e.g. an email link to `/tree/<name>/individual/I1`, or any custom-module tree URL), `TreeService::all()` hides the tree from them, so `Router::process()` reaches its `{tree}` fall-through and defers to `NotFound`, which `redirect()`s to the home page. The original URL is discarded, so after signing in the visitor lands on the home page instead of the page they clicked. This breaks "click this link to see the post" workflows on sites with private trees.

## Why here, and not in `NotFound`
As you noted in #5388, rendering a 404 *in place* at the tree URL would be a dead-end (no valid tree → no genealogy menu). This change avoids that by redirecting guests to the **login page** carrying the requested URL, which is a normal navigable page. `Router::process()` is also where the information actually is: we already know the matched route required `{tree}` and that the tree was not resolvable for this user, so `NotFound` itself stays stock.

## What it does
At the `{tree}`-null branch:

- **Guest** → `redirect(route(LoginPage::class, ['url' => (string) $request->getUri()]))`. This is the same pattern `AuthLoggedIn` and `AuthMember` already use for auth-required routes. `LoginPage` reads that URL and `LoginAction` redirects to it on success, both through `isLocalUrl()`, so it is **not** an open redirect. It also avoids disclosing to a guest whether a private tree exists: a missing tree and a private-but-inaccessible tree are handled identically.
- **Logged-in user** → unchanged (falls through to `NotFound` as before). Signing in again would not help, and we must not mislabel a genuinely missing tree as access-denied.

## Testing
- webtrees `main` + `php -l` on this branch; behaviour confirmed on 2.2.6 / PHP 8.3.
- Guest deep-link to a private tree (individual, family, and a custom-module route) → login page → after sign-in lands on the originally-requested URL.
- Guest deep-link to a **public** tree → unchanged (tree resolves, no redirect).
- Logged-in member without access → unchanged.
- Non-`{tree}` 404s (typos, unknown routes) → unchanged.

Happy to adjust. For example, if you'd prefer the logged-in-but-no-access case to render `HttpAccessDeniedException` rather than fall through, that's a small addition, though it needs distinguishing a private tree from a missing one (an extra `gedcom`-table lookup), which is why I left it out here.

### DIFF
--- a/app/Http/Middleware/Router.php
+++ b/app/Http/Middleware/Router.php
@@ -24,7 +24,9 @@ use Aura\Router\RouterContainer;
 use Aura\Router\Rule\Accepts;
 use Aura\Router\Rule\Allows;
 use Fig\Http\Message\StatusCodeInterface;
+use Fisharebest\Webtrees\Auth;
 use Fisharebest\Webtrees\Http\Dispatcher;
+use Fisharebest\Webtrees\Http\RequestHandlers\LoginPage;
 use Fisharebest\Webtrees\Registry;
 use Fisharebest\Webtrees\Services\ModuleService;
 use Fisharebest\Webtrees\Services\TreeService;
@@ -37,6 +39,8 @@ use Psr\Http\Server\RequestHandlerInterface;
 
 use function explode;
 use function implode;
+use function redirect;
+use function route;
 use function str_contains;
 
 readonly class Router implements MiddlewareInterface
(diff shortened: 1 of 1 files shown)


fisharebest:
This is still a redirect on 404, which I don't really like.

Instead, we could show the 404 at the requested URL.  Since we didn't match a route, there will be no `$tree`, and no genealogy pages.  Just the login/register link at the top.

But after the message "This page does not exist or you do not have permission to view it", we could add a "continue" button, which takes the user to the home page.

The result will be:

* your users can login from this page, and then be automatically redirected back to it.
* it won't be a dead-end.  The continue button takes you to where you were previously redirected.
* robots will see the 404, rather than 302 + 200.

fisharebest:
#5388 was closed via 08cf2b742d

PR #5410, opened 2026-07-01
No I18N for HttpBadRequestExceptions

BertKoor:
also a HttpException can be thrown before database connection is made: return a simple stack trace dump.

continuation of where https://github.com/fisharebest/webtrees/pull/5393 left off


fisharebest:
There are two distinct changes here.

1) Remove I18N from validation messages.  This seems sensible.

2) Catch errors thrown during bootstrap.  What sort of errors?  How might I introduce such an error to test this?

BertKoor:
> How might I introduce such an error to test this?

I tested it by letting the BadBotBlocker validate attribute `client-eyepea` instead of `client-ip`
https://github.com/fisharebest/webtrees/blob/437eddf9c915ae9a04a14a5f7d4e459178c69ba2/app/Http/Middleware/BadBotBlocker.php#L1530

fisharebest:
The exception handler has lots of different cases, guessing where the exception was thrown, what information might be available (e.g. routing), etc.

I am wondering if there is a simpler option.

The exception handler has various levels.  It tries to render full pages, assuming the application has booted, then simpler pages, and then a simple stack-trace.

We could split this into two (or more?) exception handlers.  The lowest level one would catch bootstrap errors, and show a simple stack-trace.

The next one (before loading modules?) would show an error page with all modules disabled.

The final one, before routing/controller would show a full webtrees error page, with menus, modules, etc.

Now, if we get an error trying to render the "full" error page, we just re-throw the error, and the next exception handler down the stack will try.  If that can't display the page, we end up at the basic stack-trace handler.

fisharebest:
This seems to have worked quite nicely :-)

We now have two ExceptionHandler middleware classes.  One handles bootstrap exceptions. The other handles application exceptions.

If there's a problem with the application handler, then the bootstrap handler will take care of it and display a stack dump

I also updated the display to include chained exceptions.  So if there is an exception while rendering an exception, we'll get two stack-traces in the output.

I also stripped out namespaces from fully-qualified-class-names in the stack dump, which makes them a lot more readable.

Fixed in 856001b35c2ccf0e3ad4ef3965053d4247992536

PR #5426, opened 2026-07-21
Respect explicit privacy restrictions even when HIDE_LIVE_PEOPLE is disabled

gnilebein:
Previously, disabling the tree's "hide living people" setting caused canShowRecord() to return true immediately, bypassing RESN restriction notices and individual-level privacy overrides for every record. Move that check after the explicit restriction checks so an individual marked private is still hidden, even when living people are otherwise shown to visitors.

### DIFF
--- a/app/Age.php
+++ b/app/Age.php
@@ -19,6 +19,8 @@ declare(strict_types=1);
 
 namespace Fisharebest\Webtrees;
 
+use Fisharebest\Webtrees\Enums\DateType;
+
 use function max;
 use function preg_match;
 use function str_starts_with;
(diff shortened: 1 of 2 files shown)


fisharebest:
Hi @gnilebein 

This is an interesting change!

We are approaching a major version release, so this is a good time to make such a change.

We would need to update the GUI to make this option clear.

Once that is done, I would want to discuss/explain the change on the project forum.

fisharebest:
FYI: https://www.webtrees.net/index.php/forum/2-open-discussion/40741-changes-to-privacy-logic

PR #5431, opened 2026-07-26
Use narrow no break space as thousands separator

ekdahl:
All these countries have narrow no break space here: https://www.localization.guide/directory


fisharebest:
Can you update the corresponding test scripts as well?  These use actual no-break-space characters, not the constants that are used in the langage files themselves.

ekdahl:
Sure, I didn't notice there were tests to update.

PR #5449, opened 2026-08-13
fix #5417: case-insensitive comparison of NAME:TYPE and FAM:MARR:TYPE in reports

### LINKED ISSUE #5417: Tag:TYPE conditions in reports should be case insensitive
This issue was reported in the forum:
https://www.webtrees.net/index.php/forum/webtrees-help-and-support/40721-type-of-marriage-in-source-reports

Since v2.1.0 (issue #4308) we store `TYPE` tags in upper case, while the old or imported values can be in lower or camel case.

I found some instances where the reports check the value of the TYPE of a fact/event. These conditions should be done case insensitive:
https://github.com/fisharebest/webtrees/blob/3661fe4a96eb321c2a694d137df9ba5d872d9cd5/resources/xml/reports/fact_sources.xml#L204-L210
https://github.com/fisharebest/webtrees/blob/3661fe4a96eb321c2a694d137df9ba5d872d9cd5/resources/xml/reports/death_report.xml#L66

Since this is coded in xml, it's probably best to supply the `TYPE` tag in canonical form. Alas I did not find where / how this data exactly flows (otherwise this would be a PR :)

BertKoor:
fixes #5417

### DIFF
--- a/app/Report/ExpressionLanguageProvider.php
+++ b/app/Report/ExpressionLanguageProvider.php
@@ -34,6 +34,7 @@ final class ExpressionLanguageProvider implements ExpressionFunctionProviderInte
     {
         return [
             ExpressionFunction::fromPhp('stristr'),
+            ExpressionFunction::fromPhp('strcasecmp'),
         ];
     }
 }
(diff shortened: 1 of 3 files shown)


fisharebest:
This was fixed in 217f7c50c8ce227dd11ae409a01ff52680e81b1e using the suggested solution in #5417

PR #5460, opened 2026-08-18
fix exception when SetupWizard was about to create a tree

### LINKED ISSUE #5446: IMPORTANT: main branch v2.3.0-dev is unstable
Original report: 
**exception when creating a new tree**

Using latest code pulled from the `main` branch, an exception is thrown when Setup Wizard creates a new tree, or when the admin wants to create a new tree on the Control Panel. That is whenever `/index.php?route=%2Fadmin%2Ftrees%2Fcreate` is visited.

Part of the stack trace:
```
Undefined array key 1 …/resources/views/admin/trees-create.phtml:44
#0 …/resources/views/admin/trees-create.phtml(44): Fisharebest\Webtrees\Http\Middleware\ErrorHandler->errorHandler(2, 'Undefined array...', '/Users/bertkoor...', 44)
#1 …/app/View.php(163): include('/Users/bertkoor...')
#2 …/app/View.php(239): Fisharebest\Webtrees\View->render()
#3 …/app/Helpers/functions.php(103): Fisharebest\Webtrees\View::make('admin/trees-cre...', Array)
#4 …/app/Http/ViewResponseTrait.php(47): view('admin/trees-cre...', Array)
#5 …/app/Http/Controllers/CreateTree.php(52): Fisharebest\Webtrees\Http\Controllers\CreateTree->viewResponse('admin/trees-cre...', Array)
```

Error line:
https://github.com/fisharebest/webtrees/blob/main/resources/views/admin/trees-create.phtml#L44

Suspicious non-existing import:
https://github.com/fisharebest/webtrees/blob/main/resources/views/admin/trees-create.phtml#L7

Because this is what the form looked like, including an "example" url which cannot be dysected into two parts:

Image

BertKoor:
this fixes #5446 
Same exception was thrown when rewrite_url=0
Avoided the problematic split by chosing another value than `tree` and without characters which need escaping.
Screenshot 2026-08-18 at 16 46 44

### DIFF
--- a/resources/views/admin/trees-create.phtml
+++ b/resources/views/admin/trees-create.phtml
@@ -35,13 +35,13 @@ use Fisharebest\Webtrees\I18N;
         <div class="col-sm-10">
             <div class="input-group" dir="ltr">
                 <span class="input-group-text" dir="ltr">
-                    <?= e(explode('{tree}', rawurldecode(route('example', ['tree' => '{tree}'])))[0]) ?>
+                    <?= e(explode('arbre', rawurldecode(route('example', ['tree' => 'arbre'])))[0]) ?>
                 </span>
 
                 <input class="form-control" id="name" maxlength="31" name="name" pattern="[^&lt;&gt;&quot;*?{}():/\\$%|]*" required="required" type="text" value="<?= e($tree_name) ?>" dir="ltr">
 
                 <span class="input-group-text" dir="ltr">
-                    <?= e(explode('{tree}', rawurldecode(route('example', ['tree' => '{tree}'])))[1]) ?>
+                    <?= e(explode('arbre', rawurldecode(route('example', ['tree' => 'arbre'])))[1]) ?>
                 </span>
             </div>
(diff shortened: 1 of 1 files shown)


fisharebest:
This code appears twice.  The preferences page and the new-tree page.

When the code was refactored, only the preferences page was updated.

I just updated the new-tree page to use the same code about an hour ago...

Same principle as yours, but uses the `{tree}` placeholder.  Using text that could legitimately appear in a URL, such as `arbre`, could lead to unexpected behaviour.

BertKoor:
> I just updated the new-tree page to use the same code about an hour ago...

Where's the commit with that fix exactly? I've pulled `main` multiple times, but still get this error.
Tried to fix it myself using the old placeholder value `{tree}`, but this needs a condition depending on `rewrite_urls`.
When rewrite_urls=0 then explode on `rawurlencode('{tree}')`, otherwise explode on `{tree}`
The encoding differs drastically depending on whether it is part of the path or a query param.

Issue #5445, opened 2026-08-11
demo-dev: too many redirects


BertKoor:
https://dev.webtrees.net/demo-dev/ redirects to itself, apparently not happy with the cookie `x=y` it had set.

<img width="997" height="973" alt="Image" src="https://github.com/user-attachments/assets/43e8098e-ad99-47aa-97aa-6d7a04c4d69a" />

fisharebest:
Fixed by #5462 - (thanks @sherburne)

PR #5462, opened 2026-08-18
Fix for Issue 5461 "URL redirect"

sherburne:
Strips the base path prefix from the route parameter in `RouteFactory::route()` before building the URL.

    // In RouteFactory::route(), after extracting the path:
    $base_path = parse_url($base_url, PHP_URL_PATH);
    $base_path = is_string($base_path) ? $base_path : '';
    if ($base_path !== '' && str_starts_with((string) $path, $base_path)) {
        $path = substr($path, strlen($base_path));
    }

This produces `?route=/tree/test` instead of `?route=/test/tree/test`, which `RouteMatcher` can match correctly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@
 /robots.txt
 
 /vendor/
+
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,3 @@
 /robots.txt
 
 /vendor/
-
-.DS_Store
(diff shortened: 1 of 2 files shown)


fisharebest:
Thanks!

This also fixes #5445

fisharebest:
Tip:  If you put `Fix: #nnnn` in the PR title, then merging the PR will automatically close the issue.

PR #5479, opened 2026-08-21
fix #5478: edit raw GEDCOM record, missing param level0

### LINKED ISSUE #5478: 2.3.0-dev: edit raw GEDCOM fails with `The parameter “level0” is missing`
The form for editing the _whole_ GEDCOM record shows up OK, but upon saving I get an error displayed:

Image

PHP error log shows:
```
[Fri Aug 21 10:10:10 2026] [::1]:61140 [400]: POST /tree/tree1/edit-raw/X1
```
This error occurs at line 68 of `app/Http/Controllers/EditRawRecord.php`.

BertKoor:
fix #5478
`level0` is readonly textarea, should not have been `disabled`, it is required by the controller!


fisharebest:
Thanks

PR #5481, opened 2026-08-21
Use the correct dutch wording ('achter' instead 'over') 

LarsRabe:
First simple fix for grandchildren and nephews/nieces

See also https://www.dutchgenealogy.nl/achternicht-achternicht/

Todo
More generations for grandparents. We do not use multiple 'bet'. Officially the dutch professional genealogists use the OSSEE-scheme; see e.g. https://www.yory.nl/hoe-noem-je-generaties-voorouders/

BUT,... this not common knowledge in the broader dutch speaking population. Most people do not go beyond the 4th Generation:
ouder, grootouder, overgrootouder, betovergrootouder.

### DIFF
--- a/app/I18N/Languages/Dutch.php
+++ b/app/I18N/Languages/Dutch.php
@@ -280,9 +280,9 @@ final readonly class Dutch extends AbstractLanguage
     }
 
     /**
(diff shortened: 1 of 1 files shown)


LarsRabe:
Wrong format

PR #5482, opened 2026-08-21
correcting dutch ancestors after commit 1937c91

LarsRabe:
First simple fix for grandchildren and nephews/nieces

See also https://www.dutchgenealogy.nl/achternicht-achternicht/

Todo
More generations for grandparents. We do not use multiple 'bet'. Officially the dutch professional genealogists use the OSSEE-scheme; see e.g. https://www.yory.nl/hoe-noem-je-generaties-voorouders/

BUT,... this not common knowledge in the broader dutch speaking population. Most people do not go beyond the 4th Generation: ouder, grootouder, overgrootouder, betovergrootouder

### DIFF
--- a/app/I18N/Languages/Dutch.php
+++ b/app/I18N/Languages/Dutch.php
@@ -280,15 +280,15 @@ final readonly class Dutch extends AbstractLanguage
     }
 
     /**
-     * Dynamic descendant relationship using the "over" prefix.
+     * Dynamic descendant relationship using the "achter" prefix.
      *
-     * kleinkind → overkleinkind → overoverkleinkind
+     * kleinkind → achterkleinkind → achterachterkleinkind
      *
      * @return array{string, string}
      */
-    private function over(int $n, string $suffix, Closure $genitive): array
+    private function achter(int $n, string $suffix, Closure $genitive): array
     {
-        return $this->repeat($n, $suffix, 'over', $genitive);
+        return $this->repeat($n, $suffix, 'achter', $genitive);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


fisharebest:
> Officially the dutch professional genealogists use the OSSEE-scheme

> BUT,... this not common knowledge in the broader dutch speaking population.
 
Are you saying that webtrees should implement OSSEE, or that we should keep using "betbetbet..." and "bet x4 ..." approach?

fisharebest:
Implementing OSSEE was easier than I thought.

88c8e5ca73

LarsRabe:
WOW. I have been struggling of how to do this. It looks really easy now. 🥇

Basic OSSEE ist going back 64 generations. That is almost reaching the Birth of Jesus Christ. If we ever going back that far?

Maybe adding more generations is a bit pedantic. ;-) But I love it. 

You're the best and have deserved yourself a coffee (or a tea?) for your work!

fisharebest:
> Basic OSSEE ist going back 64 generations. That is almost reaching the Birth of Jesus Christ. 

128 cycles.  Each cycle has 4 generations.

So 512 generations!

Should reach the birth of Adam and Eve ;-)

LarsRabe:
🤣🤣🤣 
According to Archbishop James Ussher https://en.wikipedia.org/wiki/James_Ussher , Adam und Eve came to this world in 4004 BC. According the Jewish Calender it happenend 3760 BC. 😵‍💫 So, we are way before Adam and Eve. In fact with 512 generations we are at the end of the last Ice Age.🥶 😂 

In the Netherlands we learn at school the first humans (the Batavians or Batavi) came to live in this era in the first century BC. which means approx. 70 generations would allready cover that time frame. 😇

PR #5483, opened 2026-08-22
Fix select components with backed enums

### LINKED ISSUE #5476: 2.3.0-dev: exception when Control Panel renders AccessLevel of modules
Click path: Control Panel - Modules, then any of:

 * Blocks
 * Menus
 * Charts
 * Lists
 * Reports
 * Tabs
 * Sidebars

will result in a stack trace like this:

```
Object of class Fisharebest\Webtrees\Enums\AccessLevel could not be converted to string …/resources/views/components/select.phtml:25
#0 …/app/View.php(163): include()
#1 …/app/View.php(239): Fisharebest\Webtrees\View->render()
#2 …/app/Helpers/functions.php(103): Fisharebest\Webtrees\View::make('components/sele...', Array)
#3 …/resources/views/admin/components.phtml(119): view('components/sele...', Array)
#4 …/app/View.php(163): include('/Users/bertkoor...')
#5 …/app/View.php(239): Fisharebest\Webtrees\View->render()
#6 …/app/Helpers/functions.php(103): Fisharebest\Webtrees\View::make('admin/component...', Array)
#7 …/app/Http/ViewResponseTrait.php(47): view('admin/component...', Array)
#8 …/app/Http/Controllers/AbstractModuleComponent.php(73): Fisharebest\Webtrees\Http\Controllers\AbstractModuleComponent->viewResponse('admin/component...', Array)
#9 …/app/Http/Controllers/ModulesSidebars.php(36): Fisharebest\Webtrees\Http\Controllers\AbstractModuleComponent->listComponents('Fisharebest\\Web...', '<span class="wt...', '')
#10 [internal function]: Fisharebest\Webtrees\Http\Controllers\ModulesSidebars->get()
#11 …/app/Http/Middleware/InvokeController.php(68): ReflectionMethod->invoke(Object(Fisharebest\Webtrees\Http\Controllers\ModulesSidebars))
#12 …/app/Http/MiddlewarePipeline.php(80): Fisharebest\Webtrees\Http\Middleware\InvokeController->process(Object(Nyholm\Psr7\ServerRequest), Object(Fisharebest\Webtrees\Http\RequestHandlers\NotFound))
```

somethingwithproof:
## Summary

- normalize backed-enum selections to their scalar value in the shared select component
- add regression coverage using `AccessLevel::Member`
- restore module access-level management pages affected by #5476

Fixes #5476.

## Testing

- reproduced the original exception with the real select view in Docker using PHP 8.5.9
- `vendor/bin/phpunit --no-coverage tests/Unit/ViewTest.php`
- `vendor/bin/phpcs resources/views/components/select.phtml tests/Unit/ViewTest.php`
- `vendor/bin/phpstan analyze --no-progress resources/views/components/select.phtml tests/Unit/ViewTest.php`

### DIFF
--- a/resources/views/components/select.phtml
+++ b/resources/views/components/select.phtml
@@ -6,10 +6,12 @@ declare(strict_types=1);
  * @var string|null             $class
  * @var string|null             $id
  * @var string                  $name
- * @var array-key|array<string> $selected
+ * @var \BackedEnum|array-key|array<string> $selected
  * @var array<string>           $options
  * @var string|null             $aria_label
  */
+
+$selected_value = $selected instanceof BackedEnum ? $selected->value : $selected;
 
 ?>
 
@@ -17,12 +19,12 @@ declare(strict_types=1);
     class="form-select <?= $class ?? '' ?>"
     name="<?= e($name) ?>"
     id="<?= e($id ?? $name) ?>"
-    <?= is_array($selected) ? 'multiple="multiple"' : '' ?>
+    <?= is_array($selected_value) ? 'multiple="multiple"' : '' ?>
     aria-label="<?= e($aria_label ?? '') ?>"
 
 >
     <?php foreach ($options as $key => $value) : ?>
-        <option value="<?= e((string) $key) ?>"<?= (is_array($selected) ? in_array((string) $key, $selected, true) : (string) $key === (string) $selected) ?  ' selected="selected"' : '' ?>>
+        <option value="<?= e((string) $key) ?>"<?= (is_array($selected_value) ? in_array((string) $key, $selected_value, true) : (string) $key === (string) $selected_value) ?  ' selected="selected"' : '' ?>>
             <?= $value === '' ? '&nbsp;' : e($value) ?>
         </option>
     <?php endforeach ?>
--- a/tests/Unit/ViewTest.php
+++ b/tests/Unit/ViewTest.php
@@ -19,6 +19,7 @@ declare(strict_types=1);
 
 namespace Fisharebest\Webtrees\Tests\Unit;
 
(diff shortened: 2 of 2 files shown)


fisharebest:
Thanks for this PR.

A `select-enum` option might have a place.  See my comments in #5476 

We may also need some sort of interface like this.

```
interface SelectableEnum
{
    public function label(): string;
}

enum AccessLevel: string implements SelectableEnum
{
    public function label(): string
    {
        return match ($this) {
            ...
        };
    }
}
```

But a first step would be to decide if such a component would actually be useful.  How many `select`s do we have that are based on Enum?

somethingwithproof:
I read the linked discussion in #5476, and I agree with the separation: keep `components/select` string-only for the immediate fix, and treat `components/select-enum` as a separate follow-up where both the selection and options have an enum contract.

I checked current `main`. I found **11 select call sites across 5 enums** that could cleanly fit that component:

- `AccessLevel`: 6 — module access, two tree-privacy settings, and three tree-preference settings
- `ContactMethod`: 2 — account edit and admin user edit
- `ChangeStatus`: 1 — changes-log filter
- `Role`: 1 — per-tree user role
- `Restriction`: 1 — `RestrictionNotice` through the generic element editor

There are a few boundary cases I did not count: `HIDE_LIVE_PEOPLE` borrows `AccessLevel` labels but uses legacy `0/1` values rather than the enum's backing values; `Sex` is enum-backed but renders radios; and `restriction-fields.phtml` currently hard-codes values that mirror `Restriction`.

That seems like enough reuse to justify the component and a `SelectableEnum::label()` contract. I would be happy to take the next step as a focused follow-up, starting with the component and `AccessLevel` rather than changing all five enums at once.

PR #5494, opened 2026-09-17
Suppress deprecation notices before the autoloader runs

Schoaf:
Composer eagerly includes some vendor "files" autoload entries (e.g. `illuminate/support/helpers.php`, `guzzlehttp/promises/functions.php`) as soon as `vendor/autoload.php` is required in `index.php` — before `Webtrees::run()` gets a chance to call `error_reporting()`.

On PHP 8.4, several of these vendor files emit "Implicitly marking parameter as nullable is deprecated" notices at declaration time. Since `error_reporting()` hasn't been set yet, these get echoed straight into the output buffer, which then trips `EmitResponse::assertBodyNotEmitted()`'s safeguard and produces a fatal `RuntimeException` on every single request — a hard crash instead of the intended graceful deprecation suppression.

This sets `error_reporting()` to suppress deprecations *before* the autoloader runs. `Webtrees::run()` still sets the final, environment-appropriate level (`E_ALL` in debug builds) immediately afterwards, so this only closes the gap during the autoload step itself.

Tested by reproducing the crash on a PHP 8.4 host running webtrees 2.1.27 and confirming this ordering fix resolves it without needing to touch any vendor code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


fisharebest:
This is a Laravel problem.

When PHP8.4 was released, a syntax change was required.  Laravel refused to back-port the compatibility fix because their policy is "We do not support older versions under any circumstances"

(It's the reason I am currently rewriting the rest of the application to remove the Laravel dependencies.)

fisharebest:
I think we only need this on webtrees 2.1

Can you rebase your change against the `2.1` branch, instead of `main`

fisharebest:
May I suggest this as a comment.

```
// When PHP 8.4 was released, Laravel declined to backport the syntax
// change for implicitly null parameters to versions older than 18 months
// due to their policy (even though it runs happily on newer versions of PHP).
// So to support PHP >= 8.4, we need to disable deprecations during bootstrap.
```

Schoaf:
Done on both counts — rebased against `2.1` and used your suggested comment text. Thanks for the review!

fisharebest:
Minor point.

I think we just need to suppress `E_DEPRECATED`.  Not `E_USER_DEPRECATED`.

Can you check this.

Once that's done, I'll merge this and tag a new release.

Schoaf:
Good catch — updated to only suppress `E_DEPRECATED`.

PR #5495, opened 2026-09-18
Fix: session cookie path becomes "//" for sites installed at domain root

Schoaf:
**Correction (see discussion below):** the original description below overstated this as an active bug. It isn't, for normal installs — see the comment thread for why. Left as a small defensive/idempotent normalization; happy to close instead if preferred.

---

## Summary
- `Session::start()` builds the session cookie path as `$path . '/'`, where `$path` is `parse_url($base_url, PHP_URL_PATH)`.
- If `$path` already ends in `/` (e.g. a hand-edited `config.ini.php` with a trailing slash on `base_url`), the result is a doubled `//` instead of `/`.
- On a normal, browser-driven setup this doesn't happen: `resources/js/webtrees/forms/misc.js` strips all trailing slashes from `location.href` before it's saved as `base_url`, so `$path` never already ends in `/` in practice.
- Fix: use `rtrim($path, '/') . '/'` so the cookie path always has exactly one trailing slash regardless of how many (zero or more) the parsed path already has. Idempotent, harmless either way.

## How I found this
While building a mobile client against a webtrees instance's session-cookie-based login, I found login would silently fail to persist under a standards-strict HTTP client even though `curl` and browsers worked fine. Tracing it back, `Set-Cookie` was being sent with `path=//` — but that turned out to be because my own test instance's `base_url` had a trailing slash from how I'd scripted its setup (bypassing the browser/JS), not something a real install would produce.

## Test plan
- [x] Reproduced with a `base_url` that has a trailing slash: `Set-Cookie` showed `path=//` before the fix, `path=/` after.
- [x] Verified the full login round-trip still works via `curl` both before and after the change (no regression for the normal, no-trailing-slash case).


fisharebest:
> base_url is always stored with a trailing slash

Is it?

a) it shouldn't be
b) I just looked at several sites, and also a clean installation, and the base_url is stored *without* a trailing '/'

Schoaf:
You're right, and thank you for checking — I was wrong about that.

I traced it down: I hadn't actually gone through the setup wizard in a browser for my test instance, I'd scripted it directly against the HTTP endpoints and passed `baseurl=http://localhost:8080/` (with a trailing slash) myself, assuming that was the convention. It isn't — `resources/js/webtrees/forms/misc.js` strips all trailing slashes from `location.href` before putting it in the hidden `baseurl` field (`.replace(/\/+$/, '')`). So on a real, browser-driven setup, `base_url` never has a trailing slash, and the original `$path . '/'` produces exactly one slash — no bug in practice.

So this isn't fixing an active bug for normal installs. The `rtrim($path, '/') . '/'" change is still a harmless, idempotent normalization (e.g. if someone hand-edits `config.ini.php` and adds a trailing slash), but I don't want to overstate it as something this PR needs to fix. Happy to close this if you'd rather not carry a defensive-only change — your call.

fisharebest:
I prefer to avoid code that simply hides an earlier error.

Better to fail early, and identify the underlying issue.

PR #5498, opened 2026-09-18
Missing data report: only flag missing death date/place over age 90

Schoaf:
## Summary

The "Missing data" report always lists a missing death date/place for every individual who doesn't have one, even individuals who could plausibly still be alive. On a tree with many living or recently-added people this drowns the report in false positives ("missing" facts that simply don't exist yet, not facts that were forgotten).

This PR only flags a missing death date/place when the individual has a birth date **and** would be over 90 years old today - i.e. is very likely actually deceased.

## Why

I'm building a small companion app on top of webtrees that reuses this report's logic to surface data-quality issues to users. This false-positive noise was the first thing that got in the way, so it seemed worth fixing upstream rather than only working around it locally - it should help anyone using this report on a tree with living relatives, not just my use case.

## Changes

- `app/Report/ExpressionLanguageProvider.php`: new `age_years(gedcom_date)` report-expression function, returning the number of complete years between a GEDCOM date and today (or `-1` if the date can't be parsed), built on the existing `Age`/`Date` classes.
- `resources/xml/reports/missing_facts_report.xml`: the `DEAT` date/place checks are now wrapped in `(@BIRT:DATE!='') and (age_years(@BIRT:DATE) GT 90)`.

## Test plan

I don't have a local PHP environment in the sandbox I'm working in, so I could not run the test suite or a live report against a tree. I did:
- [x] Validate the XML is well-formed (`xmllint --noout`).
- [x] Trace `PlaceholderExpander::evaluateCondition()` to confirm the `@BIRT:DATE` substitution and ` GT `/` LT ` handling work exactly as used elsewhere in this and other report XML files.
- [ ] Please run/verify against a real tree with a mix of living and long-deceased individuals with and without birth dates.
- [ ] Happy to add a unit test for `age_years()` similar to the existing (currently placeholder) `ExpressionLanguageProviderTest` if that's wanted, and to adjust naming/threshold if 90 isn't the right default.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


LarsRabe:
@Schoaf
90 years old is to early. As I'm working in the medical field, I see rapidly increasing numbers of persons that are older as 90 on a daily basis. Even 100+ is no exeption anymore. I think 110 would be a better number here. 😀

Schoaf:
Yes I know and probably it should be configurable. But right now there is nothing so 90 is a good start. Just because it is wrong about a few people doesn’t mean it’s useless. It more about the majority then about being always right.Am 19.09.2026 um 07:54 schrieb Lars van Ravenzwaaij ***@***.***>:﻿LarsRabe left a comment (fisharebest/webtrees#5498)
@Schoaf
90 years old is to early. As I'm working in the medical field, I see rapidly increasing numbers of persons that are older as 90 on a daily basis. Even 100+ is no exeption anymore. I think 110 would be a better number here. 😀

LarsRabe:
I still disagree with 90. So either make it configurable or raise at least at 99

kiwi3685:
But why not use the variable webtrees already has?
"Age at which to assume an individual is dead" : [MAX_ALIVE_AGE]

It defaults to 120, but can be changed by admin to anything.

LarsRabe:
@kiwi3685 That would be the configurable option as also proposed.

Schoaf:
I agree. Let’s do that.Am 19.09.2026 um 11:58 schrieb Lars van Ravenzwaaij ***@***.***>:﻿LarsRabe left a comment (fisharebest/webtrees#5498)
@kiwi3685 That would be the configurable option as also proposed.

Schoaf:
Pushed — now uses the tree's existing `MAX_ALIVE_AGE` preference ("Age at which to assume an individual is dead", default 120, admin-configurable) instead of a hardcoded 90, via a new `max_alive_age()` report-expression function. Condition is now:

```
(@BIRT:DATE!='') and (age_years(@BIRT:DATE) GT max_alive_age())
```

Thanks @kiwi3685 for pointing at the existing preference instead of inventing a new one.

